PR #8, opened 2021-06-11
Change Python SDK to use instances instead of class variables.

jsupun:
# Will use default config client-config.json
    client_one = Commander()

    # Will use specified client config.
    client_two = Commander(config_file_location=client-config-2.json)

Added module to mock server interactions. This should be by the
unit tests.

Added a mocking module to allow to build response message from the
secret management service. This can be used for unit tests, even in
other projects that use this SDK.

The two instances will not interfer with each other.

### DIFF
--- a/sdk/python/core/keepercommandersm/core.py
+++ b/sdk/python/core/keepercommandersm/core.py
@@ -20,11 +20,26 @@ from keepercommandersm.utils import bytes_to_url_safe_str, base64_to_bytes, sign
 
 class Commander:
 
-    client_key = None
-    server = None
-    verify_ssl_certs = True
+    def __init__(self, client_key=None, server=None, verify_ssl_certs=True, config=None):
 
-    config: KeyValueStorage = None
+        self.client_key = client_key
+        self.server = server
+
+        # Accept the env var PYTHONHTTPSVERIFY. Modules like 'requests' already use it.
+        self.verify_ssl_certs = bool(os.environ.get("PYTHONHTTPSVERIFY", verify_ssl_certs))
+
+        if config is None:
+            config = FileKeyValueStorage()
+
+            # If the server is set in the args, make sure it's makes it's way into the config.
+            if server is not None:
+                config.set(ConfigKeys.KEY_SERVER, server)
+
+        self.config: KeyValueStorage = config
+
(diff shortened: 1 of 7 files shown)


sdk/python/core/keepercommandersm/core.py
```diff
@@ -20,11 +20,26 @@
 
 class Commander:
 
-    client_key = None
-    server = None
-    verify_ssl_certs = True
+    def __init__(self, client_key=None, server=None, verify_ssl_certs=True, config=None, config_file_location=None):
```
maksimu: with  `config_file_location` looks like you are assuming that `Commander`'s default config storage is file. Not sure if should be here but rather as part of the `FileKeyValueStorage` (which it is)

jsupun: For the config, the user can pass in "config" which can be either a FileKeyValueStorage or InMemoryKeyValueStorage. If they don't, then it defaults to FileKeyValueStorage and passes it the config_file_location. If they didn't specify a config_file_location,  it default's the FileKeyValueStorage.default_config_file_location

So if a user did a "c = Commander()", it would be looking, or creating, the file "client-config.json"

Basically "Commander.config" became "Commander(config=FileKeyValueStorage())

maksimu: So, later on when we will have `AWSKeyValueStore`, `GCPKeyValueStore`, `MacOsKeyChainKeyValueStore`, `WindowsKeyValueStore`, `LinuxKeyValueStore`, etc, etc, etc. Then `config_file_location` property is going to be obsolete (will be applied to 1 out of 100 different config storage types). Correct?

jsupun: It would, most likely, not be used in the other. I can remove it and require the storage instance be passed in.

maksimu: we already have storage instance expected and let it be defaulted to `FileKeyValueStorage` with default file name. If someone wants to change the name of the config file then provide instance of the `FileKeyValueStorage('filename.json')`. Let's remove it to keep things clean.

jsupun: Sounds good.

PR #39, opened 2021-07-16
Golang sdk

idimov-keeper:


.github/workflows/test.go.yml
```diff
@@ -0,0 +1,27 @@
...
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  test-go:
+    strategy:
+      matrix:
+        go-version: [1.16.x]
```
maksimu, line 13: why not testing agains GoLang 1.15 also?

idimov-keeper: I'm trying to fix Windows test errors first - I was hoping 1.16 would handle TempDir files removal better :(

PR #44, opened 2021-07-20
Golang sdk

idimov-keeper:


sdk/python/core/keepercommandersm/mock.py
```diff
@@ -70,15 +70,7 @@ class Response:
```
jsupun: Looks like you are reverting some of my changes here,

sdk/python/core/tests/notation_test.py
```diff
@@ -133,56 +133,3 @@ def test_get_notation(self):
             # Custom field, get last name
```
jsupun: Looks like you are reverting some of my changes here.

idimov-keeper: this is resolved now - all your changes are preserved

PR #51, opened 2021-07-21
Java Lib typed records

saldoukhov:


sdk/javascript/packages/core/src/keeper.ts
```diff
@@ -228,11 +228,17 @@ export const getClientId = async (clientKey: string): Promise<string> => {
     return platform.bytesToBase64(clientKeyHash)
 }
 
-export const initializeStorage = async (storage: KeyValueStorage, clientKey: string, domain: string | 'keepersecurity.com' | 'keepersecurity.eu' | 'keepersecurity.au') => {
+export const initializeStorage = async (storage: KeyValueStorage, clientKey?: string, domain?: string | 'keepersecurity.com' | 'keepersecurity.eu' | 'keepersecurity.au') => {
```
maksimu, line 231: also need to add GOV Cloud `govcloud.keepersecurity.us` 

In python we have [THIS COOL](https://github.com/Keeper-Security/secrets-manager/blob/master/sdk/python/core/keepercommandersm/helpers.py#L20) function that gives you ability to get servers by abbreviation (ex. `AU`) . Maybe we should have something like this in other SDKs?

examples/java/hello-secret/src/main/java/hello/App.java
```diff
@@ -41,8 +30,8 @@ public static void main(String[] args) {
 
             // get the password from the first record
             KeeperRecord firstRecord = secrets.getRecords().get(0);
-            String firstRecordPassword = getRecordField(firstRecord, "password");
-            System.out.println(firstRecordPassword);
+            List<String> firstRecordPassword = ((Password) firstRecord.getData().getField(Password.class)).getValue();
```
maksimu, line 33: became a bit verbose.

saldoukhov: But strongly typed. Blame Java.

maksimu: Yeah, this is how we have our data in the record.

sdk/javascript/packages/core/src/keeper.ts
maksimu: here is the link to the function: https://github.com/Keeper-Security/secrets-manager/blob/master/sdk/python/core/keepercommandersm/helpers.py#L20

PR #53, opened 2021-07-22
Replace placeholder notation inline for exec

jsupun:
The reason for this is if a command like this ...

    export MYSECRET=keeper://XXXX/field/login
    ksm exec -- ./runme ${MYSECRET}

is run, Linux will replace the env var before we get to it. This will
cause ./runme to get keeper://XXXX/field/login in the ARGV instead of a
secret.

The inline will parse the cmd items and replace any keeper:// notation
with values.

### DIFF
--- a/integration/keeper_sm_cli/keeper_sm_cli/exec.py
+++ b/integration/keeper_sm_cli/keeper_sm_cli/exec.py
@@ -23,46 +23,60 @@ class Exec:
     def __init__(self, cli):
         self.cli = cli
 
+        # Since the cli is short lived, this won't stick around long.
+        self.local_cache = {}
+
+    def _get_secret(self, notation):
+
+        # If not in the cache, go get the secret and then store it in the cache.
+        if notation not in self.local_cache:
+            value = self.cli.client.get_notation(notation)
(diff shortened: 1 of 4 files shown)


integration/keeper_sm_cli/keeper_sm_cli/exec.py
```diff
@@ -23,46 +23,60 @@ class Exec:
     def __init__(self, cli):
         self.cli = cli
 
-    def env_replace(self):
+        # Since the cli is short lived, this won't stick around long.
+        self.local_cache = {}
```
maksimu, line 27: can local cache be stored to file?

maksimu: checkout [Java](https://github.com/Keeper-Security/secrets-manager/commit/a10d9c1c2a8b33599d93fdf3d91b74dfd7a50816) caching

jsupun: This is not SDK level caching. I'd prefer to do SDK caching, but for v1 this will prevent calling to get the same record four times in a row for a command.

PR #55, opened 2021-07-22
Golang sdk

idimov-keeper:


.gitignore
```diff
@@ -51,13 +51,3 @@ env.bak/
 venv.bak/
 .python-version
 .pytest_cache
-
```
maksimu, line 54: is there a reason why below lines were removed?

PR #62, opened 2021-07-28
Add a test for test UID that start with dashes

jsupun:
It looks like click can handle an arg value that starts with a dash.
Even made it look a parameter.

Also made a few PEP8 fixes. And add the click module to the
requirements.txt.

### DIFF
--- a/sdk/python/core/keeper_secrets_manager_core/mock.py
+++ b/sdk/python/core/keeper_secrets_manager_core/mock.py
@@ -50,7 +50,7 @@ class ResponseQueue:
         response = self.queue.popleft()
         return response.instance(context)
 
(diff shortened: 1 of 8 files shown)


sdk/python/core/keeper_secrets_manager_core/mock.py
```diff
@@ -50,7 +50,7 @@ def get_response(self, context):
         response = self.queue.popleft()
         return response.instance(context)
 
-    def auto_responder_patch(self, path, context, payload_and_signature):
+    def auto_responder_patch(self, _path, context, _payload_and_signature):
```
jsupun, line 53: This is called by mock and needs three params, however only 1 is used. So if you prefix the var was a _, PEP8 validation won't complain about unused vars.

PR #63, opened 2021-07-28
Add key rotation and fix client version

jsupun:
Added a handler for http error/exceptions. It is for both
the get_secrets() and save() method. The exception handler will
format errors messages and throw the correct exception. It will
also handle the key rotation if it gets an 'key' error.

If a key error happens, the public key in the dictionary with
the server provided key_id will be used. The config is updated
with this key_id. The request that caused the excpetion will be
retried.

### DIFF
--- a/sdk/python/core/keeper_secrets_manager_core/core.py
+++ b/sdk/python/core/keeper_secrets_manager_core/core.py
@@ -303,10 +303,10 @@ class SecretsManager:
             # The server wants us to use a different public key.
             elif error == 'key':
                 key_id = response_dict.get("key_id")
-                logging.info("Server has requested we public key {}".format(key_id))
+                logging.info("Server has requested we use public key {}".format(key_id))
 
                 if key_id is None:
-                    raise ValueError("The public key is is blank from the server")
+                    raise ValueError("The public key is blank from the server")
                 elif str(key_id) not in keeper_public_keys:
                     raise ValueError("The public key at {} does not exist in the SDK".format(key_id))
 


sdk/python/core/keeper_secrets_manager_core/core.py
```diff
@@ -273,53 +284,75 @@ def _post_query(self, path, context, payload_and_signature):
...
         return rs
 
+    def handler_http_error(self, rs):
+
+        try:
+            # Decode the JSON content, throw exception if not JSON
+            response_dict = json_to_dict(rs.text)
+            if response_dict is None:
+                raise json.JSONDecodeError("Not JSOM", "NONE", 0)
+
```
maksimu, line 294: JSOM?

jsupun: It's the budget off-brand of JSON

sdk/python/core/keeper_secrets_manager_core/core.py
```diff
@@ -273,54 +284,75 @@ def _post_query(self, path, context, payload_and_signature):
...
+            if error == 'invalid_client_version':
+                logging.error("Client version {} was not registered in the backend".format(
+                    keeper_secrets_manager_sdk_client_id))
+                msg = response_dict.get('additional_info')
+
+            # The server wants us to use a different public key.
+            elif error == 'key':
+                key_id = response_dict.get("key_id")
+                logging.info("Server has requested we public key {}".format(key_id))
```
maksimu: you meant to say "Server has requested we **use** public key" ?

sdk/python/core/keeper_secrets_manager_core/core.py
```diff
@@ -273,54 +284,75 @@ def _post_query(self, path, context, payload_and_signature):
...
+
+                if key_id is None:
+                    raise ValueError("The public key is is blank from the server")
```
maksimu: typo

sdk/python/core/keeper_secrets_manager_core/core.py
maksimu: 💸

PR #85, opened 2021-08-11
java library tune-up

saldoukhov:


sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
```diff
@@ -129,7 +151,7 @@ fun initializeStorage(storage: KeyValueStorage, clientKey: String? = null, hostN
     if (existingClientId != null) {
         throw Exception("The storage is already initialized with a different client Id (${existingClientId})")
     }
-    storage.saveString(KEY_HOSTNAME, hostName)
+    storage.saveString(KEY_HOSTNAME, hostName!!)
```
maksimu: Nullpointer exception will be thrown if nothing supplied. Need to add error checking and thrown appropriate exception

PR #150, opened 2021-10-11
TOTP and b64 config parsing

idimov-keeper:
Added TOTP parsing

### DIFF
--- a/sdk/dotNet/SecretsManager/SecretsManager.csproj
+++ b/sdk/dotNet/SecretsManager/SecretsManager.csproj
@@ -24,7 +24,5 @@
     <ItemGroup>
       <PackageReference Include="Portable.BouncyCastle" Version="1.8.10" />
       <PackageReference Include="System.Text.Json" Version="5.0.2" />
-      <PackageReference Include="System.Text.Encodings.Web" Version="5.0.1" />
     </ItemGroup>
-
 </Project>


sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/CryptoUtils.kt
```diff
@@ -13,6 +16,8 @@ import javax.crypto.KeyAgreement
 import javax.crypto.Mac
 import javax.crypto.spec.GCMParameterSpec
 import javax.crypto.spec.SecretKeySpec
+import kotlin.experimental.and
```
maksimu: What does `kotlin.experimental` mean? Maybe it should not be used in production?

sdk/dotNet/SecretsManager/SecretsManager.csproj
```diff
@@ -27,4 +27,7 @@
       <PackageReference Include="System.Text.Encodings.Web" Version="5.0.1" />
     </ItemGroup>
 
+    <ItemGroup>
```
saldoukhov: This should not be necessary

sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
```diff
@@ -9,11 +9,18 @@ import kotlinx.serialization.encodeToString
 import kotlinx.serialization.json.*
 import java.net.HttpURLConnection.HTTP_OK
 import java.net.URL
+import java.net.URLDecoder
```
saldoukhov: This should not be necessary

sdk/javascript/packages/core/src/platform.ts
```diff
@@ -1,3 +1,5 @@
+import { createHash, createHmac, Hmac } from "crypto"
```
saldoukhov: This is still not fixed

PR #158, opened 2021-10-21
Feature/generate password

idimov-keeper:


sdk/dotNet/SecretsManager/CryptoUtils.cs
```diff
@@ -306,5 +306,91 @@ internal static byte[] Base32ToBytes(string base32)
...
+        const string AsciiSpecialCharacters = @"""!@#$%()+;<>=?[]{}^.,";
+
+        internal static string RandomSample(int sampleLength=0, string sampleString="")
+        {
+            sampleLength = sampleLength < 0 ? 0 : sampleLength;
+            var result = new StringBuilder(sampleLength);
+            if (sampleLength > 0 && !string.IsNullOrEmpty(sampleString))
+            {
+                var data = new byte[4 * sampleLength];
+                using (var crypto = RandomNumberGenerator.Create())
```
saldoukhov: This random generator is different from one that the class uses.

sdk/dotNet/SecretsManager/CryptoUtils.cs
```diff
@@ -306,5 +306,91 @@ internal static byte[] Base32ToBytes(string base32)
...
+            return result.ToString();
+        }
+
+        internal static string Shuffle(string text)
+        {
+            var result = "";
+
+            if (!string.IsNullOrWhiteSpace(text))
+            {
+                Random rng = new Random();
```
saldoukhov: Yet another instance of randomizer, different type this time.

sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/CryptoUtils.kt
```diff
@@ -244,6 +245,65 @@ fun getTotpCode(url: String, unixTimeSeconds: Long = 0): Triple<String?, Int, In
...
+const val AsciiDigits = "0123456789"
+const val AsciiSpecialCharacters = "\"!@#$%()+;<>=?[]{}^.,"
+
+internal fun randomSample(sampleLength: Int=0, sampleString: String=""): String
+{
+    var result = ""
+    val sampleLen = if (sampleLength < 0) 0 else sampleLength
+    if (sampleLen > 0 && sampleString.isNotEmpty())
+    {
+        val random = SecureRandom()
```
saldoukhov: Different than SecureRandom.getInstanceStrong() that is used in other places

PR #215, opened 2022-01-27
Python SDK: Refactoring response from the server and adding warning logs

maksimu:


maksimu:
Example code to get data object and configure warning log level:

```
storage = FileKeyValueStorage(config_file)

secrets_manager = SecretsManager(
    token=token,
    config=storage,
    log_level="DEBUG"
)

all_records = secrets_manager.get_secrets(full_response=True)

print("App name: " + all_records.appData.title)
print("Access expires on: " + (str(all_records.expires_on_str()) if all_records.expiresOn is not 0 else "Never"))

for r in all_records.records:
    print(r)
```

jsupun (approved):

PR #223, opened 2022-02-15
Java stop using bouncycastle dependency

saldoukhov:


mike-jumper:
Is there somewhere I can go for the context of this change?

I'd like to understand why BC is being removed as a dependency of the core Java library, and what implications that may have for downstream usages of KSM like the recently-developed KSM support for Glyptodon/Guacamole. The commit messages themselves are fairly blank with respect to context, sadly.

saldoukhov:
> 

BC is removed because we have used BCFIPS which would conflict with the regular BC if the application uses one already. The final changes make the library to use BCFIPS or FIPS if it is already added to the Security, or fall back to Sun providers if there is no BC present.
The new recommendation for the users of the library is to include BC (preferably FIPS) as a dependency and add it to Java Security before using the KSM

maksimu (approved):

maksimu:
> Is there somewhere I can go for the context of this change?
> 
> I'd like to understand why BC is being removed as a dependency of the core Java library, and what implications that may have for downstream usages of KSM like the recently-developed KSM support for Glyptodon/Guacamole. The commit messages themselves are fairly blank with respect to context, sadly.

The change should not impact the downstream integrations. As Sergey pointed out that if there is BC or BCFIPS in the path then library will pickup Sun's provider. But you should include BCFISP if possible. We will have it documented as we move closer to the release. Right now we will publish only SNAPSHOT version to run some tests and ask customers that are awaiting for this change.

PR #224, opened 2022-02-23
KSM CLI - Expand hostname abbreviations in JSON profile export

Kytech:
In working with the KSM SDKs, I encountered errors stating that the hostname could not be resolved by the system's DNS. Upon further inspection of the errors produced by the SDK, it appears that the request URL is built by pulling whatever value is included for the hostname field in the config file. This appeared consistent across the SDK for multiple languages.

If the config file was created/initialized using the SDK, the hostname is fully expanded to a domain name in the config file. However, if the config file was created with a JSON export from KSM CLI, the hostname will be kept in the two-letter abbreviated form if it uses that format in the keeper.ini file used by KSM CLI.

Since KSM CLI does not output JSON files that conform to what the SDKs are expecting, I made a small addition that decodes the two-letter abbreviation to the full hostname when exporting since the primary purpose of exporting the config file in JSON is for consumption by the SDK.

### DIFF
--- a/integration/keeper_secrets_manager_cli/keeper_secrets_manager_cli/export.py
+++ b/integration/keeper_secrets_manager_cli/keeper_secrets_manager_cli/export.py
@@ -4,6 +4,7 @@ import base64
 import json
 
 from keeper_secrets_manager_core.utils import base64_to_bytes
+from keeper_secrets_manager_core.keeper_globals import keeper_servers
 
 
 class Export:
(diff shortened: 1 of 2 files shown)


jsupun:
In the unit test ...

`integration/keeper_secrets_manager_cli/tests/init_test.py`

There are 3 lines (78, 100, 172) that are looking for **US** as the hostname. Those need to be changed to **keepersecurity.com** for the PR to pass its test.


Beside that, the changes look good. Thanks.

jsupun (approved):

PR #276, opened 2022-05-13
Get record by title

idimov-keeper:


jsupun (approved):

sdk/python/core/keeper_secrets_manager_core/core.py
```diff
@@ -32,6 +32,18 @@
...
+    # Find all records with specified title
+    if records:
+        return [x for x in records if x.title == record_title]
+    return None
+
+def find_secret_by_title(record_title, records):
+    # Find first record with specified title
+    if records:
+        return next((x for x in records if x.title == record_title), None)
+    return None
```
maksimu: maybe here it will be better to return empty list instead of `None`

sdk/python/core/keeper_secrets_manager_core/core.py
```diff
@@ -546,6 +558,27 @@ def get_secrets(self, uids=None, full_response=False):
...
+    def get_secret_by_title(self, record_title):
+        """
+        Retrieve first record with specified title
+        """
+
+        records = self.get_secrets_by_title(record_title) or []
+        if len(records) > 0:
+            return records[0]
+
+        return None
```
maksimu: maybe here it will be better to return empty list instead of `None`

sdk/python/core/keeper_secrets_manager_core/core.py
maksimu: Actually, never mind on this one.

sdk/python/core/keeper_secrets_manager_core/core.py
```diff
@@ -32,6 +32,18 @@
     url_safe_str_to_bytes, bytes_to_base64, generate_random_bytes
 
 
+def find_secrets_by_title(record_title, records):
+    # Find all records with specified title
+    if records:
+        return [x for x in records if x.title == record_title]
+    return None
```
maksimu: maybe here it will be better to return empty list instead of None

maksimu (approved):

PR #326, opened 2022-07-27
Ksm 240

jsupun:


jsupun:
I might want to move this from the SDK to the helper module. If using SDK, then it's the developers job to secure the config. They can include the helper module if they want us to do it for them.

Issue #336, opened 2022-08-10
keeper ansible keeper_create does not use custom record types


mxmorin:
keeper_create in the ansible plugin does not use custom record types.
This is due to test is done BEFORE the load of the custom record types

jsupun:
This is fixed and released in version 1.1.4

PR #340, opened 2022-08-15
Allow more record field args

jsupun:


jsupun:
This is in relation to https://github.com/Keeper-Security/secrets-manager/issues/336

Fixing this problem caused an error on the SDK because privacyScreen, enforceGeneration, and complexity not accepted by RecordField.

PR #519, opened 2023-08-22
Ksm 458 remove core dep on helper

jwalstra-keeper:


sdk/python/core/keeper_secrets_manager_core/dto/dtos.py
```diff
@@ -184,16 +183,32 @@ def set_custom_field_value(self, field_type, value):
         field["value"] = value
         self._update()
 
-    def add_custom_field(self, field: FieldType) -> bool:
-        if issubclass(type(field), FieldType):
-            if self.dict.get('custom', None) is None:
-                self.dict['custom'] = []
-            custom = self.dict['custom']
+    def add_custom_field(self, field_type=None, label=None, value=None, field=None) -> bool:
```
idimov-keeper: This is probably already in use - like this `add_custom_field(field)` so it may be better to move field as first parameter or split into two methods with different mutually exclusive parameter sets

PR #548, opened 2023-12-13
KSM-479 - Python SDK - Removed dependency of strtobool from distutils.util

idimov-keeper:


ewindsor31:
Do we know when this will be approved? I'd like to upgrade to python 3.12 but I have to wait until this change for everything to remain compatible.

PR #568, opened 2024-02-14
KSM-472 Add keeper_get_record action

jwalstra-keeper:
Added a new action to return all the fields in a record as a dictionary. The key in the dictionary is the field label, or field type if the label is missing. The key is normalized to remove any non-alphanumeric characters. Non-alphanumeric characters replaceds with underscores.

To use this action must be used with the 'register' attribute in order to store the results in memory. The results can be accessed the same way other stored structures are stored in Ansbile.


integration/keeper_secrets_manager_ansible/keeper_secrets_manager_ansible/__init__.py
```diff
@@ -617,37 +663,42 @@ def get_field_type_enum_and_key(args):
...
         So this is the third standard.
 
-        kwargs
+        Kwargs
 
-        * length - Length of the password
+        * Length - Length of the password
         * allow_lowercase - Allow lowercase letters. Default is True.
-        * allow_uppercase - Allow uppercase letters. Default is True.
+        * Allow_uppercase - Allow uppercase letters. Default is True.
```
maksimu: did you mean to capitalize these first letters?

jwalstra-keeper: No, I think PyCharm fixed it for me

maksimu: How about other vars above?

jwalstra-keeper: fixed

PR #606, opened 2024-06-07
Release .Net SDK v16.6.6

idimov-keeper:


sdk/dotNet/.editorconfig
maksimu, line 1: @idimov-keeper  do we need this file to be in this PR?

PR #641, opened 2024-08-16
Add Keychain Utility Storage class

amigus:
Create a Keychain Utility Storage class that delegates get/set to a 'utility' binary. The binary must support `set <configuration as string>` and `get` that returns what was set.
The Storage class requires that the binary have a known SHA256 sum unless hash checking is disabled.


maksimu (approved):

stas-schaller:
Closing. Superseded by #888 (OS-Native Secure Storage), which added macOS Keychain support to the CLI in v1.3.0. No further action needed here.

PR #666, opened 2024-09-25
Fix parsing unknown fields breaks record deserialisation

borrow-checker:
This PR fixes an issue where deserialising a record with unknown fields breaks deserialisation of all records. With this change, an error message is displayed and the record is skipped.
***
- fix records parsing failure on unknown fields (now skips)
- cleaned up decryptRecord()


sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
```diff
@@ -301,7 +301,31 @@ data class KeeperFile(
...
+
+    override fun hashCode(): Int {
+        var result = fileKey.contentHashCode()
+        result = 31 * result + fileUid.hashCode()
+        result = 31 * result + data.hashCode()
+        result = 31 * result + url.hashCode()
+        result = 31 * result + (thumbnailUrl?.hashCode() ?: 0)
+        return result
+    }
+}
```
borrow-checker, line 328: We have a bytearray in this data class which means we need to implement our own equals and hashcode

sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
```diff
@@ -744,22 +768,18 @@ private fun fetchAndDecryptSecrets(
     if (response.records != null) {
         response.records.forEach {
             val recordKey = decrypt(it.recordKey, appKey)
-            val decryptedRecord = decryptRecord(it, recordKey)
-            if (decryptedRecord != null) {
-                records.add(decryptedRecord)
-            }
+            decryptRecord(it, recordKey).onSuccess { keeperRecord -> records.add(keeperRecord) }
```
borrow-checker, line 771: this is the magic part where if there's a failure it skips gracefully (i.e. `onSuccess` only gets called if we ge a successful `Result`)

sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
```diff
@@ -777,46 +797,45 @@ private fun fetchAndDecryptSecrets(
...
+        val fileKey = decrypt(recordFile.fileKey, recordKey)
+        val decryptedFile = decrypt(recordFile.data, fileKey)
+        KeeperFile(
+            fileKey,
+            recordFile.fileUid,
+            Json.decodeFromString(bytesToString(decryptedFile)),
+            recordFile.url,
+            recordFile.thumbnailUrl
+        )
+    } ?: emptyList()
```
borrow-checker, line 814: functional > imperative (plus why allocate when you don't need to?)

sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
```diff
@@ -777,46 +797,45 @@ private fun fetchAndDecryptSecrets(
...
+
+    val recordData: KeeperRecordData = try {
```
borrow-checker, line 816: Restructuring this ensures that there are only two  non exception pathways out - either you get the record data you want, or you don't, and both are ok.

sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
saldoukhov: No, we don't do equality checks on this class

borrow-checker: That hashcode is more important here - if this ever goes in a hashed data structure without it it won't work

mike-jumper:
> There is no jira ticket for this. Can’t release to prod w/o a ticket

How apt that this PR is #666. :smiling_imp:

PR #676, opened 2024-10-23
Revert "Fix parsing unknown fields breaks record deserialisation"

idimov-keeper:
Reverts Keeper-Security/secrets-manager#666

### DIFF
--- a/sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
+++ b/sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
@@ -301,31 +301,7 @@ data class KeeperFile(
     val data: KeeperFileData,
     val url: String,
     val thumbnailUrl: String?
-) {
-    override fun equals(other: Any?): Boolean {
-        if (this === other) return true
-        if (javaClass != other?.javaClass) return false
-
-        other as KeeperFile
-
-        if (!fileKey.contentEquals(other.fileKey)) return false
-        if (fileUid != other.fileUid) return false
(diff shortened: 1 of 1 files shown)


sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
```diff
@@ -797,45 +777,51 @@ private fun fetchAndDecryptSecrets(
...
+    // will fail on any unknown field/key - currently just log the error
+    // and continue without the field - nb! field will be lost on save
+    var recordData: KeeperRecordData? = null
+    try {
+        recordData = Json.decodeFromString<KeeperRecordData>(bytesToString(decryptedRecord))
+    } catch (e: Exception) {
+        // New/missing field: Polymorphic serializer was not found for class discriminator 'UNKNOWN'...
+        // New/missing field property (field def updated): Encountered unknown key 'UNKNOWN'.
+        // Avoid 'ignoreUnknownKeys = true' to prevent erasing new properties on save/update
+        println("Record ${record.recordUid} has unexpected data properties (ignored).\n"+
```
maksimu, line 812: This message is very confusing, is the gateway ahead OR behind? can we cahnge it something like this:

```
println("Record ${record.recordUid} contains unrecognized data properties and could not be fully parsed.\n" +
        "This may occur if the Keeper Secrets Manager (KSM) SDK version you're using is not compatible with the record's data schema.\n" +
        "Please ensure that you are using the latest version of the KSM SDK. If the issue persists, contact support@keepersecurity.com for assistance.")
```

PR #732, opened 2025-05-05
[KA-6991]: read GraphSync links

ybodiak:
Add requestLinks option to QueryOptions and CreateOptions types

### DIFF
--- a/sdk/javascript/packages/core/src/keeper.ts
+++ b/sdk/javascript/packages/core/src/keeper.ts
@@ -48,6 +48,7 @@ export type SecretManagerOptions = {
 export type QueryOptions = {
     recordsFilter?: string[]
     foldersFilter?: string[]
+    requestLinks?: boolean
 }
 
 export type CreateOptions = {
@@ -80,6 +81,7 @@ type GetPayload = CommonPayload & {
     publicKey?: string   // passed once when binding
     requestedRecords?: string[] // only return these records
     requestedFolders?: string[] // only return these folders
+    requestLinks?: boolean
 }
 
 type DeletePayload = CommonPayload & {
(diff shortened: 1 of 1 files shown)


saldoukhov (approved):

saldoukhov (changes requested):
This is good for requesting links, but how the client will get the links from the response?
Please verify this code by running against dev environment.

sdk/javascript/packages/core/src/keeper.ts
```diff
@@ -162,6 +164,12 @@ type SecretsManagerResponseRecord = {
     revision: number
     files: SecretsManagerResponseFile[]
     innerFolderUid: string
+    links?: SecretsManagerResponseLink[]
+}
+
+type SecretsManagerResponseLink = {
+    recordUid: string
+    data: string
```
saldoukhov: should be data? since data can be null

sdk/javascript/packages/core/src/keeper.ts
```diff
@@ -162,6 +164,12 @@ type SecretsManagerResponseRecord = {
     revision: number
     files: SecretsManagerResponseFile[]
     innerFolderUid: string
+    links?: SecretsManagerResponseLink[]
+}
+
+type SecretsManagerResponseLink = {
```
saldoukhov: Let's rename to KeeperRecordLink

sdk/javascript/packages/core/src/keeper.ts
```diff
@@ -687,9 +702,12 @@ export const initializeStorage = async (storage: KeyValueStorage, oneTimeToken:
     await platform.generatePrivateKey(KEY_PRIVATE_KEY, storage)
 }
 
-export const getSecrets = async (options: SecretManagerOptions, recordsFilter?: string[]): Promise<KeeperSecrets> => {
+export const getSecrets = async (options: SecretManagerOptions, recordsFilter?: string[], requestLinks?: boolean): Promise<KeeperSecrets> => {
```
saldoukhov: Don't modify this, users which want links will call getSecrets2 which already has this flag in options.

saldoukhov (changes requested):

saldoukhov (approved):

saldoukhov (approved):

PR #803, opened 2025-09-04
KSM-534: JS sdk proxy support

m-afanasiuk:


saldoukhov (changes requested):
We should look into creating an additional package that implements this functionality. One of the requirements for the JS SDK os zero dependencies, this breaks the rule.

sdk/javascript/packages/core/src/platform.ts
```diff
@@ -1,3 +1,5 @@
+import * as http from 'http';
```
saldoukhov: Remove import, platform should stay independent of node types

sdk/javascript/packages/core/src/platform.ts
```diff
@@ -28,6 +30,7 @@ export type Platform = {
     get(url: string, headers: any): Promise<KeeperHttpResponse>
     post(url: string, request: Uint8Array, headers?: { [key: string]: string }, allowUnverifiedCertificate?: boolean): Promise<KeeperHttpResponse>
     fileUpload(url: string, uploadParameters: any, data: Uint8Array | Blob): Promise<any>
+    setCustomProxyAgent(proxyAgent: http.Agent): void
```
saldoukhov: change param to any

sdk/javascript/packages/core/src/platform.ts
```diff
@@ -141,3 +144,7 @@ export const inMemoryStorage = (storage: any): KeyValueStorage => {
         }
     }
 }
+
+export const setCustomProxyAgent = (proxyAgent: http.Agent) => {
```
saldoukhov: change param to any

sdk/javascript/packages/core/package.json
```diff
@@ -23,6 +23,7 @@
     "@types/node": "^22.10.1",
     "babel-jest": "^29.7.0",
     "jest": "^29.7.0",
+    "https-proxy-agent": "^7.0.6",
```
saldoukhov, line 26: No need for the dependency

saldoukhov (changes requested):

sdk/javascript/packages/core/src/platform.ts
m-afanasiuk: Removed

sdk/javascript/packages/core/src/platform.ts
m-afanasiuk: Updated

sdk/javascript/packages/core/src/platform.ts
m-afanasiuk: Updated

PR #814, opened 2025-09-17
Fix missing country enum

stas-schaller:
This fixes the broken GHA Tests across all branches when it finally gets merged into master


stas-schaller:
Should be against a release branch, not master.

PR #825, opened 2025-09-26
KSM-669 KSM JavaScript SDK: Fixed getFolders on browser platform

idimov-keeper:


saldoukhov (changes requested):
Getting rid of CryptoKey's is not acceptable, will break existing one time shares.

PR #888, opened 2025-12-24
Add OS-Native Secure Storage for KSM CLI Configuration

pvagare-ks:
## PR: Add OS-Native Secure Storage for KSM CLI Configuration

This PR replaces plaintext `keeper.ini` credential storage with OS-native secure keyring storage as the default for the KSM CLI.

### Highlights
- Introduces cross-platform keyring-based storage (macOS Keychain, Windows Credential Manager, linux keyring)
- Enables dynamic selection between keyring and INI-based storage
- Provides explicit support for legacy config file usage via CLI option
- Exposes storage utilities for external use

This improves credential security while maintaining backward compatibility.

## Usage

```bash
# Default: stores credentials securely in the OS keyring (no file created)
profile init --token US:YOUR_TOKEN

# Explicit file-based storage (opt-in)
profile init --ini-file keeper.ini --token US:YOUR_TOKEN

### DIFF
--- a/integration/keeper_secrets_manager_cli/keeper_secrets_manager_cli/profile.py
+++ b/integration/keeper_secrets_manager_cli/keeper_secrets_manager_cli/profile.py
@@ -70,15 +70,7 @@ class Profile:
                 launched_from_app=self._config.launched_from_app
             )
         else:
-            found_ini_file = self._find_ini_file()
-            
-            if found_ini_file is not None:
-                self.use_keyring = False
-                self.keyring_storage = None
-                self._config = Config(ini_file=found_ini_file)
-                self._config.load()
-                self.ini_file = self._config
-            elif use_keyring is not None:
+            if use_keyring is not None:
                 self.use_keyring = use_keyring
                 self.keyring_storage = KeyringConfigStorage() if self.use_keyring else None
                 if self.use_keyring:
@@ -90,6 +82,12 @@ class Profile:
                     except Exception as e:
                         self.logger.debug("Unexpected keyring error: %s", e, exc_info=True)
                         print(Fore.YELLOW + "Warning: Keyring access failed (see debug logs)" + Style.RESET_ALL, file=sys.stderr)
+                else:
+                    found_ini_file = self._find_ini_file()
+                    if found_ini_file is not None:
+                        self._config = Config(ini_file=found_ini_file)
+                        self._config.load()
+                        self.ini_file = self._config
             elif KeyringConfigStorage.is_available():
                 self.use_keyring = True
                 self.keyring_storage = KeyringConfigStorage()
@@ -102,9 +100,16 @@ class Profile:
                     self.logger.debug("Unexpected keyring error: %s", e, exc_info=True)
                     print(Fore.YELLOW + "Warning: Keyring access failed (see debug logs)" + Style.RESET_ALL, file=sys.stderr)
             else:
-                # No storage available
-                self.use_keyring = False
-                self.keyring_storage = None
+                found_ini_file = self._find_ini_file()
+                if found_ini_file is not None:
+                    self.use_keyring = False
+                    self.keyring_storage = None
+                    self._config = Config(ini_file=found_ini_file)
+                    self._config.load()
+                    self.ini_file = self._config
+                else:
+                    self.use_keyring = False
+                    self.keyring_storage = None
 
         self.has_profiles = len(self._config.profile_list()) > 0
     
@@ -189,25 +194,7 @@ class Profile:
         """
         from .keyring_config import KeyringConfigStorage
         
-        # First, check if INI file exists (takes priority)
-        ini_file = self._find_ini_file()
-        
-        if ini_file is not None:
-            # INI file found - use file storage
-            self.use_keyring = False
-            self.keyring_storage = None
-            self._config.ini_file = ini_file
-            self._config.has_config_file = True
-            self._config._profiles = {}
-            self._config.load()
-        elif self.use_keyring and self.keyring_storage:
-            # No INI file, reload from keyring
-            self._config._profiles = {}
-            self._load_from_keyring()
-        elif KeyringConfigStorage.is_available():
-            # Try keyring as fallback
-            self.use_keyring = True
-            self.keyring_storage = KeyringConfigStorage()
+        if self.use_keyring and self.keyring_storage:
             self._config._profiles = {}
             try:
                 self._load_from_keyring()
@@ -215,6 +202,30 @@ class Profile:
                 self.logger.warning("Could not load from keyring: %s", e)
             except Exception as e:
                 self.logger.debug("Unexpected keyring error: %s", e, exc_info=True)
+        elif self.ini_file is not None:
+            self._config._profiles = {}
+            self._config.load()
+        else:
+            ini_file = self._find_ini_file()
+            
+            if ini_file is not None:
+                self.use_keyring = False
+                self.keyring_storage = None
+                self._config.ini_file = ini_file
+                self._config.has_config_file = True
+                self._config._profiles = {}
+                self._config.load()
+            elif KeyringConfigStorage.is_available():
+                # Try keyring as fallback
+                self.use_keyring = True
+                self.keyring_storage = KeyringConfigStorage()
+                self._config._profiles = {}
+                try:
+                    self._load_from_keyring()
+                except KsmCliException as e:
+                    self.logger.warning("Could not load from keyring: %s", e)
+                except Exception as e:
+                    self.logger.debug("Unexpected keyring error: %s", e, exc_info=True)
         
         self.has_profiles = len(self._config.profile_list()) > 0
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: sdk/python/core/keeper_secrets_manager_core/storage.py [hunk 1]
```diff
@@ -225,3 +228,215 @@ def is_base64(s):
...
+        if self.use_python_keyring:
+            import keyring
+            keyring.set_password(self.keyring_application_name, key, value)
+        else:
+            self.__run_keyring_utility(["set", key, value])
+
+    def __run_keyring_utility(self, args: list[str]) -> str:
+        """Run lkru utility (Linux only fallback)."""
+        try:
+            match args[0]:
```
stas-schaller: `match` was introduced in python 3.10 which is higher than our minimum of 3.7. Please replace with an if/elif statement to maintain compatibility.

integration/keeper_secrets_manager_cli/setup.py
```diff
@@ -22,7 +22,8 @@
     'pyyaml',
     'update-checker',
     'psutil',
-    'boto3'
+    'boto3',
+    'keyring'
```
stas-schaller: Per [KC-1018](https://keeper.atlassian.net/browse/KC-1018) I agree we should include keyring but I think we should make it an optional part of the installation process. 

Currently, this PR is a breaking change for anyone in a serverless or headless environment. Keyring requires `D-Bus` and that means serverless on linux (Lambda, Azure functions, etc) fail during runtime due to missing daemons, and headless (EC2, etc) need to install and manage D-Hub daemons to run. Basically because those are environments without a user at a GUI, there's no expectation of a keyring. They use env vars for that. 

I propose making this an optional dependency that is then strongly communicated to the user during profile initialization. 

Optional installation: 
remove keyring from `install_requires` and add 
```python
    extras_require={
        'keyring'
    }
```
This means that to install keyring users would have to do `pip install keeper-secrets-manager-cli[keyring]`

See my feedback in profile.py for a recommendation on graceful fallback and UX for this change

integration/keeper_secrets_manager_cli/keeper_secrets_manager_cli/profile.py
```diff
@@ -40,32 +41,170 @@ def __init__(self, cli, ini_file=None, config=None):
...
+                self.keyring_storage = KeyringConfigStorage() if self.use_keyring else None
+                if self.use_keyring:
+                    try:
+                        self._load_from_keyring()
+                    except Exception:
+                        pass
+            elif KeyringConfigStorage.is_available():
+                self.use_keyring = True
+                self.keyring_storage = KeyringConfigStorage()
+                try:
```
stas-schaller, line 94: Given that the keyring package should be an extra and not required (see my comment in requirements.txt for my justification) we should robustly inform the user about keyring as a more robust alternative. Consider :

```python
try:
    self._load_from_keyring()
except KsmCliException as e:
    self.logger.warning(f"Could not load from keyring: {e}")
    print(Fore.YELLOW + f"Warning: Failed to load profiles from keyring: {e}" + Style.RESET_ALL, file=sys.stderr)
except Exception as e:
    self.logger.debug(f"Unexpected keyring error: {e}", exc_info=True)
    print(Fore.YELLOW + "Warning: Keyring access failed (see debug logs)" + Style.RESET_ALL, file=sys.stderr)
```

integration/keeper_secrets_manager_cli/keeper_secrets_manager_cli/profile.py
```diff
@@ -138,30 +281,113 @@ def init(token, ini_file=None, server=None, profile_name=None, launched_from_app
...
+                               app_key=config_storage.get(ConfigKeys.KEY_APP_KEY),
+                               hostname=config_storage.get(ConfigKeys.KEY_HOSTNAME),
+                               app_owner_public_key=config_storage.get(ConfigKeys.KEY_OWNER_PUBLIC_KEY),
+                               server_public_key_id=config_storage.get(ConfigKeys.KEY_SERVER_PUBLIC_KEY_ID))
+
+            if config.config.active_profile is None:
+                config.config.active_profile = profile_name
 
-        config.save()
+            config.save()
```
stas-schaller, line 419: After we save the config, let's ensure that it's owned by the user instead of everyone. 

Consider: 
```python
if not use_keyring and ini_file and os.path.exists(ini_file):
    try:
        os.chmod(ini_file, 0o600)  # Owner read/write only
        print(f"Set keeper.ini permissions to 0600 (owner-only access)", file=sys.stderr)
    except Exception as e:
        self.logger.warning(f"Could not set file permissions: {e}")
```

integration/keeper_secrets_manager_cli/keeper_secrets_manager_cli/keyring_config.py
```diff
@@ -0,0 +1,235 @@
...
+            except KsmCliException as e:
+                self.logger.debug(f"Failed to delete profile during clear_all: {e}")
+            except Exception as e:
+                self.logger.debug(f"Unexpected error deleting profile during clear_all: {e}")
+        
+        # Delete common config
+        self.delete_common_config()
+    
+    @staticmethod
+    def is_available() -> bool:
```
stas-schaller, line 465: My testing revealed that this is insufficient to know if keyring is available. It checks for install but there could be underlying issues that only come up during runtime (like a missing daemon). 

Consider checking for "fail" in the messaging instead (I kept getting `print(backend) -> <keyring.backends.fail.Keyring object>` for example):
```python
@staticmethod
def is_available() -> bool:
    """Check if keyring is available and working."""
    try:
        import keyring
        backend = keyring.get_keyring()

        # Reject fail.Keyring backend (doesn't actually store anything)
        backend_module = backend.__class__.__module__
        if 'fail' in backend_module.lower():
            return False

        return True
    except ImportError:
        return False
```

sdk/python/core/keeper_secrets_manager_core/storage.py
```diff
@@ -225,3 +228,215 @@ def is_base64(s):
             return base64.b64encode(base64.b64decode(s)) == str.encode(s)
         except (Exception,):
             return False
+
+class KeyringUtilityStorage(KeyValueStorage):
```
stas-schaller: This and all changes in the python sdk core should be moved to keeper _secrets_manager_cli. Keyring is a CLI specific feature and we don't want all SDK users to have to install it for a CLI specific feature they might never use.

stas-schaller (changes requested):
Along with the comments I've made, we also need to include unit and integration tests for any new features. This is pretty major change and we need to make sure it gets introduced with minimal risk of regressions.

sdk/python/core/keeper_secrets_manager_core/storage.py
pvagare-ks: Done

integration/keeper_secrets_manager_cli/setup.py
pvagare-ks: added extras_require in setup.py for 'keyring'

integration/keeper_secrets_manager_cli/keeper_secrets_manager_cli/profile.py
pvagare-ks: Done

integration/keeper_secrets_manager_cli/keeper_secrets_manager_cli/profile.py
pvagare-ks: Done

integration/keeper_secrets_manager_cli/keeper_secrets_manager_cli/keyring_config.py
pvagare-ks: Done

sdk/python/core/keeper_secrets_manager_core/storage.py
pvagare-ks: Moved the keyring specific code to keeper _secrets_manager_cli.

integration/keeper_secrets_manager_cli/keeper_secrets_manager_cli/profile.py
```diff
@@ -138,35 +300,127 @@ def init(token, ini_file=None, server=None, profile_name=None, launched_from_app
...
+                    print("  No keeper.ini file will be created - credentials stored securely", file=sys.stderr)
+                else:
+                    print(f"Added profile {profile_name} to system keyring", file=sys.stderr)
+                    print("  No keeper.ini file will be created - credentials stored securely", file=sys.stderr)
+            except KsmCliException:
+                raise
+            except Exception as e:
+                raise KsmCliException("Failed to save profile to keyring: {}".format(e))
+        else:
+            # User is using INI file storage - block if keychain has profiles
```
stas-schaller: Lines 378 - 394 should be deleted for backwards compatibility. There are situations where users will want to be able to use both keyring and ini at the same time. For example, if a user using keyring and they wanted to generate an INI for Docker or CI/CD, this code would stop them from generating it.

integration/keeper_secrets_manager_cli/keeper_secrets_manager_cli/profile.py
```diff
@@ -98,12 +249,23 @@ def get_common_config(self):
...
+
+        logger = logging.getLogger(__name__)
+
+        keyring_available = KeyringConfigStorage.is_available()
+        use_keyring = not use_config_file and keyring_available
+
+        if not use_keyring and not use_config_file:
+            use_config_file = True  # Fall back to file storage
+            if not keyring_available:
+                logger.debug("Keyring not available, falling back to INI file storage")
```
stas-schaller: We should also warn the user about this, consider adding 
```python
print(Fore.YELLOW + "Warning: Keyring not available, using keeper.ini file instead" + Style.RESET_ALL, file=sys.stderr)
print(Fore.YELLOW + "Warning: For better security, install: pip install keeper-secrets-manager-cli[keyring]" + Style.RESET_ALL, file=sys.stderr)
```

integration/keeper_secrets_manager_cli/keeper_secrets_manager_cli/profile.py
pvagare-ks: However, this would result in two configurations—one in the keyring and one in an INI file. Which configuration should the host user use in this case?

integration/keeper_secrets_manager_cli/keeper_secrets_manager_cli/profile.py
pvagare-ks: Done

integration/keeper_secrets_manager_cli/keeper_secrets_manager_cli/profile.py
pvagare-ks: Added a precedence rule: use the OS keyring when it’s available; otherwise, fall back to the INI file.

PR #975, opened 2026-03-23
Release Java SDK v17.2.1

### LINKED ISSUE #973: Keeper Java/Kotlin: Field 'lastModified' is required for type with serial name...
Using Keeper Java/Kotlin JDK 1.7.2 I'm getting an exception
```
File zsBJgB2olCGtrWCCqYz7fA skipped due to error: MissingFieldException, Field 'lastModified' is required for type with serial name 'com.keepersecurity.secretsManager.core.KeeperFileData', but it was missing at path: $
```
As is written in exception message the `KeeperRecordField` is not returned.

I see that for [.Net ](https://docs.keeper.io/en/release-notes/enterprise/keeper-secrets-manager/2026/february-2026) is fixed (KSM-674: Fixed parsing of lastModified file data field). What about Java/Kotlin?

stas-schaller:
## Summary
Release branch for Java SDK v17.2.1 — seven changes: IL5 region support, serialization correctness for record create, file attachment crash fixes, shared folder record decryption, null file URL handling, file descriptor leak fixes, and Base64 null guard.

### New Features
- **IL5 region mapping** (KSM-902): added `"IL5" -> "il5.keepersecurity.us"` to the `initializeStorage()` region map in `SecretsManager.kt`. Tokens prefixed `IL5:` now route to the DoD Impact Level 5 environment.

### Bug Fixes
- **Custom field omitted from record create payload** (KSM-823): `KeeperRecordData.custom` defaulted to `null`, causing `kotlinx-serialization` to omit `"custom"` from the V3 API payload when no custom fields were set. Changed default to `mutableListOf()` and added `@EncodeDefault` to force inclusion. Regression test added.
- **KeeperFileData crash when `lastModified` absent** (KSM-854): files uploaded by non-SDK clients (iOS, Android, Web Vault) omit `lastModified`; `kotlinx-serialization` treated it as required, throwing `MissingFieldException` and silently dropping the attachment. Added `= 0` default, consistent with .NET SDK behavior (KSM-674).
- **Shared folder records return null fields** (KSM-753): records created via Commander/PowerShell in shared folders appear in `response.records[]` with `innerFolderUid` set, but their `recordKey` is encrypted with the folder key, not the app key. Now builds a folder-key map before processing flat records and uses the correct key when `innerFolderUid` matches.
- **NPE crash when file `url` absent** (KSM-765): `KeeperFile.url` and `SecretsManagerResponseFile.url` changed from `String` to `String?`; `downloadFile()` now throws a typed exception when `url` is null rather than crashing with an NPE.
- **File descriptor leaks** (KSM-855): `LocalConfigStorage` replaced four manual `stream.close()` patterns with Kotlin `.use { }` — the init block never closed its reader at all. `downloadFile`, `uploadFile`, and `postFunction` now wrap `HttpsURLConnection` in `try/finally` + `disconnect()` — `with()` is a scope function, not try-with-resources, and did not guarantee cleanup on exception.
- **Opaque NPE from Base64 decoders** (KSM-985): `base64ToBytes("")` and `webSafe64ToBytes("")` now throw a typed Keeper exception on empty input instead of an NPE from inside `java.util.Base64` (KSM-808 parity).

### Breaking Changes
- `KeeperFile.url` changed from `String` to `String?`. Callers that previously treated `url` as non-null must now handle the nullable case. This is a correctness fix — the server can omit this field and the previous behavior was an NPE crash.

## Related Issues
- Closes #973
- Merges #974 (KSM-854)
- Merges #979 (KSM-823)
- Merges #996 (KSM-902)
- KSM-753, KSM-765, KSM-855, KSM-985

### DIFF
--- a/.github/workflows/publish.maven.java.core.yml
+++ b/.github/workflows/publish.maven.java.core.yml
@@ -1,13 +1,22 @@
 name: Publish to Maven (Java Core SDK)
 on:
   workflow_dispatch:
+    inputs:
+      publish:
+        description: 'Publish to Maven Central (uncheck to build only)'
+        required: false
+        default: 'true'
+        type: boolean
 
 jobs:
   publish:
+    permissions:
+      contents: read
     uses: ./.github/workflows/reusable.maven.central.publish.yml
     with:
       working-directory: ./sdk/java/core
       project-name: keeper-secrets-manager-java
       project-title: Keeper Secrets Manager Java Core SDK
       java-version: '11'
+      publish: ${{ inputs.publish }}
     secrets: inherit
--- a/.github/workflows/reusable.maven.central.publish.yml
+++ b/.github/workflows/reusable.maven.central.publish.yml
@@ -159,6 +159,7 @@ jobs:
 
   publish-to-maven-central:
     needs: [get-version, test, generate-and-upload-sbom, confirm-publish]
+    if: ${{ inputs.publish == true }}
     environment: prod
     runs-on: ubuntu-latest
 
--- a/.github/workflows/reusable.maven.central.publish.yml
+++ b/.github/workflows/reusable.maven.central.publish.yml
@@ -44,7 +44,7 @@ jobs:
         env:
           PROJECT_NAME: ${{ inputs.project-name }}
         run: |
-          VERSION=$(grep -Po 'version\s*=\s*"\K[^"]*' build.gradle.kts || echo "0.0.0-unknown")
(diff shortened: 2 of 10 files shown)


stas-schaller:
Superseded by #1049. The release was promoted from a patch (17.2.1) to a minor bump (17.3.0) since it ships a new feature (IL5 dynamic key, KSM-902) alongside the fixes. The branch `release/sdk/java/core/v17.2.1` has been renamed to `release/sdk/java/core/v17.3.0` (same commits) and #1049 picks up where this left off. REL-5094.

PR #1033, opened 2026-06-08
feat(python): KSM-877 throttle retry with exponential backoff

mgallego-keeper:
> **Supersedes #1031**, which was auto-closed when its base branch `release/sdk/python/core/v17.2.2` was deleted. Same change, retargeted to `release/sdk/python/core/v17.3.0` with the CHANGELOG entry moved under `## 17.3.0`.

## Summary
Implements throttle retry with exponential backoff for the Python SDK (epic KSM-876, task KSM-877). The Keeper backend throttles with **HTTP 403 `{"error":"throttled"}`**; previously the SDK raised `KeeperError` on the first throttle. Now `_post_query` retries transparently, so every `SecretsManager` operation gains throttle resilience with no caller changes.

## What changed
- **`core.py`** — in the `_post_query` loop, detect throttle responses *before* `handler_http_error` and retry with exponential backoff + jitter; raise `KeeperThrottleError` once retries are exhausted. Two static helpers added: `_parse_throttle` (detection / `retry_after`) and `_throttle_delay` (backoff + jitter).
- **`keeper_globals.py`** — `MAX_THROTTLE_RETRIES = 5`, `BASE_THROTTLE_DELAY_SEC = 11` (1s margin over the backend's 10s memcached TTL). No new config keys or env vars.
- **`exceptions.py`** — new `KeeperThrottleError(KeeperError)`; subclassing keeps existing `except KeeperError` handlers working.
- **`CHANGELOG.md`** — entry under `## 17.3.0` (matching the v17.3.0 release branch; `_version.py` is already `17.3.0` on this branch).

## Algorithm
Delays `11s, 22s, 44s, 88s, 176s` (each ±25% jitter) → ~341s worst case over 5 retries. A `retry_after` in the response body takes precedence when present. `handler_http_error` is left untouched, so **key-rotation retry is unaffected**.

## Tests
New `tests/throttle_test.py` — 23 unit + end-to-end cases (`time.sleep` / `random.uniform` patched) covering delay computation, throttle detection, retry-then-success, multi-throttle delay verification, exhaustion → `KeeperThrottleError`, `retry_after` honored, write-path (`save`) retry, throttle + key-rotation compose, non-throttle 403/502 not retried, per-call counter reset, and WARNING logging; plus an optional env-gated live-backend test.

```
$ python -m pytest tests/ -q
102 passed, 1 skipped, 2 warnings
```

Jira: [KSM-877](https://keeper.atlassian.net/browse/KSM-877)

[KSM-877]: https://keeper.atlassian.net/browse/KSM-877?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sdk/python/core/keeper_secrets_manager_core/core.py
+++ b/sdk/python/core/keeper_secrets_manager_core/core.py
@@ -13,9 +13,11 @@ import hmac
 import json
 import logging
 import os
+import random
 import re
 import requests
 import sys
+import time
 from base64 import urlsafe_b64decode
 from http import HTTPStatus
 from typing import List, Tuple, Optional
@@ -31,9 +33,10 @@ from keeper_secrets_manager_core.dto.payload import GetPayload, \
     EncryptedPayload, KSMHttpResponse, CreatePayload, FileUploadPayload, \
     DeletePayload, CreateFolderPayload, UpdateFolderPayload, \
     DeleteFolderPayload, CreateOptions, QueryOptions
-from keeper_secrets_manager_core.exceptions import KeeperError
+from keeper_secrets_manager_core.exceptions import KeeperError, KeeperThrottleError
 from keeper_secrets_manager_core.keeper_globals import keeper_public_keys, \
-    keeper_secrets_manager_sdk_client_id, logger_name, keeper_servers
+    keeper_secrets_manager_sdk_client_id, logger_name, keeper_servers, \
+    MAX_THROTTLE_RETRIES, BASE_THROTTLE_DELAY_SEC
 from keeper_secrets_manager_core.storage import FileKeyValueStorage, \
     KeyValueStorage, InMemoryKeyValueStorage
 from keeper_secrets_manager_core.utils import base64_to_bytes, dict_to_json, \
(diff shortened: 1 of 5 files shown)


sdk/python/core/keeper_secrets_manager_core/core.py
```diff
@@ -669,6 +674,23 @@ def _post_query(self, path, payload):
             if ksm_rs.status_code == 200:
                 break
 
+            # The backend throttles with HTTP 403 {"error":"throttled"}. Retry with exponential
+            # backoff + jitter before falling through to generic error handling. Checked before
+            # handler_http_error so the existing key-rotation retry path is left untouched.
+            retry_after = self._parse_throttle(ksm_rs.http_response)
```
stas-schaller: Looks like this method doesn't actually check the status code, so a non-403 response (like 500) with {"error":"throttled"} in the body would trigger all 5 retries every time. I recommend wrapping the method call in a check for a 403: 
```python
if ksm_rs.status_code == 403:
    retry_after = self._parse_throttle(ksm_rs.http_response)
    if retry_after is not None:
    ...
```

stas-schaller (changes requested):

PR #1036, opened 2026-06-11
feat(python): KSM-992 typed KeeperRecordLink linked-credential accessors

mgallego-keeper:
## Summary

Adds a typed `KeeperRecordLink` linked-credential accessor layer and `Record.get_links()` to the Python Core SDK ([KSM-992](https://keeper.atlassian.net/browse/KSM-992), epic KSM-991). Callers no longer need to base64-decode, JSON-parse, or AES-decrypt `record.links` entries by hand. Purely additive — the raw `record.links` list of dicts and all existing parsing are unchanged, and there are no protocol changes (`QueryOptions.request_links` already existed).

## API

`record.get_links()` returns `List[KeeperRecordLink]` (entries without a `recordUid` are skipped). Each typed link exposes `record_uid` / `data` / `path` attributes, the untouched original dict in `.raw`, and:

**Java-parity accessors** — `is_admin_user()`, `is_launch_credential()`, `allows_rotation()`, `allows_connections()`, `allows_port_forwards()`, `allows_session_recording()`, `allows_typescript_recording()`, `allows_remote_browser_isolation()`, `rotates_on_termination()`, `get_link_data_version()`, `get_decoded_data()`, `has_readable_data()`, `might_be_encrypted()`, `has_encrypted_data()`, `get_decrypted_data(record_key)` (AES-256-GCM), `get_link_data(record_key)`, `get_ai_settings_data(record_key)`, `get_jit_settings_data(record_key)`, `get_settings_for_path(path, record_key)`.

**Live-shape accessors (beyond Java)** — `is_iam_user()`, `belongs_to()`, `no_update_services()`, `ai_enabled()`, `ai_session_terminate()`, `get_allowed_settings()`, `get_rotation_settings()`, `get_meta_data()`.

All accessors are non-throwing: decode/parse/decrypt failures yield `None`/`False`. Values are read strictly (no string→bool/int coercion; bool is not accepted as an int version), matching Java/Rust semantics.

## Verified backend contract

The implementation was mapped against a live QA PAM app (24 links across pamMachine/pamDatabase/pamRemoteBrowser/pamUser/pamNetworkConfiguration records). Link objects carry exactly `{recordUid, data: base64|null, path: string|null}`. Observed payloads:

| `path` | Encoding | Payload |
|---|---|---|
| `"meta"` (self-link) | plain b64 JSON | `allowedSettings{rotation, connections, portForwards, sessionRecording, typescriptRecording, aiEnabled, aiSessionTerminate, remoteBrowserIsolation}`, `rotateOnTermination`, `version`, `no_update_services` |
| `null` (credential link) | plain b64 JSON | `is_admin`, `is_launch_credential`, `is_iam_user`, `belongs_to`, `rotation_settings{schedule, pwd_complexity, disabled, noop, saas_record_uid_list}` |
| `null`, `data: null` | — | pure record reference |
| `"ai_settings"` (self-link) | AES-256-GCM under record key | `version` (string), `riskLevels{…}` |
| `"jit_settings"` (self-link) | AES-256-GCM under record key | `createEphemeral`, `elevate`, `elevationMethod`, `elevationString`, `baseDistinguishedName` |

Two deliberate deviations, both grounded in the live data:

1. **No `get_pam_users()`-style type filters** (despite the ticket AC): link payloads carry no record type, so per-type filtering is not buildable. Same conclusion the Rust SDK reached in KSM-997.
2. **Permission booleans read `allowedSettings` as a fallback** (top-level key wins): the backend now nests permission flags under `allowedSettings` in `meta` links. Java reads top-level keys only and therefore returns `false` on all current data; cross-SDK follow-ups are tracked in JIRA.

## Testing

- New `tests/record_link_test.py` — 18 tests: the 10 cases from the Rust KSM-997 suite (boolean defaults, strict typing/no coercion, decrypt round-trip incl. wrong/missing key, plain-vs-encrypted auto-detection, path gating, encryption detection, end-to-end `mock.Response` → `get_secrets()` → `get_links()` with malformed-entry skip) plus 7 live-shape cases (meta/allowedSettings fallback + precedence, credential `rotation_settings`, data-less references, current ai/jit payloads, losslessness of unknown fields).
- Full suite: **125 passed, 1 skipped** (pre-existing) — no regressions; `record_test.py::test_record_graphsync_links` confirms the raw `links` behavior is untouched.
- Live verification against the QA PAM app: all 24 links round-trip through every accessor (including ai/jit AES-GCM decryption with the record key) with zero failures.

## Related Issues

- KSM-992 (epic KSM-991)
- Rust precedent: KSM-997 (commit c715c71)

[KSM-992]: https://keeper.atlassian.net/browse/KSM-992?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sdk/python/core/keeper_secrets_manager_core/dto/dtos.py
```diff
@@ -291,6 +315,303 @@ def print(self):
...
+        record's key. The returned dict preserves all fields sent by the server,
+        including ones this SDK version doesn't know about yet.
+
+        :param record_key: Optional record key bytes for encrypted link data
+        :return: Parsed payload as a dict, or None if parsing fails
+        """
+        decoded = self.get_decoded_data()
+        if decoded is None:
+            return None
+        if decoded.startswith("{") or decoded.startswith("["):
```
stas-schaller, line 536: Minor issue but still possible to run into: 

AES-GCM output starts with a random byte, so it's possible to get { or [ as the first character. The payload is then misclassified as plain JSON, json.loads fails, and we return `None` without ever trying to decrypt, even with the correct key. This flows through get_ai_settings_data() / get_jit_settings_data() / get_settings_for_path().

The 24-link live verification wouldn't surface this (expected hits ≈ 0.19), but across a customer's ai/jit links it would intermittently and non-deterministically drop data. The class already has a content-based detector (has_encrypted_data()); get_link_data() just doesn't use it. Minimal fix is to fall back to decryption when the fast-path parse fails:

```python
if decoded.startswith("{") or decoded.startswith("["):
    parsed = KeeperRecordLink._parse_json_to_dict(decoded)
    if parsed is not None:
        return parsed
    # leading {/[ was coincidental ciphertext — fall through to decrypt
decrypted = self.get_decrypted_data(record_key)
if decrypted is None:
    return None
return KeeperRecordLink._parse_json_to_dict(decrypted)
```

stas-schaller (changes requested):
Almost there!

sdk/python/core/keeper_secrets_manager_core/dto/dtos.py
mgallego-keeper: Good catch — reproduced exactly as described: with an attacker-chosen IV starting `{` (or `[`), `get_decrypted_data(key)` returned the plaintext while `get_link_data(key)` returned `None` (and `get_jit_settings_data`/`get_ai_settings_data` dropped the data with it).

Fixed in 8855cb5 with the suggested parse-then-fall-through, plus a deterministic regression test (`test_ciphertext_with_json_like_first_byte`) that crafts ciphertexts beginning with both `{` and `[` via the IV parameter of `CryptoUtils.encrypt_aes` and asserts `get_link_data(key)`, `get_jit_settings_data(key)` and `get_settings_for_path(path, key)` all decrypt, while the keyless path still returns `None`. Genuine plain JSON and no-key/wrong-key behavior are unchanged (pinned by the existing tests — suite now 18).

Also carried the same fix to the Rust port in #1037, which inherited the identical early-return. The Java SDK shares this flaw (same pattern origin) — noted on epic KSM-1007 so KSM-1008 picks it up.

stas-schaller (approved):

PR #1043, opened 2026-06-12
feat(java): KSM-878 throttle retry with exponential backoff

mgallego-keeper:
## Summary

Implements automatic **throttle retry with exponential backoff** for the Java/Kotlin SDK, part of epic **KSM-876** — ticket **KSM-878**. Ports the algorithm already shipped in the Python (KSM-877) and Go (KSM-881) SDKs.

The Keeper backend throttles per `{endpoint}+{clientId}` (100 req / 10s window) and returns **HTTP 403** with body `{"error":"throttled"}`. Previously the SDK surfaced this as an immediate `Exception`; now `postQuery` retries transparently.

## Algorithm

- Constants `MAX_THROTTLE_RETRIES = 5`, `BASE_THROTTLE_DELAY_SEC = 11` (1s margin over the backend's 10s memcached TTL).
- On a **403** whose body is `result_code`/`error` == `"throttled"` (detected **before** the key-rotation handler so that path — including the IL5 custom-key suppression — is untouched, and **gated on 403** so a 500/502 carrying a throttled body is not retried):
  - if retries are exhausted → throw the new `KeeperThrottleException`;
  - otherwise log a warning, sleep `retry_after` (if > 0) else `11 * 2^attempt` (11/22/44/88/176s) with **±25% jitter**, then retry.

## Changes

- `SecretsManager.kt` — `MAX_THROTTLE_RETRIES` / `BASE_THROTTLE_DELAY_SEC` constants; `parseThrottle` / `throttleDelayMillis` / `throttleJitter` helpers (internal, for unit tests); retry logic in `postQuery`; optional `throttleSleepMillis` seam on `SecretsManagerOptions` (defaults to `Thread.sleep`).
- `SecretsManagerExceptions.kt` — new public `KeeperThrottleException`.
- `ThrottleTest.kt` — new test class. No version bump; the Java SDK keeps no CHANGELOG.

## Tests (kotlin.test / JUnit)

- Unit: exponential sequence `[11,22,44,88,176]s` with pinned jitter, `retry_after` precedence, jitter bounds; `parseThrottle` table.
- E2E (via `getSecrets` with a mocked `queryFunction` + recording `throttleSleepMillis`): retry-then-success, exhaustion → `KeeperThrottleException`, `retry_after` honored, non-throttle 403 not retried, 502-with-throttled-body not retried.
- `ThrottleTest`: **9 tests pass**; full module suite: **38 tests, 0 failures** (existing `getSecretsE2E` unaffected).

Base branch: `release/sdk/java/core/v17.2.1` (can be retargeted once a dedicated release branch is cut).

### DIFF
--- a/sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
+++ b/sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
@@ -1730,10 +1730,12 @@ private inline fun <reified T> postQuery(
                         throw KeeperThrottleException("Request throttled by Keeper backend; exhausted $MAX_THROTTLE_RETRIES retries")
                     }
                     val delayMs = throttleDelayMillis(throttleAttempt, retryAfter, throttleJitter())
-                    System.err.println(
-                        "WARNING: Request throttled (attempt ${throttleAttempt + 1}/$MAX_THROTTLE_RETRIES); " +
-                            "retrying in ${"%.1f".format(delayMs / 1000.0)}s"
-                    )
+                    if (options.loggingEnabled) {
+                        System.err.println(
+                            "WARNING: Request throttled (attempt ${throttleAttempt + 1}/$MAX_THROTTLE_RETRIES); " +
+                                "retrying in ${"%.1f".format(delayMs / 1000.0)}s"
+                        )
+                    }
                     throttleSleep(delayMs)
                     throttleAttempt++
                     continue
--- a/sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
+++ b/sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
@@ -1700,6 +1700,7 @@ internal fun throttleDelayMillis(attempt: Int, retryAfter: Double, jitter: Doubl
 // throttleDelayMillis with a pinned jitter instead.
 internal fun throttleJitter(): Double = Random.nextDouble(-0.25, 0.25)
 
+@ExperimentalSerializationApi
 private inline fun <reified T> postQuery(
     options: SecretsManagerOptions,
     path: String,
--- a/sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
+++ b/sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
@@ -8,6 +8,7 @@ import kotlinx.serialization.json.JsonObject
 import kotlinx.serialization.json.JsonPrimitive
 import kotlinx.serialization.json.jsonObject
 import kotlinx.serialization.json.jsonPrimitive
+import java.net.HttpURLConnection.HTTP_FORBIDDEN
 import java.net.HttpURLConnection.HTTP_OK
 import java.net.URI
 import java.security.KeyManagementException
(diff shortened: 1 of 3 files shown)


sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
```diff
@@ -1679,6 +1719,26 @@ private inline fun <reified T> postQuery(
...
+            // gated on the 403 status so a non-403 response carrying a {"error":"throttled"} body
+            // is not mistaken for a throttle and retried.
+            if (response.statusCode == HTTP_FORBIDDEN) {
+                val retryAfter = parseThrottle(errorMessage)
+                if (retryAfter != null) {
+                    if (throttleAttempt >= MAX_THROTTLE_RETRIES) {
+                        throw KeeperThrottleException("Request throttled by Keeper backend; exhausted $MAX_THROTTLE_RETRIES retries")
+                    }
+                    val delayMs = throttleDelayMillis(throttleAttempt, retryAfter, throttleJitter())
+                    System.err.println(
```
stas-schaller: The throttle warning bypasses `options.loggingEnabled`. Every other diagnostic print in this file is gated on that flag — callers who set `loggingEnabled = false` to silence SDK output will still see this on stderr.

```kotlin
if (options.loggingEnabled) {
    System.err.println(
        "WARNING: Request throttled (attempt ${throttleAttempt + 1}/$MAX_THROTTLE_RETRIES); " +
            "retrying in ${"%.1f".format(delayMs / 1000.0)}s"
    )
}
```

stas-schaller (changes requested):
Almost there, just one inconsistency and we're good to go.

sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
mgallego-keeper: Good catch — gated the throttle warning on `options.loggingEnabled` so it lines up with the other diagnostics in `postQuery`, and added `ThrottleTest` coverage (`throttleWarning_suppressedWhenLoggingDisabled` / `throttleWarning_emittedWhenLoggingEnabled`) asserting it stays silent when `loggingEnabled = false` and is emitted when enabled. Pushed in 1794251.

sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
```diff
@@ -1662,13 +1672,43 @@ private inline fun <reified T> encryptAndSignPayload(
...
+    val baseSec = if (retryAfter > 0.0) retryAfter else BASE_THROTTLE_DELAY_SEC.toDouble() * (1L shl attempt)
+    val sec = baseSec + baseSec * jitter
+    return (maxOf(sec, 0.0) * 1000).toLong()
+}
+
+// Random jitter multiplier in [-0.25, 0.25). Kept separate so unit tests exercise
+// throttleDelayMillis with a pinned jitter instead.
+internal fun throttleJitter(): Double = Random.nextDouble(-0.25, 0.25)
+
 private inline fun <reified T> postQuery(
```
stas-schaller, line 1778: `postQuery` lost its `@ExperimentalSerializationApi` opt-in. The annotation used to sit directly above `postQuery`, but the three new helpers were inserted between them, so it now binds to `parseThrottle` instead. `postQuery` still calls experimental serialization APIs (`nonStrictJson.decodeFromString`, plus `parseThrottle`), so the compiler now emits three opt-in warnings that weren't on the base branch (lines 1714, 1727, 1745). Let's keep the annotation on `parseThrottle` and add it back onto `postQuery` too. One line, clears all three.

mgallego-keeper: Done — re-added `@ExperimentalSerializationApi` to `postQuery` (kept on `parseThrottle` too, since each carries its own declaration-site opt-in). Clean compile now with no opt-in warnings on those three lines, and all 40 core tests still pass. Pushed in 8916b7a.

stas-schaller (approved):
throttled 👍

Issue #1044, opened 2026-06-17
Python core: KSMCache ignores kms_cache_file_name override after KSM-1004


Wade-0:
## Bug Summary

After PR #1034 (`KSM-1004: resolve KSMCache path lazily so KSM_CACHE_DIR is honored after import`), `KSMCache.kms_cache_file_name` is no longer used by cache operations.

This appears to be a breaking change for consumers that override `KSMCache.kms_cache_file_name` programmatically instead of using `KSM_CACHE_DIR`.

## Current Behavior

`save_cache`, `get_cached_data`, and `remove_cache_file` now call `KSMCache.get_cache_file_path()`, which always resolves the path from `KSM_CACHE_DIR` and the default filename:

```python
os.path.join(os.environ.get("KSM_CACHE_DIR", ""), "ksm_cache.bin")
```

As a result, changing:

```python
KSMCache.kms_cache_file_name = "/custom/path/my_cache.bin"
```

has no effect.

## Expected Behavior

If `kms_cache_file_name` is kept for backward compatibility, cache operations should still honor it.

At minimum, one of these should be true:

1. `kms_cache_file_name` continues to override the cache file path when explicitly set, or
2. the attribute is removed or deprecated clearly as a breaking change and documented accordingly.

## Why this is a problem

The PR description says:

> Keep the `kms_cache_file_name` class attribute for backward compatibility (no longer governs behavior)

But if existing integrations depended on setting that attribute, behavior has changed incompatibly.

## Reproduction

Before PR #1034, code like this worked:

```python
from keeper_secrets_manager_core.core import KSMCache

KSMCache.kms_cache_file_name = "/tmp/custom_cache.bin"
KSMCache.save_cache(b"test")
assert KSMCache.get_cached_data() == b"test"
KSMCache.remove_cache_file()
```

After PR #1034, the override is ignored and the cache is written to the path derived from `KSM_CACHE_DIR` or the default CWD-relative location instead.

## Suggested Fix

Consider updating `get_cache_file_path()` to preserve the legacy override behavior, for example by honoring `kms_cache_file_name` when it has been explicitly changed from the default value.

Alternatively, document this as a breaking change and provide a supported migration path.

## Related

- PR #1034
- Jira: KSM-1004

mgallego-keeper:
Hi @Wade-0,

Thank you for brining this to our attention. We have opened a PR to fix this issue and will be part of our next release.

PR #1046, opened 2026-06-18
fix(sdk/python/core): honor kms_cache_file_name override in KSMCache cache path (KSM-1019)

mgallego-keeper:
**Jira:** [KSM-1019](https://keeper.atlassian.net/browse/KSM-1019) · **Fixes** #1044

## Summary

PR #1034 (KSM-1004) rerouted `KSMCache.save_cache` / `get_cached_data` / `remove_cache_file` through a new `get_cache_file_path()` that re-derives the path from `KSM_CACHE_DIR` and no longer reads the `kms_cache_file_name` class attribute. As a result, assigning `KSMCache.kms_cache_file_name = "/custom/path.bin"` was silently ignored: a backward-incompatible change for consumers that set it (and it was the only way to customize the cache filename, since `KSM_CACHE_DIR` only controls the directory).

## Changes

- **`core.py`**: `get_cache_file_path()` honors an explicitly assigned `kms_cache_file_name`. The override is detected by object identity against an import-time `_DefaultCachePath(str)` sentinel (`_default_cache_file_name`), so it wins even when the assigned text happens to equal the default. The default-path expression lives in a single module-level helper, `_default_cache_path()`, used by both the sentinel and the lazy branch. #1034's lazy `KSM_CACHE_DIR` resolution is preserved; the three cache methods already route through `get_cache_file_path()`, so no caller changes were needed.
- **`tests/cache_test.py`**: five new tests. `test_kms_cache_file_name_override_is_honored` (basic override), `test_override_equal_to_default_text_is_still_honored` (override text collides with the default while `KSM_CACHE_DIR` is set afterward), `test_kms_cache_file_name_override_can_be_restored_to_default` (assigning the sentinel back re-enables lazy resolution), `test_default_path_without_override_or_env_is_cwd` (untouched default resolves to `ksm_cache.bin` in the CWD, end to end through `save_cache`), and `test_default_sentinel_text_matches_lazy_default_path` (sentinel text stays equal to the `_default_cache_path()` output, so a re-inlined copy of the expression fails loudly).
- **`README.md`**: KSM-1019 changelog entry under `### 17.4.0`.

No version bump in this PR: the fix ships with 17.4.0, and the release branch owns that bump.

## Behavior (verified)

| Scenario | Result |
|---|---|
| Explicit `kms_cache_file_name` override | honored (regression fixed) |
| Override text equal to the default, `KSM_CACHE_DIR` set afterward | override still honored (identity detection) |
| Override restored via `kms_cache_file_name = _default_cache_file_name` | lazy `KSM_CACHE_DIR` resolution re-enabled |
| No override, no env var | `ksm_cache.bin` in CWD (unchanged) |
| `KSM_CACHE_DIR` set after import, attribute untouched | honored (KSM-1004 preserved) |

## Testing

`cd sdk/python/core && python -m pytest tests/` at head: **148 passed, 1 skipped** locally (the suite has grown on the rebased v17.4.0 base since the earlier 127/128 counts). The cache tests were additionally run with `KSM_CACHE_DIR` preset at process start to confirm they hold when the env var is already set at import time.

> Note: the `Test-Python` workflow only triggers on PRs targeting `master`, so it will not auto-run against this release branch; the suite was run locally instead.

## Target

`release/sdk/python/core/v17.4.0`; the fix lands in the 17.4.0 release.

[KSM-1019]: https://keeper.atlassian.net/browse/KSM-1019?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sdk/python/core/keeper_secrets_manager_core/core.py
```diff
@@ -1904,14 +1904,20 @@ def inflate_field_value(self, uids, replace_fields):
...
+    # not set, the cache file is created in the current working directory. Assigning
+    # kms_cache_file_name directly still overrides the full path (backward compatibility).
     kms_cache_file_name = os.path.join(os.environ.get("KSM_CACHE_DIR", ""), 'ksm_cache.bin')
+    # Snapshot of the import-time default, used to detect an explicit override above.
+    _default_cache_file_name = kms_cache_file_name
 
     @classmethod
     def get_cache_file_path(cls):
+        # An explicit assignment to kms_cache_file_name takes precedence (backward compat)...
+        if cls.kms_cache_file_name != cls._default_cache_file_name:
```
stas-schaller: This equality check can't tell "explicitly overridden" from "happens to equal the default." Reproduced locally: set `KSMCache.kms_cache_file_name = "ksm_cache.bin"` (same text as the default when `KSM_CACHE_DIR` was unset at import), then set `KSM_CACHE_DIR` afterward. The override gets silently dropped, a narrower version of the bug this PR fixes.

Fix with identity comparison instead of equality, backed by a `str` subclass sentinel. No API change needed:
```python
class _DefaultCachePath(str):
    pass

class KSMCache:
    _default_cache_file_name = _DefaultCachePath(
        os.path.join(os.environ.get("KSM_CACHE_DIR", ""), 'ksm_cache.bin')
    )
    kms_cache_file_name = _default_cache_file_name

    @classmethod
    def get_cache_file_path(cls):
        if cls.kms_cache_file_name is not cls._default_cache_file_name:
            return cls.kms_cache_file_name
        return os.path.join(os.environ.get("KSM_CACHE_DIR", ""), 'ksm_cache.bin')
```
Tested against all four cases (no override, env-dir-only, the collision case, a genuinely distinct override); this passes where `!=` fails on the collision case.

sdk/python/core/tests/cache_test.py
```diff
@@ -39,6 +39,33 @@ def test_ksm_cache_dir_honored_when_set_after_import(self):
                 else:
                     os.environ["KSM_CACHE_DIR"] = original
 
+    def test_kms_cache_file_name_override_is_honored(self):
```
stas-schaller, line 45: This test only exercises the override with `KSM_CACHE_DIR` unset. Missing: override wins when `KSM_CACHE_DIR` is also set and collides with the default name, the exact gap in the current implementation. Add, and it should fail against the current code before the fix lands:
```python
def test_kms_cache_file_name_override_beats_env_var_when_names_collide(self):
    original = KSMCache.kms_cache_file_name
    try:
        with tempfile.TemporaryDirectory() as temp_dir, \
             unittest.mock.patch.dict(os.environ, {"KSM_CACHE_DIR": temp_dir}):
            KSMCache.kms_cache_file_name = "ksm_cache.bin"
            self.assertEqual(KSMCache.get_cache_file_path(), "ksm_cache.bin")
    finally:
        KSMCache.kms_cache_file_name = original
```
Ran this locally against the PR-head package: fails with `AssertionError: '.../tmp_xyz/ksm_cache.bin' != 'ksm_cache.bin'`.

stas-schaller (changes requested):
Good work, two things before we go forward: `README.md` still ends at `### 17.3.0`, needs a `### 17.3.1` entry.

sdk/python/core/keeper_secrets_manager_core/core.py
mgallego-keeper: Good catch — fixed in 6e590b0. Switched to identity comparison against a `_DefaultCachePath(str)` sentinel as suggested: `_default_cache_file_name` is the sentinel and `kms_cache_file_name` is assigned *from* it (same object at import), so an explicit assignment — always a distinct object — is detected via `cls.kms_cache_file_name is not cls._default_cache_file_name` and honored even when its text equals the default. Being a `str` subclass, it behaves as an ordinary path everywhere else.

sdk/python/core/tests/cache_test.py
mgallego-keeper: Added in 6e590b0 as `test_override_equal_to_default_text_is_still_honored`. I kept it in this file's existing manual `try/finally` + `os.environ` idiom (matching the other two tests) rather than `patch.dict`, since `patch.dict` only manages the env var and we'd still save/restore the class attribute by hand — this way avoids a new import. It derives the colliding value from `str(KSMCache._default_cache_file_name)`, so it holds regardless of whether `KSM_CACHE_DIR` happens to be set at import.

Verified fail-then-pass: against the pre-fix `!=` code it fails with `AssertionError: 'ksm_cache.bin' != 'some/other/dir/ksm_cache.bin'`, and passes after the identity fix.

mgallego-keeper:
Thanks for the review — all three points addressed in 6e590b0:

- **core.py** — override detection now uses object identity against a `_DefaultCachePath(str)` sentinel instead of value-equality, so an override whose text equals the default is no longer silently dropped.
- **cache_test.py** — added the collision-case regression test (fails on the old `!=` code, passes after the fix).
- **README.md** — added the `### 17.3.1` changelog entry above `### 17.3.0`.

Full suite green locally: `python -m pytest tests/` → 128 passed, 1 skipped. Re-requesting review.

sdk/python/core/keeper_secrets_manager_core/core.py
```diff
@@ -1903,15 +1903,35 @@ def inflate_field_value(self, uids, replace_fields):
...
+    kms_cache_file_name = _default_cache_file_name
 
     @classmethod
     def get_cache_file_path(cls):
+        # An explicit assignment to kms_cache_file_name takes precedence (backward compat);
+        # detected by identity so a same-text override is not mistaken for the default.
+        if cls.kms_cache_file_name is not cls._default_cache_file_name:
+            return cls.kms_cache_file_name
+        # Otherwise resolve the directory from KSM_CACHE_DIR at call time.
         return os.path.join(os.environ.get("KSM_CACHE_DIR", ""), 'ksm_cache.bin')
```
stas-schaller: This expression is identical to the one on line 1924 that builds `_default_cache_file_name`. The identity check works because the sentinel and the lazy branch are two separate objects, but the correctness of that check depends on both expressions producing the same text. Change the filename in one place and they silently disagree, with no test catching the drift.

Extract a module-level helper so there is one source of truth:

```python
def _default_cache_path():
    return os.path.join(os.environ.get("KSM_CACHE_DIR", ""), 'ksm_cache.bin')

class _DefaultCachePath(str):
    pass

class KSMCache:
    _default_cache_file_name = _DefaultCachePath(_default_cache_path())
    kms_cache_file_name = _default_cache_file_name

    @classmethod
    def get_cache_file_path(cls):
        if cls.kms_cache_file_name is not cls._default_cache_file_name:
            return cls.kms_cache_file_name
        return _default_cache_path()
```

sdk/python/core/keeper_secrets_manager_core/core.py
```diff
@@ -1903,15 +1903,35 @@ def inflate_field_value(self, uids, replace_fields):
...
-    # lazily from KSM_CACHE_DIR at call time (see get_cache_file_path), so setting the
-    # env var after import is honored. If not set, the cache file is created in the
-    # current working directory.
-    kms_cache_file_name = os.path.join(os.environ.get("KSM_CACHE_DIR", ""), 'ksm_cache.bin')
+    # Import-time default cache path. The directory is read lazily from KSM_CACHE_DIR at
+    # call time (see get_cache_file_path), so setting the env var after import is honored;
+    # if unset, the cache file is created in the current working directory. Assigning
+    # kms_cache_file_name directly overrides the full path (backward compatibility); the
+    # override is detected by object identity against this sentinel, so it holds even when
+    # the assigned value equals the default text. (Re-assign _default_cache_file_name to
```
stas-schaller: The note `(Re-assign _default_cache_file_name to restore the default.)` reads as an instruction to modify `_default_cache_file_name` itself, which would corrupt the sentinel for every future caller. The intent is the opposite: reassign `kms_cache_file_name` back to the sentinel.

Change to: `(To restore the default, assign KSMCache.kms_cache_file_name = KSMCache._default_cache_file_name.)`

Also worth adding a test that exercises this restore path, since the comment describes it and nothing currently locks in the correct way to do it:

```python
def test_kms_cache_file_name_override_can_be_restored_to_default(self):
    original = KSMCache.kms_cache_file_name
    KSMCache.kms_cache_file_name = "/some/custom/path.bin"
    KSMCache.kms_cache_file_name = KSMCache._default_cache_file_name
    try:
        # After restore, lazy env resolution must work again.
        with unittest.mock.patch.dict(os.environ, {"KSM_CACHE_DIR": "/lazy/dir"}):
            self.assertEqual(
                KSMCache.get_cache_file_path(),
                "/lazy/dir/ksm_cache.bin",
                "restoring kms_cache_file_name to _default_cache_file_name should re-enable lazy KSM_CACHE_DIR resolution",
            )
    finally:
        KSMCache.kms_cache_file_name = original
```

stas-schaller (changes requested):
Good progress on the identity-sentinel approach, and both of my previous comments are addressed. Three more things before this can merge.

**1. Rewrite the PR description.** The body still describes a 17.3.1 point release with a version bump across `_version.py`, `keeper_globals.py`, and `smoke_test.py`. None of that is in this PR — the diff is two files and the version reads 17.3.0 at head. The PR was retargeted to `release/sdk/python/core/v17.4.0` and the last commit deliberately dropped the version bump, which is the right call, but the description was never updated to match. Please rewrite it to reflect what the PR actually does: fix lands in 17.4.0, no version bump included here. Also correct the test count from 127 to 128 (the second test was added after the description was written).

**2. Extract the duplicated default-path expression** (inline comment below).

**3. Fix the misleading restore-default note and add a restore test** (inline comment below).

sdk/python/core/keeper_secrets_manager_core/core.py
mgallego-keeper: Extracted in 3274cddd as module-level `_default_cache_path()`, used by both the sentinel construction and the lazy branch, matching your snippet. Its comment states the invariant explicitly (the sentinel and the lazy branch must produce the same text), so the single-source requirement is documented at the definition.

sdk/python/core/keeper_secrets_manager_core/core.py
mgallego-keeper: Both in 3274cddd. The note now reads: `(To restore the default, assign KSMCache.kms_cache_file_name = KSMCache._default_cache_file_name.)` And added `test_kms_cache_file_name_override_can_be_restored_to_default`, which asserts that assigning the sentinel back re-enables lazy `KSM_CACHE_DIR` resolution.

Two small deviations from the suggested snippet: it keeps this file's manual try/finally env save/restore idiom (same as the collision test), and it builds the expected path with `os.path.join("lazy", "dir")` instead of the literal `"/lazy/dir/ksm_cache.bin"`, which would fail on Windows since `os.path.join` joins with a backslash there.

mgallego-keeper:
Thanks for the review, all three items addressed:

1. **PR description rewritten** to match the current state: target is `release/sdk/python/core/v17.4.0`, no version bump in this PR, the actual diff (`core.py`, `cache_test.py`, README 17.4.0 changelog entry), and the real test count at head (146 passed, 1 skipped; the suite has grown on the rebased base since the 127/128 counts).
2. **`_default_cache_path()` extracted** as the single source of the default-path expression (replied inline).
3. **Restore note fixed and restore test added** (replied inline).

Also removed a stray blank line the dropped 17.3.1 section had left in the README, so the README diff is now just the changelog entry. All code changes in 3274cddd. Re-requesting review.

mgallego-keeper:
Follow-up commit 4e4c013d: did a coverage pass over the path-resolution matrix and closed two gaps, both test-only:

- `test_default_path_without_override_or_env_is_cwd`: the untouched default (no override, no `KSM_CACHE_DIR`) resolving to `ksm_cache.bin` in the current working directory, exercised end to end through `save_cache`. Every other test set an env var or an override, so the plain-default row of the behavior table had no lock.
- `test_default_sentinel_text_matches_lazy_default_path`: asserts the sentinel's text equals `_default_cache_path()` output. Resolution ignores the sentinel's text (identity only), but `kms_cache_file_name` is a public attribute consumers read directly (the Ansible plugin assigns it, for example), so if either side is ever re-inlined and drifts, a test fails instead of the attribute silently advertising a path the cache methods never use. This complements the helper extraction from the review: the helper removes the duplication, the test makes reintroducing it loud.

Suite at head: 148 passed, 1 skipped. The cache tests were also run with `KSM_CACHE_DIR` preset at process start to confirm they hold when the env var is already set at import. Description updated to match.

stas-schaller (approved):
All three items from the last round are addressed: PR description matches the current diff, `_default_cache_path()` is the single source of truth for the default-path expression, and the restore note is corrected with a locking test. The sentinel pattern and identity check are correct; the Ansible integration was verified to be unaffected. Good to merge.

PR #1049, opened 2026-06-26
Release Java SDK v17.3.0

### LINKED ISSUE #973: Keeper Java/Kotlin: Field 'lastModified' is required for type with serial name...

stas-schaller:
## Summary
Release branch for Java SDK v17.3.0 — eleven changes: throttle retry backoff with one-sided jitter and a retry-after cap, KeeperRecordLink accessor alignment, IL5 region support, a public SDK exception type, serialization correctness for record create, file attachment crash fixes, shared folder record decryption, null file URL handling, file descriptor leak fixes, and Base64 null guard.

> Continues from #975. That PR targeted `release/sdk/java/core/v17.2.1`; the release was promoted to a minor bump (17.2.1 → 17.3.0) because it carries a new feature (IL5 dynamic key, KSM-902) alongside the fixes. Same commits, retargeted to `release/sdk/java/core/v17.3.0` ([REL-5094](https://keeper.atlassian.net/browse/REL-5094)).

### New Features
- **Throttle retry with exponential backoff** (KSM-878, refined by KSM-1032): `postQuery` now automatically retries HTTP 403 `{"error":"throttled"}` responses up to 5 times. Backoff starts at 11s (1s margin over the backend's 10s memcached TTL) with one-sided 0-25% jitter — a delay is only ever padded up, never undercuts the server's requested `retry_after` or the exponential floor; `retry_after` from the response body is honored when present, capped at 176s (`MAX_THROTTLE_DELAY_SEC`) so a misbehaving response can't force an excessive wait. Each retry logs a warning to stderr (gated on `loggingEnabled`). Exhausted retries throw `KeeperThrottleException` (public, catchable separately from `Exception`). `SecretsManagerOptions` accepts an injectable `throttleSleepMillis` hook for test overrides.
- **`KeeperRecordLink` accessor alignment with Python reference** (KSM-1008): recursive `jsonElementToValue`/`jsonObjectToMap` helpers preserve JSON nulls and typed scalars (String/Boolean/Int/Long/Double) across nested objects and arrays. `getBooleanValue` now supports an optional `checkAllowedSettings` parameter to look up permission flags nested under `allowedSettings` in `path:"meta"` links; `allowsRotation()` and related accessors updated accordingly. Added `KeeperRecordLinkTest` suite covering all accessor methods, nested structures, and null preservation.
- **IL5 region mapping** (KSM-902): added `"IL5" -> "il5.keepersecurity.us"` to the `initializeStorage()` region map in `SecretsManager.kt`. Tokens prefixed `IL5:` now route to the DoD Impact Level 5 environment. Extended OTT tokens are strictly validated to 2 or 4 segments; a malformed 3-segment token is now rejected rather than silently parsed as a 2-segment token with the third segment dropped. Region-prefix matching (`tokenParts[0].uppercase(...)`) is pinned to `Locale.US` — on a Turkish-locale JVM, `"il5".uppercase(Locale.getDefault())` produces `"İL5"` (dotted capital I), which would fail to match `"IL5"`.
- **Public SDK exception type** (KSM-1026): `SecretsManagerException` changed from `internal` to a public `open class`, so Java consumers can `catch (SecretsManagerException e)` to handle SDK-level errors by type instead of catching bare `Exception`. `SecureRandomException`/`SecureRandomSlowGenerationException` remain internal (no public use case identified).

### Bug Fixes
- **Custom field omitted from record create payload** (KSM-823): `KeeperRecordData.custom` defaulted to `null`, causing `kotlinx-serialization` to omit `"custom"` from the V3 API payload when no custom fields were set. Changed default to `mutableListOf()` and added `@EncodeDefault` to force inclusion. Regression test added.
- **KeeperFileData crash when `lastModified` absent** (KSM-854): files uploaded by non-SDK clients (iOS, Android, Web Vault) omit `lastModified`; `kotlinx-serialization` treated it as required, throwing `MissingFieldException` and silently dropping the attachment. Added `= 0` default, consistent with .NET SDK behavior (KSM-674).
- **Shared folder records return null fields** (KSM-753): records created via Commander/PowerShell in shared folders appear in `response.records[]` with `innerFolderUid` set, but their `recordKey` is encrypted with the folder key, not the app key. Now builds a folder-key map before processing flat records and uses the correct key when `innerFolderUid` matches.
- **NPE crash when file `url` absent** (KSM-765): `KeeperFile.url` and `SecretsManagerResponseFile.url` changed from `String` to `String?`; `downloadFile()` now throws a typed exception when `url` is null rather than crashing with an NPE.
- **File descriptor leaks** (KSM-855): `LocalConfigStorage` replaced four manual `stream.close()` patterns with Kotlin `.use { }` — the init block never closed its reader at all. `downloadFile`, `uploadFile`, and `postFunction` now wrap `HttpsURLConnection` in `try/finally` + `disconnect()` — `with()` is a scope function, not try-with-resources, and did not guarantee cleanup on exception.
- **Opaque NPE from Base64 decoders** (KSM-985): `base64ToBytes("")` and `webSafe64ToBytes("")` now throw a typed Keeper exception on empty input instead of an NPE from inside `java.util.Base64` (KSM-808 parity).

### Breaking Changes
- `KeeperFile.url` changed from `String` to `String?`. Callers that previously treated `url` as non-null must now handle the nullable case. This is a correctness fix — the server can omit this field and the previous behavior was an NPE crash.
- `KeeperRecordData.custom` changed from `MutableList<KeeperRecordField>?` to a non-null `MutableList<KeeperRecordField>` (KSM-823). Assigning `null` no longer compiles; use an empty list instead.
- `KEY_SERVER_PUBIC_KEY_ID` (the misspelled constant published in 17.2.0) is deprecated in favor of `KEY_SERVER_PUBLIC_KEY_ID`. The old name is retained as a `@Deprecated` alias so existing source and bytecode keep working; the stored config value (`"serverPublicKeyId"`) is unchanged.

## Related Issues
- Closes #973
- Supersedes #975
- KSM-878, KSM-1008, KSM-902, KSM-1026, KSM-823, KSM-854, KSM-753, KSM-765, KSM-855, KSM-985, KSM-1032

[REL-5094]: https://keeper.atlassian.net/browse/REL-5094?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/publish.maven.java.core.yml
+++ b/.github/workflows/publish.maven.java.core.yml
@@ -1,13 +1,22 @@
 name: Publish to Maven (Java Core SDK)
 on:
   workflow_dispatch:
+    inputs:
+      publish:
+        description: 'Publish to Maven Central (uncheck to build only)'
+        required: false
+        default: 'true'
+        type: boolean
 
 jobs:
   publish:
+    permissions:
+      contents: read
     uses: ./.github/workflows/reusable.maven.central.publish.yml
     with:
       working-directory: ./sdk/java/core
       project-name: keeper-secrets-manager-java
       project-title: Keeper Secrets Manager Java Core SDK
       java-version: '11'
+      publish: ${{ inputs.publish }}
     secrets: inherit
--- a/.github/workflows/reusable.maven.central.publish.yml
+++ b/.github/workflows/reusable.maven.central.publish.yml
@@ -26,9 +26,13 @@ on:
         type: boolean
         default: true
 
+permissions: {}
+
 jobs:
   get-version:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     defaults:
       run:
         working-directory: ${{ inputs.working-directory }}
@@ -87,6 +91,8 @@ jobs:
   test:
     needs: validate-version
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     defaults:
       run:
         working-directory: ${{ inputs.working-directory }}
@@ -109,6 +115,8 @@ jobs:
   generate-and-upload-sbom:
     needs: [get-version, validate-version]
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     defaults:
       run:
         working-directory: ${{ inputs.working-directory }}
@@ -162,6 +170,8 @@ jobs:
     if: ${{ inputs.publish == true }}
     environment: prod
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
 
     defaults:
       run:
--- a/.github/workflows/reusable.maven.central.publish.yml
+++ b/.github/workflows/reusable.maven.central.publish.yml
@@ -159,6 +159,7 @@ jobs:
 
   publish-to-maven-central:
     needs: [get-version, test, generate-and-upload-sbom, confirm-publish]
+    if: ${{ inputs.publish == true }}
     environment: prod
     runs-on: ubuntu-latest
 
--- a/.github/workflows/reusable.maven.central.publish.yml
+++ b/.github/workflows/reusable.maven.central.publish.yml
@@ -44,7 +44,7 @@ jobs:
         env:
           PROJECT_NAME: ${{ inputs.project-name }}
         run: |
-          VERSION=$(grep -Po 'version\s*=\s*"\K[^"]*' build.gradle.kts || echo "0.0.0-unknown")
+          VERSION=$(grep -Po '^version\s*=\s*"\K[^"]*' build.gradle.kts || echo "0.0.0-unknown")
           echo "Version retrieved: $VERSION"
           echo "version=$VERSION" >> "$GITHUB_OUTPUT"
 
@@ -66,8 +66,48 @@ jobs:
           echo "Artifact ID retrieved: $ARTIFACT_ID"
           echo "artifact-id=$ARTIFACT_ID" >> "$GITHUB_OUTPUT"
 
-  generate-and-upload-sbom:
+  validate-version:
     needs: get-version
+    permissions: {}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Verify version not already published on Maven Central
+        env:
+          VERSION: ${{ needs.get-version.outputs.version }}
+          ARTIFACT_ID: ${{ needs.get-version.outputs.artifact-id }}
+        run: |
+          GROUP_PATH="com/keepersecurity/secrets-manager"
+          POM_URL="https://repo1.maven.org/maven2/${GROUP_PATH}/${ARTIFACT_ID}/${VERSION}/${ARTIFACT_ID}-${VERSION}.pom"
+          if curl -sf "$POM_URL" > /dev/null 2>&1; then
+            echo "ERROR: ${ARTIFACT_ID} ${VERSION} already exists on Maven Central"
+            exit 1
+          fi
+          echo "Version ${VERSION} not yet published — OK to proceed"
+
+  test:
+    needs: validate-version
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ${{ inputs.working-directory }}
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+      - name: Set up Java ${{ inputs.java-version }}
+        uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # v5.2.0
+        with:
+          java-version: ${{ inputs.java-version }}
+          distribution: 'temurin'
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@50e97c2cd7a37755bbfafc9c5b7cafaece252f6e # v6.1.0
+        with:
+          gradle-version: '8.14'
+      - name: Build and Test
+        run: ./gradlew clean build --no-daemon
+
+  generate-and-upload-sbom:
+    needs: [get-version, validate-version]
     runs-on: ubuntu-latest
     defaults:
       run:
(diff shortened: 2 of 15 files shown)


mgallego-keeper:
## Release review — Java SDK v17.3.0 (REL-5094)

Audited all 9 KSM tickets linked to REL-5094 against this PR and reviewed the full diff (13 files, +1,150/−252) and commit history.

**Ticket coverage** — all 9 implemented and in the 17.3.0 changelog: KSM-902, 878, 1008, 823, 854, 765, 855, 985, 753. CI green across JDK 8/11/17/21 + CodeQL + Socket. (Note: PR targets `master` and is currently blocked on a required approval.)

### Strengths
- Strong tests: `ThrottleTest` (unit + e2e) and `KeeperRecordLinkTest` (17 cases mirroring the Python reference).
- Throttle path is 403-gated, runs **before** key-rotation, honors `retry_after`, applies jitter, throws `KeeperThrottleException` on exhaustion, with an injectable sleep seam.
- IL5 3-layer key injection (config → extended OTT → constructor) with rotation-overwrite suppression is well-reasoned.
- Real CI hardening: SHA-pinned actions, `persist-credentials: false`, CycloneDX SBOM, and the `base64 -w 0` auth-token fix (Linux `base64` wraps at 76 chars and silently corrupts the Bearer header).

### Findings
1. **[Med] Public API break:** `KEY_SERVER_PUBIC_KEY_ID` → `KEY_SERVER_PUBLIC_KEY_ID`. The typo'd constant is a public top-level `const val` shipped in **17.2.0** (it's on `master`), so the rename removes a published symbol — Kotlin source break / `NoSuchFieldError` for old bytecode — in a minor bump. (The commit message says it was "introduced in this release branch", but it predates this branch.) The storage *value* `"serverPublicKeyId"` is unchanged, so configs still work. Suggest keeping a deprecated alias:
   ```kotlin
   @Deprecated("Typo; use KEY_SERVER_PUBLIC_KEY_ID", ReplaceWith("KEY_SERVER_PUBLIC_KEY_ID"))
   const val KEY_SERVER_PUBIC_KEY_ID = KEY_SERVER_PUBLIC_KEY_ID
   ```
2. **[Low-Med] Other source-breaking changes:** `KeeperRecordData.custom` (nullable → non-null) and `KeeperFile.url` (`String` → `String?`). Both are needed for the fixes — recommend calling them out as breaking in the changelog.
3. **[Low-Med] Test gap (KSM-753):** the shared-folder flat-record decryption has no dedicated test. Logic is correct (matches the existing folder-path code at `SecretsManager.kt:1276-1297`), but a regression test — flat `records[]` entry with `innerFolderUid` set and a folder-key-encrypted `recordKey`, asserting fields decrypt — would lock it in.
4. **[Low] CodeQL #215/216/219:** the `test` and `generate-and-upload-sbom` jobs in `reusable.maven.central.publish.yml` lack explicit `permissions:` blocks — inconsistent with the rest of the PR's hardening. Suggest top-level `permissions: {}` + per-job `contents: read`.
5. **[Low] KSM-985 nuance:** the guard is on empty-string (not null), and `base64ToBytes("")` now throws instead of returning empty bytes — intended/tested; just confirm no caller relied on the old behavior.
6. **[Nits]** a 3-segment OTT is silently treated as 2-segment; `"%.1f".format()` in the throttle warning is locale-sensitive (stderr only).

Nice work overall — functionally complete and well-tested. Finding #1 is the main one worth a 1-line fix before merge.

mgallego-keeper (dismissed):
## Follow-up review — Approving ✅

Re-reviewed the 6 commits pushed since my last review (`ebc7c4e2..0467110a`). They resolve the items I raised, and CI is green (18/18) on `0467110a`. Approving.

**Resolved**
- **`KEY_SERVER_PUBIC_KEY_ID` API break** (`ca4d6483`) — restored as a `@Deprecated` alias to `KEY_SERVER_PUBLIC_KEY_ID`; the old symbol links again and the storage value is unchanged. Correct fix, and I confirmed no internal code still references the deprecated name.
- **Breaking changes documented** (`f0fdfea2`) — changelog now calls out `KeeperFile.url` → `String?` and `KeeperRecordData.custom` → non-null. (Minor: the constant deprecation is listed under "Breaking Changes" but is actually back-compat-preserving — reads fine either way.)
- **KSM-753 regression test** (`5fa2aae5`) — solid end-to-end test; it asserts the flat `innerFolderUid` record decrypts with the folder key, and would fail (GCM auth error → record skipped) if the logic regressed to the app key.
- **Workflow permissions** (`07a83530`) — top-level `permissions: {}` + per-job `contents: read`; the CodeQL check is now green.
- **3-segment OTT + throttle-log locale** (`4f6b2161`, `0467110a`) — malformed 3-segment tokens are now rejected loudly (new `testExtendedOttRejectsThreeSegments`), and the retry log is pinned to `Locale.US`. The unified `when (tokenParts.size)` is cleaner and correctly scoped — 1-segment tokens are handled earlier, so `else -> throw` only fires for 3/5/6+ segments.

**One non-blocking follow-up (pre-existing — not introduced by this push):**
Region matching in `initializeStorage` uses `tokenParts[0].uppercase(Locale.getDefault())`. On a Turkish-locale JVM, `"il5".uppercase()` → `"İL5"` (dotted capital I), which would not match `"IL5"` — the same locale-sensitivity class that `4f6b2161` just fixed for the throttle log. Recommend `Locale.ROOT` (or `Locale.US`) here in a future change. Impact is limited to lowercase region prefixes on non-US-locale hosts, so it's not a blocker for this release.

Nice work addressing the review — LGTM.

maksimu (approved):

PR #1037, opened 2026-06-11
feat(sdk/rust): KSM-1009 align KeeperRecordLink accessors with Python reference

mgallego-keeper:
## Summary

Aligns the Rust SDK's typed `KeeperRecordLink` linked-credential accessor layer (introduced by KSM-997) with the Python SDK reference implementation ([KSM-992](https://keeper.atlassian.net/browse/KSM-992), PR #1036), under epic [KSM-1009](https://keeper.atlassian.net/browse/KSM-1009) / KSM-1007. The KSM-997 layer mirrored the Java API, which is stale vs. today's backend: its permission booleans read top-level keys only, while the backend nests them under `allowedSettings` in `path:"meta"` links — so they returned `false` on all current data, and the `meta` path plus several credential-link fields were unmapped.

Reshaping the struct is safe: KSM-997 is unreleased (pending in release PR #1030, which this PR targets).

## Changes

- **Permission booleans read `allowedSettings` as a fallback** (top-level key wins): `allows_rotation()`, `allows_connections()`, `allows_port_forwards()`, `allows_session_recording()`, `allows_typescript_recording()`, `allows_remote_browser_isolation()` now return correct values on current `meta` links.
- **New accessors** (mirroring Python): `is_iam_user()`, `belongs_to()`, `no_update_services()`, `ai_enabled()`, `ai_session_terminate()`, `get_allowed_settings()`, `get_rotation_settings()`, `get_meta_data()` (path-gated to `"meta"`).
- **Lossless retention**: each typed link keeps the untouched original entry in `.raw` (`#[serde(skip)]`, populated by `from_map`); `KeeperRecordLink::new(record_uid, data, path)` constructs one from parts. Unknown payload fields already pass through `get_link_data()`.
- **Doc comments** updated to the live-verified payload shapes (`meta`/credential/`ai_settings`/`jit_settings`) instead of the stale Java-derived field lists.
- Raw `Record.links` field and `Record::get_links()` behavior unchanged (entries without `recordUid` skipped); no protocol changes.

## Verified backend contract

Same live QA PAM app used for the Python reference (24 links across pamMachine/pamDatabase/pamRemoteBrowser/pamUser/pamNetworkConfiguration). Links carry exactly `{recordUid, data: base64|null, path: string|null}`:

| `path` | Encoding | Payload |
|---|---|---|
[...]

PR #1050, opened 2026-06-26
feat(sdk/java): align KeeperRecordLink accessors with Python reference (KSM-1008)

mgallego-keeper:
## Summary

`KeeperRecordLink`'s accessors were stale against the live backend. The PAM permission flags now arrive nested under `allowedSettings` on `path:"meta"` self-links, but the booleans (`allowsRotation()`, …) only read top-level keys — so they returned `false` on all current `meta` data. This aligns the Java accessors with the Python reference (KSM-992, #1036), already shipped for Rust (KSM-1009, #1037) and JavaScript (KSM-1010).

**Purely additive — no protocol/wire changes; raw `links` deserialization is unchanged.**

## What changes

- **`allowedSettings` fallback (top-level wins).** `allowsRotation`, `allowsConnections`, `allowsPortForwards`, `allowsSessionRecording`, `allowsTypescriptRecording`, `allowsRemoteBrowserIsolation` now read the top-level key first, then fall back to the nested `allowedSettings` object.
- **New accessors:** `isIamUser()`, `belongsTo()`, `noUpdateServices()`, `aiEnabled()`, `aiSessionTerminate()`, `getAllowedSettings()`, `getRotationSettings()`, and `getMetaData()` (path-gated to `"meta"`).
- **Lossless `getLinkData()`.** A single shared recursive JSON converter (objects → `Map`, arrays → `List`, primitives → typed scalars) replaces the previous `value.toString()` stringification, so nested objects/arrays are preserved. `getLinkData()` also falls through to decryption when ciphertext coincidentally starts with `{`/`[`.
- **KDoc refresh** for the class and `getAiSettingsData`/`getJitSettingsData` to the live-verified payload shapes (drops the stale "as of SDK v17.1.0" field lists).
- Strict typing preserved: quoted `"3"`/`"true"` are not coerced to int/bool.

## Why it matters

On current data, every PAM permission boolean returns `false` because the flags moved into `allowedSettings` — callers can't tell what a linked credential is actually allowed to do. This fixes that and adds the credential/meta/AI accessors the other SDKs already expose.

## Tests

New `KeeperRecordLinkTest` (17 cases, mirroring Python `tests/record_link_test.py`); existing suites unaffected. `./gradlew test` → **46 tests, 0 failures** (KeeperRecordLinkTest 17/17, SecretsManagerTest 14/14).

| Test | Asserts |
|------|---------|
| `metaLinkLiveShape` | all eight permission booleans read via the `allowedSettings` fallback on a `path:"meta"` payload; `version` / `no_update_services` / `getAllowedSettings` / `getMetaData` |
| `topLevelWinsOverAllowedSettings` | a top-level boolean overrides `allowedSettings`; fallback applies only when the key is absent |
| `credentialLinkLiveShape` | `is_admin` / `is_iam_user` / `belongs_to` / `is_launch_credential` + nested `rotation_settings`; absent → null |
| `dataLessReferenceLink` | pure record reference (no data) → every accessor null/false/empty |
| `aiSettingsLiveShape`, `jitSettingsLiveShape` | encrypted payloads decrypt; nested structures preserved; string `version` ⇒ `getLinkDataVersion()` null |
| `losslessnessPreservesNestedStructures` | `getLinkData()` returns nested objects/arrays as `Map`/`List`, not strings |
| `ciphertextWithJsonLikeFirstByte` | ciphertext starting with `{`/`[` falls through to decryption; no key ⇒ null |
| + 10 more | boolean defaults, strict typing / no coercion, version & decoded-data, `mightBeEncrypted` by path, decrypt roundtrip + wrong key, plain-vs-encrypted, path-gating, encrypted-data detection |

## Follow-ups (not in this PR)

- .NET (KSM-1011) and Ruby (KSM-1013) still need the same accessor update; both are in Triage.

Jira: KSM-1008

### DIFF
--- a/sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
+++ b/sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
@@ -241,7 +241,7 @@ data class KeeperRecordLink(
     /**
      * Parse the link data as a JSON object, handling errors gracefully
      */
-    private fun parseJsonData(): Map<String, Any>? {
+    private fun parseJsonData(): Map<String, Any?>? {
         if (data == null) return null
         
         return try {
@@ -312,14 +312,14 @@ data class KeeperRecordLink(
 
     /**
      * Recursively convert a [JsonElement] to a plain Kotlin value: objects become
-     * `Map<String, Any>`, arrays become `List<Any>`, and primitives become typed scalars
+     * `Map<String, Any?>`, arrays become `List<Any?>`, and primitives become typed scalars
      * (String/Boolean/Int/Long/Double). Strings are never coerced, so a quoted "3"/"true"
-     * stays a String. JSON nulls are dropped by [jsonObjectToMap].
+     * stays a String. JSON nulls become Kotlin null and are preserved on both branches.
      */
     private fun jsonElementToValue(element: JsonElement): Any? = when (element) {
         is JsonNull -> null
         is JsonObject -> jsonObjectToMap(element)
-        is JsonArray -> element.mapNotNull { jsonElementToValue(it) }
+        is JsonArray -> element.map { jsonElementToValue(it) }
         is JsonPrimitive ->
             if (element.isString) element.content
             else element.booleanOrNull
@@ -330,11 +330,11 @@ data class KeeperRecordLink(
     }
 
     /**
-     * Convert a [JsonObject] to a nested `Map<String, Any>`, preserving nested objects and arrays
-     * (lossless). Keys whose value is JSON null are omitted.
+     * Convert a [JsonObject] to a nested `Map<String, Any?>`, preserving nested objects, arrays and
+     * JSON null values (lossless, matching the Python reference).
      */
-    private fun jsonObjectToMap(obj: JsonObject): Map<String, Any> =
-        obj.entries.mapNotNull { (key, value) -> jsonElementToValue(value)?.let { key to it } }.toMap()
+    private fun jsonObjectToMap(obj: JsonObject): Map<String, Any?> =
+        obj.entries.associate { (key, value) -> key to jsonElementToValue(value) }
 
     /**
      * Check if the link data indicates admin status for a user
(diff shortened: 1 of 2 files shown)


sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
```diff
@@ -288,6 +310,32 @@ data class KeeperRecordLink(
...
+    /**
+     * Recursively convert a [JsonElement] to a plain Kotlin value: objects become
+     * `Map<String, Any>`, arrays become `List<Any>`, and primitives become typed scalars
+     * (String/Boolean/Int/Long/Double). Strings are never coerced, so a quoted "3"/"true"
+     * stays a String. JSON nulls are dropped by [jsonObjectToMap].
+     */
+    private fun jsonElementToValue(element: JsonElement): Any? = when (element) {
+        is JsonNull -> null
+        is JsonObject -> jsonObjectToMap(element)
+        is JsonArray -> element.mapNotNull { jsonElementToValue(it) }
```
stas-schaller: The converter drops JSON nulls, so it isn't lossless, and it diverges from the Python reference, which keeps them. `mapNotNull` here removes null array elements (shrinking length and shifting indices); `jsonObjectToMap` (line 337) drops null-valued keys. This contradicts the "(lossless)" claim in that same KDoc (line 334) and "the complete parsed payload ... preserved ... retained" in the class KDoc (lines 231-232). Verified by running it:

```
{"a": null, "b": 1}    -> {b=1}
{"list": [1, null, 2]} -> [1, 2]
```

Python's `json.loads` yields `{'a': None, 'b': 1}` and `[1, None, 2]`. Suggest widening the value type to `Any?` and using `map`/`associate` instead of `mapNotNull` so nulls survive on both branches. If dropping nulls is intentional instead, drop the "lossless" wording from the KDoc and the `losslessnessPreservesNestedStructures` test name.

stas-schaller (changes requested):
One issue, then this is good to merge. Also filed KSM-1031 for a separate pre-existing UTF-8 charset bug in the same class (not part of this PR).

stas-schaller (approved):
Aligned

stas-schaller (approved):
Fix looks good!

PR #1052, opened 2026-07-07
feat(servicenow)!: FQCN resolver for Xanadu+ and KSM SDK pin for PAM records

mgallego-keeper:
## Summary
Fixes the ServiceNow escalation (IMP-3033, TX Dept. of Public Safety, on Zurich): the resolver failed with `Serializer for subclass 'pamSettings' is not found in the polymorphic scope of 'KeeperRecordField'` whenever PAM records were shared to the KSM application. Also delivers the move to a Keeper-owned FQCN so the resolver can coexist with other vendors' resolvers on Xanadu+ MID Servers.

## What changed
- **SDK fix:** pin KSM Java SDK `core` to **17.2.0** (was `16.6.4+`, a Gradle prefix wildcard stuck at 16.6.4, predating the `pamSettings` field type). 17.x registers all PAM field types and skips unparseable records on a full fetch instead of failing the whole batch. Bump to 17.3.0 once it publishes to Maven Central.
- **FQCN:** implementation moved to `com.keepersecurity.discovery.KeeperCredentialResolver`; `com.snc.discovery.CredentialResolver` reduced to a thin backward-compatible shim.
- **Two JAR variants** (`-PresolverVariant`): `fqcn` (Xanadu+, ships only the Keeper FQCN class) and `legacy` (pre-Xanadu, keeps `com.snc.discovery.CredentialResolver`).
- **Tests:** extracted MID-free logic into `KeeperCredentialMapper` + JUnit tests including a `pamSettings` regression.
- **Release workflow:** build utah/vancouver/washingtondc (legacy) and xanadu/yokohama/zurich (fqcn) under one v1.0.0 release; real Zurich/Yokohama MID build URLs; robust source-jar path and re-enabled/fixed version-in-tag check; dropped rome/sandiego/tokyo (past support).
- **Docs:** README FQCN-per-release guidance, compatibility matrix, cache/PAM notes, `pamSettings` troubleshooting. docs.keeper.io update tracked separately.

## Testing
Built and tested locally against SDK 17.3.0 and 17.2.0 — 13/13 unit tests pass on both. Verified the `fqcn` JAR excludes `com/snc/discovery/**` while the `legacy` JAR includes it, and `PamSettings` is bundled in both.

## Supported ServiceNow releases (v1.0.0)
| Release | Variant | Class / FQCN to configure |
| --- | --- | --- |
| Utah / Vancouver / Washington DC | legacy | `com.snc.discovery.CredentialResolver` |
| Xanadu / Yokohama (Patch 7+) / Zurich | fqcn | `com.keepersecurity.discovery.KeeperCredentialResolver` |

## Before release
- Publish SDK 17.3.0 to Maven Central, then bump the pin (currently 17.2.0).
- Smoke-test on a real Zurich MID Server (UID and `type:title` lookups with a PAM record shared) plus a pre-Xanadu legacy-JAR check.
- Apply the docs.keeper.io updates.

## Breaking change
On ServiceNow Xanadu and newer, the JAR no longer ships `com.snc.discovery.CredentialResolver`; configure the External Credential Resolver with the FQCN `com.keepersecurity.discovery.KeeperCredentialResolver`. Pre-Xanadu MID Servers are unaffected.

Refs IMP-3033.


.github/workflows/publish.servicenow.plugin.yml
```diff
@@ -43,11 +43,10 @@ jobs:
       - name: Check if version is included in the new release tag
         env:
           TAG: ${{ github.event.inputs.tagname }}
-        if: ${{ false }} # disable for now
         run: |
           GRADLE_CFG=./build.gradle.kts
-          BUILD_VERSION=`fgrep version $GRADLE_CFG | sed -r 's/^version\s+//g' | sed 's/"//g'`
-          if [[ "$TAG" == *"-$BUILD_VERSION"* ]]; then
+          BUILD_VERSION=`fgrep version $GRADLE_CFG | sed -r 's/^version[[:space:]]*=?[[:space:]]*//g' | sed 's/"//g'`
```
stas-schaller: `fgrep version` is an unanchored literal match. `build.gradle.kts` now contains a second line with "version" in it:

```
version = "1.0.0"
// Maven Central. mavenCentral() is listed first, so it is a no-op for versions on Central.
```

The `sed` is anchored with `^`, so only the first line gets stripped; the comment passes through unchanged. `BUILD_VERSION` ends up as a two-line string, and `[[ "$TAG" == *"$BUILD_VERSION"* ]]` never matches any valid tag name. This step exits 1 on every run.

The same bug is present at line 85 (`buildversion`) and line 137 (`BUILD_VERSION` in the Create Release step).

Fix: anchor the grep to the assignment line at all three sites:

```bash
# Before:
BUILD_VERSION=`fgrep version $GRADLE_CFG | sed -r 's/^version[[:space:]]*=?[[:space:]]*//g' | sed 's/"//g'`

# After:
BUILD_VERSION=`grep '^version = ' $GRADLE_CFG | sed -r 's/^version[[:space:]]*=[[:space:]]*//g' | sed 's/"//g'`
```

.github/workflows/publish.servicenow.plugin.yml
```diff
@@ -77,8 +82,8 @@ jobs:
           gradlecfg=${buildroot}/build.gradle.kts
           
           buildname=`fgrep archivesName ${gradlecfg} | sed -r 's/^\s*archivesName\s*=\s*//g' | sed 's/"//g'`
-          buildversion=`fgrep version ${gradlecfg} | sed -r 's/^version\s+//g' | sed 's/"//g'`
-          srcjarpath=${buildroot}/build/libs/${buildname}.jar
+          buildversion=`fgrep version ${gradlecfg} | sed -r 's/^version[[:space:]]*=?[[:space:]]*//g' | sed 's/"//g'`
+          srcjarpath=$(ls ${buildroot}/build/libs/${buildname}*.jar | head -n1)
```
stas-schaller: `srcjarpath` is assigned via glob before any JAR has been built. On a fresh GitHub Actions runner, `build/libs/` doesn't exist at this point; the glob returns nothing and `srcjarpath` is an empty string. Every `mv ${srcjarpath} ${dstjarpath}` in the loop then receives no source argument and exits 2 via the `|| { echo "File Copy Failed!" && exit 2; }` guard. No JARs reach `releases/`.

Fix A: move the glob inside the loop, after `./gradlew jar`:

```bash
./gradlew jar -PresolverVariant=${variant} || gradlew_return_code=$?
if (( gradlew_return_code != 0 )); then
  echo "Gradle build failed with exit status $gradlew_return_code"
  exit $gradlew_return_code
fi
srcjarpath=$(ls ${buildroot}/build/libs/${buildname}*.jar | head -n1)
```

Fix B: once the `fgrep` is corrected so `buildversion` is a clean single value, use a deterministic path that doesn't require the file to exist at assignment time:

```bash
srcjarpath=${buildroot}/build/libs/${buildname}-${buildversion}.jar
```

stas-schaller (changes requested):
Two subtle issues to fix and we're good to go

.github/workflows/publish.servicenow.plugin.yml
mgallego-keeper: Good catch, thank you. You're right — the `mavenLocal()` comment I'd added has "versions" in it, so `fgrep version` matched two lines and `BUILD_VERSION` came out as a multi-line string; the tag check would've failed on every run (and `buildversion` / the Create Release step had the same problem).

Fixed all three sites (48, 85, 137) with the anchored `grep '^version = '` you suggested — e6061b9. I also ended up removing that `mavenLocal()` comment entirely in 46404a7 (17.3.0 is on Maven Central now, so the bridge isn't needed), which removes the trigger too — but I kept the anchored grep since that's the actual fix. Sanity-checked that `grep '^version = ' build.gradle.kts` now returns just `1.0.0`.

.github/workflows/publish.servicenow.plugin.yml
mgallego-keeper: Thanks — yep, I switched it to a glob but left it above the loop, so on a fresh runner `build/libs/` doesn't exist yet and `srcjarpath` resolved to empty (then every `mv` tripped the `|| exit 2` guard).

Went with your Fix B — `srcjarpath=${buildroot}/build/libs/${buildname}-${buildversion}.jar` (e6061b9) — since the version-extraction fix above makes `buildversion` a clean single value now. Confirmed locally that `./gradlew jar` produces exactly `keeper-external-credentials-1.0.0.jar`, so the deterministic path matches.

mgallego-keeper:
Thanks for the review, and for both catches — fixed and pushed in e6061b9.

A couple of related things I did while I was in here, flagging so nothing surprises you:
- Bumped the KSM SDK to `17.3.0` (it published to Maven Central) and dropped the `mavenLocal()` bridge that was only there for the pre-release — 46404a7.
- Your release-branch prep (CHANGELOG + version bump + the VM-111 fix) conflicted with the PR, so I merged it in — the shell-injection fix was already on my branch, so it merged cleanly. I updated the CHANGELOG to match where the code actually landed: FQCN is now `com.snc.discovery.keeper.KeeperCredentialResolver` (moved under `com.snc.discovery.*` on Ivan's suggestion — it's where the other vendor resolvers keep their class; still need to confirm it loads on a live Xanadu+ MID), SDK `17.3.0`, and I added Australia. Since I edited your CHANGELOG, please tell me if you'd rather I revert any of it.

I'll re-request review when you have a minute — thanks again!

stas-schaller (approved):
Both bugs addressed cleanly. The anchored grep and deterministic srcjarpath are correct. Code and docs consistent. One minor typo in the embedded script ("Grade" → "Gradle") — I'll fix it on the release branch after merge. Approving.

mgallego-keeper:
Superseded by #1062. Testing on a live Australia instance (and decompiling the Zurich/Australia MID) showed ServiceNow has no way to configure a custom FQCN for a JAR-import resolver — the MID Server always loads `com.snc.discovery.CredentialResolver`, so the fqcn-only JAR would fail with `ClassNotFoundException` on Xanadu+. #1062 reverts to the single shared class while keeping the KSM SDK 17.3.0 pamSettings fix. Ivan agreed; per him, independent per-connection FQCN is expected in a future ServiceNow release.

PR #1057, opened 2026-07-09
fix(sdk/python/core): raise clear KeeperError for malformed InMemoryKeyValueStorage config (KSM-299)

mgallego-keeper:
## Summary

`InMemoryKeyValueStorage` raised a cryptic `TypeError: object of type 'NoneType' has no len()` when constructed with a config string that is not valid JSON or base64-encoded JSON (e.g. `InMemoryKeyValueStorage("A")`). This fixes KSM-299 so a malformed config produces a clear, actionable `KeeperError`.

## Root cause

In `storage.py` the string-config branch reassigns `config` to the result of `json_to_dict(config)` (which is `None` for invalid JSON), then builds the error with `... % str(len(config))`. Python evaluates the `len(config)` operand before constructing the `KeeperError`, so `len(None)` raises `TypeError` first — the intended "Could not load config data" message never surfaces.

## Fix

- Keep the original config string in `config_str` and report its length, so the error path no longer calls `len()` on the parsed (possibly `None`) result.
- The message now explains the problem: `Could not load config data. The configuration string is not valid JSON or base64-encoded JSON (text length: N).`

## Tests

Added two regression tests to `config_error_test.py`, following the existing "must raise `KeeperError`, not `TypeError`" pattern:
- single-character config (`"A"`) — the exact ticket repro
- a non-JSON, non-base64 string

Verified locally: full `config_error_test.py` passes (17 tests) and the version smoke test passes.

## Release / version

- Bumps python-core `17.3.0 → 17.4.0` across the three coupled locations (`_version.py`, the `keeper_globals.py` hardcoded fallback, and the `smoke_test.py` assertions).
- Adds a consolidated `### 17.4.0` README changelog covering KSM-299 and the other fixes staged on this release branch (KSM-1019, KSM-1080, KSM-1085).

## Notes for reviewers

- Base branch `release/sdk/python/core/v17.4.0` already has the latest `master` merged in (`keeper_secrets_manager_storages` v1.1.0).
- The changelog lists KSM-1019 even though its code lands via #1046 (being re-targeted onto this release branch). This PR owns the single 17.4.0 version bump and the consolidated changelog, so #1046's own version-bump/changelog content is dropped there to avoid conflicts.


stas-schaller (approved):
The bug reproduction is confirmed and the fix is correct. The error path no longer calls `len(None)`; `config_str` preserves the original string for length measurement. Version bumps are consistent. One minor suggestion: add a second assertion on the improved message text ("not valid JSON or base64-encoded JSON") in the two new tests, but that is not a blocker.

PR #1063, opened 2026-07-14
feat(ruby): KSM-906 add IL5 dynamic transmission key support (4-part OTT, keyId 20)

mgallego-keeper:
## Summary

Adds IL5 dynamic transmission-key support to the Ruby SDK so it can redeem the 4-part IL5 one-time token `IL5:clientKey:serverPublicKeyId:serverPublicKey` and talk to `il5.keepersecurity.us`. IL5 uses transmission key id **20**, which is outside the embedded key table (`KEEPER_PUBLIC_KEYS` caps at 18), so Vault ships the key material in the OTT. This is the Ruby counterpart of the Python KSM-932 design (already shipped for Python/JS/Java/Rust); the region mapping (KSM-906) alone is not sufficient for a live IL5 redemption.

Tracked under KSM-906.

## Changes

- **`core.rb#process_token_binding`** — region-gated 4-part OTT parse (`split(':', -1)`, exactly-4 segments, base64 validation of the key). Non-IL5 tokens unchanged.
- **`core.rb#bind_one_time_token`** — persists the OTT-supplied `serverPublicKeyId` and new `serverPublicKey`.
- **`core.rb#generate_transmission_key`** — prefers a configured custom key over the built-in table (so keyId 20 works).
- **`core.rb#initialize`** — new `server_public_key` / `server_public_key_id` options (precedence: programmatic > token > config) and a non-clobbering key-id fallback (mirrors Python).
- **`core.rb#handle_http_error`** — actionable error when the backend rejects a configured custom key (instead of looping on it).
- **`config_keys.rb`** — new `ConfigKeys::KEY_SERVER_PUBLIC_KEY`.
- **`keeper_globals.rb`** — opt-in `KSM_CLIENT_VERSION` override (default unchanged).
- CHANGELOG updated.

Storage already round-trips arbitrary keys, so `serverPublicKey` persists across restarts with no storage change (Ruby has no Java KSM-1066 persistence bug).

## Testing

- New `spec/keeper_secrets_manager/unit/il5_dynamic_key_spec.rb` — **19 examples, 0 failures** (parse validation, precedence, persistence, transmission-key round-trip, rotation handling).
- Full unit suite: 636 examples, 3 failures — all pre-existing, Windows-only `storage_spec` `FileStorage` cases (`Tempfile` held open + atomic `File.rename`), unrelated to this change and green on POSIX CI.


stas-schaller (approved):
Reviewed. Crypto matches the P-256 assumption in the Python, Java, and JS IL5 implementations. Loop-break logic is correct (raises, gated on custom key), KEY_SERVER_PUBLIC_KEY is in ALL_KEYS, split(':', -1) edge case is covered. Two pre-release style nits to address before tagging v17.2.0: CHANGELOG entry format (KSM-906 uses different prefix than its neighbors) and one stale line in the PR description (keeper_globals.rb KSM_CLIENT_VERSION change that isn't in the diff).

PR #1071, opened 2026-07-30
fix(cli): preserve empty custom fields in secret add clone (KSM-1136)

mgallego-keeper:
## Problem

[KSM-1136](https://keeper.atlassian.net/browse/KSM-1136): `ksm secret add clone` silently drops custom fields whose value is empty — the clone is not field-faithful. Regression introduced by the KSM-1118 fix (`b0d660e4`, unreleased, on this branch).

## Root cause

The KSM-1118 workaround strips every field with a falsy `value` from the clone payload before `RecordV3.create_from_data`, to avoid an `IndexError` in `keeper-secrets-manager-helper` (`FieldType.__init__` indexes `self.value[0]` for dict-typed fields when `value: []`). Standard fields survive the strip because the record-type schema recreates them empty; custom fields belong to no schema, so stripped empty custom fields are silently lost (type + label gone from the clone).

The helper root-cause fix (KSM-1119) is still unreleased — verified helper 1.1.2 on PyPI still has the unguarded `self.value[0]` — so the strip must stay for now.

## Fix

In `Secret.add_record_from_clone`:

- Keep the empty-value strip before `create_from_data` (still required to avoid the helper crash).
- After `get_record_create_obj()`, re-attach each source custom field with an empty value to `RecordCreate.custom` as an SDK `RecordField`, preserving `type`, `label`, `required` and `privacyScreen` with `value: []`.

Appended entries match exactly what the helper puts in `custom` (SDK `RecordField` objects), `RecordCreate._validate()` does not type-check `custom`, and serialization was verified to emit the standard wire format: `{"type": "text", "label": "x", "value": [], "required": true}`.

Note: re-attached empty custom fields land at the end of the custom list rather than in source order — a deliberate trade-off to keep the QA-validated helper path for non-empty fields untouched.

Scope is clone-only per KSM-1136. `add_record_from_file` has the same silent drop (same origin commit) and can be handled under its own ticket.

## Tests

- New `test_clone_preserves_empty_custom_fields`: source record with 4 custom fields — non-empty `text`, empty `text`, empty dict-typed `name` (the KSM-1118 crash shape), and an empty `secret` with `required: True`. Asserts all 4 survive into `RecordCreate.custom` and into the serialized `to_dict()` payload, values and flags intact.
- Existing KSM-1118 test (`test_clone_record_with_empty_complex_field`) and KSM-702 test (`test_create_record_custom_field_empty_list`) pass unchanged.
- Full CLI suite: **150 passed, 9 skipped, 0 failures** (Python 3.14, core 17.3.0, helper 1.1.2, storage 1.1.0).

## Suggested QA validation

Clone a record that has both a populated custom field and an empty custom field (ideally one empty dict-typed, e.g. `name`); confirm the clone succeeds and both custom fields appear on the cloned record with labels intact.

[KSM-1136]: https://keeper.atlassian.net/browse/KSM-1136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


stas-schaller (approved):
Verified the root cause, the fix, and the test claims against the actual code, and ran the full CLI suite on this branch (150 passed, 9 skipped, matching the PR description). Approving.

PR #1073, opened 2026-07-30
fix(cli): raise KeeperError correctly in keyring storage fatal handler (KSM-1155)

mgallego-keeper:
## Problem

[KSM-1155](https://keeper.atlassian.net/browse/KSM-1155): `KeyringUtilityStorage.__fatal` crashes with `TypeError: KeeperError.__init__() takes 2 positional arguments but 3 were given` — masking every keyring storage error it is supposed to report.

## Root cause

`__fatal` raises `exceptions.KeeperError(message, error)`, but the SDK's `KeeperError.__init__(self, message)` accepts only a message. Any failure routed through `__fatal` (missing secret name, missing keyring backend, unparseable keyring contents) therefore surfaced as a `TypeError` about the exception constructor instead of the intended `KeeperError` with the real message.

Pre-fix repro:
```
>>> KeyringUtilityStorage(secret_name="")
TypeError: KeeperError.__init__() takes 2 positional arguments but 3 were given
```

## Fix

Raise `KeeperError(message)` and chain the original exception as the cause (`raise ... from error`), preserving the diagnostic chain the two-argument call was presumably reaching for. The pre-raise logging (`logger.error(..., exc_info=error)`) is unchanged. No control-flow changes anywhere else: callers that catch `KeeperError` (e.g. `profile.py`) now receive the intended exception type, and the broad `except Exception` in `__load_config` treats the corrected exception exactly as it treated the `TypeError`.

## Tests

New `KeyringUtilityStorageFatalTest` in `tests/keyring_test.py`:
- `__fatal` raises `KeeperError` (not `TypeError`) with the class-prefixed message
- the original error is chained as `__cause__`
- call-site coverage: `KeyringUtilityStorage(secret_name="")` raises `KeeperError` with the "requires a secret name" message

Full CLI suite: **153 passed, 9 skipped, 0 failures** (Python 3.14, core 17.3.0, helper 1.1.2, storage 1.1.0).

## Suggested QA validation

With the keyring feature installed, trigger a storage error (e.g. corrupt the keyring entry contents so they are not valid JSON, or construct a profile flow with a missing secret name) and confirm the CLI reports a `KeeperError` message naming the actual problem instead of the `TypeError`.

[KSM-1155]: https://keeper.atlassian.net/browse/KSM-1155?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


integration/keeper_secrets_manager_cli/keeper_secrets_manager_cli/keyring_config.py
```diff
@@ -55,7 +55,9 @@ def __fatal(cls, message: str, error: Exception = None):
             cls.logger.error(message, exc_info=error)
         else:
             cls.logger.error(message)
-        raise exceptions.KeeperError(message, error)
+        # KeeperError only accepts a message; chain the original error as the
+        # cause instead of passing it as a second positional argument.
+        raise exceptions.KeeperError(message) from error
```
stas-schaller: `raise ... from error` sets `__suppress_context__ = True` even when `error` is `None`. That hits the `except ImportError:` case above (line 80): under `KSM_DEBUG=1`, the traceback now shows only the `KeeperError`, not the original `ImportError` that caused it, since `from None` explicitly suppresses chaining Python would otherwise add automatically inside an active `except` block. Someone debugging "no keyring backend" loses the specific reason (module missing vs. broken native backend vs. version mismatch). Debug-only, no functional impact, but let's only chain when there's a real cause:
```python
if error is not None:
    raise exceptions.KeeperError(message) from error
raise exceptions.KeeperError(message)
```

stas-schaller (changes requested):
Good stuff, only one fix needed and we're good for merge.

integration/keeper_secrets_manager_cli/keeper_secrets_manager_cli/keyring_config.py
mgallego-keeper: Good catch — applied verbatim in 30d65a30, plus a regression test (`test_missing_keyring_backend_preserves_import_error_context`) that simulates the missing-keyring ImportError and asserts `__context__` is the ImportError with `__suppress_context__` False (it was red against the unconditional `from error`, green now). Full suite 154 passed / 9 skipped.

stas-schaller (approved):

PR #1076, opened 2026-07-31
fix(javascript): KSM-748 use folder key for shared-folder records in flat array

stas-schaller:
## Summary
JavaScript SDK: records created via non-SDK clients inside a shared folder can arrive in the flat `response.records[]` array with `innerFolderUid` set. The `recordKey` for these is wrapped with the folder key, not the app key, but `fetchAndDecryptSecrets` always unwrapped with the app key, so the record silently failed to decrypt and was dropped.

### Fixed
- Records in shared folders now decrypt correctly when returned in the flat `records[]` array instead of being silently skipped (KSM-748)

## Testing
```
```
Two new regression tests in `test/keeper.test.ts`:
- confirms a flat record with `innerFolderUid` decrypts using the matching folder's key
- confirms a flat record whose `innerFolderUid` has no matching folder in the response still falls back to the app key

Also verified live against a real vault: a record created via Commander inside a shared folder, with that folder shared to a throwaway KSM Application, decrypted correctly through this code path with the fix in place.

## Security Impact
This SDK previously selected the wrong decryption key for a specific class of shared-folder records, causing them to fail decryption and be dropped rather than exposing data to the wrong party — no information disclosure, but a correctness/availability gap for legitimate shared-folder access. The fix selects the key based on `innerFolderUid` matching an entry in the folder-key map built from `response.folders`, mirroring the pattern already shipped in the Java SDK (KSM-753). No change to how keys are transmitted, wrapped, or stored.

## Breaking Changes
None.

## Related Issues
- Jira: KSM-748

### DIFF
--- a/sdk/javascript/packages/core/src/keeper.ts
+++ b/sdk/javascript/packages/core/src/keeper.ts
@@ -883,13 +883,34 @@ const fetchAndDecryptSecrets = async (options: SecretManagerOptions, queryOption
         await storage.delete(KEY_CLIENT_KEY)
         await storage.saveString(KEY_OWNER_PUBLIC_KEY, response.appOwnerPublicKey!)
     }
+    const folderKeyIds = new Set<string>()
+    if (response.folders) {
+        for (const folder of response.folders) {
+            try {
+                if (!folder.folderKey) continue
+                await platform.unwrap(platform.base64ToBytes(folder.folderKey), folder.folderUid, KEY_APP_KEY, storage, true)
+                folderKeyIds.add(folder.folderUid)
+            } catch (e: Error | any) {
+                // ignored here; the folders loop below re-attempts the same unwrap and logs the failure
+            }
+        }
+    }
     if (response.records) {
         for (const record of response.records) {
             try {
                 if (record.recordKey) {
-                    await platform.unwrap(platform.base64ToBytes(record.recordKey), record.recordUid, KEY_APP_KEY, storage, true)
+                    // Records created outside the SDK in a shared folder arrive in the flat
+                    // response.records[] with innerFolderUid set; their recordKey is wrapped
+                    // with the folder key, not the app key.
+                    const unwrappingKeyId = record.innerFolderUid && folderKeyIds.has(record.innerFolderUid)
+                        ? record.innerFolderUid
+                        : KEY_APP_KEY
+                    await platform.unwrap(platform.base64ToBytes(record.recordKey), record.recordUid, unwrappingKeyId, storage, true)
                 }
                 const decryptedRecord = await decryptRecord(record, storage)
+                if (record.innerFolderUid && folderKeyIds.has(record.innerFolderUid)) {
+                    decryptedRecord.folderUid = record.innerFolderUid
+                }
                 records.push(decryptedRecord)
             } catch (e: Error | any) {
                 console.error(`Record ${record.recordUid} skipped due to error: ${e.constructor.name}, ${e.message}`)
--- a/sdk/javascript/packages/core/test/keeper.test.ts
+++ b/sdk/javascript/packages/core/test/keeper.test.ts
@@ -405,3 +405,95 @@ test('getFolders skips an undecryptable folder and returns the good one', async 
     expect(folders[0].folderUid).toBe('good-uid')
     expect(folders[0].name).toBe('Good Folder')
 })
+
+test('flat record with innerFolderUid decrypts recordKey using the folder key, not the app key', async () => {
+    const transmissionKey = new Uint8Array(32).fill(1)
+    const appKey = new Uint8Array(32).fill(2)
+    const folderKey = new Uint8Array(32).fill(3)
+    const recordKey = new Uint8Array(32).fill(4)
+    const folderUid = 'folder-uid-1'
+    const recordUid = 'record-uid-1'
+
+    const wrappedFolderKey = await platform.encryptWithKey(folderKey, appKey)
+    const wrappedRecordKey = await platform.encryptWithKey(recordKey, folderKey)
+    const recordData = await platform.encryptWithKey(
+        platform.stringToBytes(JSON.stringify({ title: 'Shared Record', type: 'login', fields: [], custom: [] })), recordKey)
+
+    const serverResponse = {
+        folders: [
+            { folderUid, folderKey: platform.bytesToBase64(wrappedFolderKey), data: '', records: [] }
+        ],
+        records: [
+            {
+                recordUid,
+                recordKey: platform.bytesToBase64(wrappedRecordKey),
+                data: platform.bytesToBase64(recordData),
+                revision: 1,
+                files: [],
+                innerFolderUid: folderUid
+            }
+        ],
+        expiresOn: 0,
+        warnings: []
+    }
+    const encryptedResponse = await platform.encryptWithKey(
+        platform.stringToBytes(JSON.stringify(serverResponse)), transmissionKey)
+
+    platform.getRandomBytes = () => transmissionKey
+    const queryFn = (): Promise<KeeperHttpResponse> => Promise.resolve({ data: encryptedResponse, statusCode: 200, headers: [] })
+
+    const kvs = inMemoryStorage({})
+    await initializeStorage(kvs, 'US:FAKE_CLIENT_KEY')
+    await kvs.saveBytes('appKey', appKey)
+
+    const secrets = await getSecrets({ storage: kvs, queryFunction: queryFn })
+
+    // Bug: the flat-records loop always unwraps recordKey with KEY_APP_KEY, ignoring
+    // innerFolderUid. Since recordKey here is wrapped with the folder key, unwrapping
+    // with the app key throws, the record is caught and silently skipped, and
+    // secrets.records comes back empty instead of containing the decrypted record.
+    expect(secrets.records.length).toBe(1)
+    expect(secrets.records[0].data.title).toBe('Shared Record')
+    expect(secrets.records[0].folderUid).toBe(folderUid)
+})
+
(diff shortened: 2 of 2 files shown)


sdk/javascript/packages/core/src/keeper.ts
```diff
@@ -883,13 +883,34 @@ const fetchAndDecryptSecrets = async (options: SecretManagerOptions, queryOption
...
+    }
     if (response.records) {
         for (const record of response.records) {
             try {
                 if (record.recordKey) {
-                    await platform.unwrap(platform.base64ToBytes(record.recordKey), record.recordUid, KEY_APP_KEY, storage, true)
+                    // Records created outside the SDK in a shared folder arrive in the flat
+                    // response.records[] with innerFolderUid set; their recordKey is wrapped
+                    // with the folder key, not the app key.
+                    const unwrappingKeyId = record.innerFolderUid && folderKeyIds.has(record.innerFolderUid)
```
mgallego-keeper, line 905: Question, not blocking: in the nested path, `innerFolderUid` means the subfolder inside the shared folder (that is what `decryptRecord` copies for folder records). This lookup assumes the flat-path `innerFolderUid` names the shared folder itself, which your live test confirmed for a record in the folder root. If a non-SDK client creates the record inside a subfolder of a shared folder and the server emits the subfolder uid here, it will not match `folderKeyIds` (the `get_secret` `folders[]` entries are the app-key-wrapped shared folders), so the record falls back to the app key and stays dropped, same as before the fix.

Was the subfolder scenario part of the live verification? If it turns out broken, it would need ancestry resolution (`getSharedFolderUid` further down this file is precedent) and could ride the KSM-752 epic as a follow-up, since the shipped Java fix has the identical limitation. Either way behavior degrades to the status quo, never worse, hence approving.

sdk/javascript/packages/core/src/keeper.ts
```diff
@@ -883,13 +883,34 @@ const fetchAndDecryptSecrets = async (options: SecretManagerOptions, queryOption
         await storage.delete(KEY_CLIENT_KEY)
         await storage.saveString(KEY_OWNER_PUBLIC_KEY, response.appOwnerPublicKey!)
     }
+    const folderKeyIds = new Set<string>()
```
mgallego-keeper, line 886: Question for QA rather than a change request: when `getSecrets` is called with a records filter (`requestedRecords` payload), does the `get_secret` response still include the `folders[]` entry carrying the folder key? If the server omits `folders` on filtered fetches, this set stays empty and the fix never engages on that path. A quick live check would settle it; the same exposure exists in the Java implementation.

sdk/javascript/packages/core/src/keeper.ts
```diff
@@ -883,13 +883,34 @@ const fetchAndDecryptSecrets = async (options: SecretManagerOptions, queryOption
...
+                    // Records created outside the SDK in a shared folder arrive in the flat
+                    // response.records[] with innerFolderUid set; their recordKey is wrapped
+                    // with the folder key, not the app key.
+                    const unwrappingKeyId = record.innerFolderUid && folderKeyIds.has(record.innerFolderUid)
+                        ? record.innerFolderUid
+                        : KEY_APP_KEY
+                    await platform.unwrap(platform.base64ToBytes(record.recordKey), record.recordUid, unwrappingKeyId, storage, true)
                 }
                 const decryptedRecord = await decryptRecord(record, storage)
+                if (record.innerFolderUid && folderKeyIds.has(record.innerFolderUid)) {
```
mgallego-keeper, line 911: Nit: this predicate duplicates line 905. Hoisting it once keeps the two sites from drifting:

```ts
const folderKeyId = record.innerFolderUid && folderKeyIds.has(record.innerFolderUid)
    ? record.innerFolderUid
    : undefined
if (record.recordKey) {
    await platform.unwrap(platform.base64ToBytes(record.recordKey), record.recordUid, folderKeyId ?? KEY_APP_KEY, storage, true)
}
const decryptedRecord = await decryptRecord(record, storage)
if (folderKeyId) {
    decryptedRecord.folderUid = folderKeyId
}
```

sdk/javascript/packages/core/src/keeper.ts
```diff
@@ -883,13 +883,34 @@ const fetchAndDecryptSecrets = async (options: SecretManagerOptions, queryOption
         await storage.delete(KEY_CLIENT_KEY)
         await storage.saveString(KEY_OWNER_PUBLIC_KEY, response.appOwnerPublicKey!)
     }
+    const folderKeyIds = new Set<string>()
+    if (response.folders) {
+        for (const folder of response.folders) {
+            try {
+                if (!folder.folderKey) continue
+                await platform.unwrap(platform.base64ToBytes(folder.folderKey), folder.folderUid, KEY_APP_KEY, storage, true)
```
mgallego-keeper, line 891: Nit, take it or leave it: every folder key that unwraps here gets unwrapped again at line 924 in the folders loop, so one extra AES-GCM per folder per fetch. Leaving the folders loop untouched is a defensible call on a release branch, so fine as is. If you do want to dedupe later, the folders loop could skip the unwrap when `folderKeyIds.has(folder.folderUid)`, but then the failure logging this catch comment relies on has to move with it.

sdk/javascript/packages/core/test/keeper.test.ts
```diff
@@ -405,3 +405,95 @@ test('getFolders skips an undecryptable folder and returns the good one', async
...
+
+    const secrets = await getSecrets({ storage: kvs, queryFunction: queryFn })
+
+    // Bug: the flat-records loop always unwraps recordKey with KEY_APP_KEY, ignoring
+    // innerFolderUid. Since recordKey here is wrapped with the folder key, unwrapping
+    // with the app key throws, the record is caught and silently skipped, and
+    // secrets.records comes back empty instead of containing the decrypted record.
+    expect(secrets.records.length).toBe(1)
+    expect(secrets.records[0].data.title).toBe('Shared Record')
+    expect(secrets.records[0].folderUid).toBe(folderUid)
```
mgallego-keeper, line 457: Optional: also asserting `expect(secrets.records[0].innerFolderUid).toBe(folderUid)` would pin the full output convention for flat shared-folder records (`decryptRecord` copies `innerFolderUid`, this fix adds `folderUid`), matching what the Java result carries.

sdk/javascript/packages/core/test/keeper.test.ts
```diff
@@ -405,3 +405,95 @@ test('getFolders skips an undecryptable folder and returns the good one', async
...
+    const queryFn = (): Promise<KeeperHttpResponse> => Promise.resolve({ data: encryptedResponse, statusCode: 200, headers: [] })
+
+    const kvs = inMemoryStorage({})
+    await initializeStorage(kvs, 'US:FAKE_CLIENT_KEY')
+    await kvs.saveBytes('appKey', appKey)
+
+    const secrets = await getSecrets({ storage: kvs, queryFunction: queryFn })
+
+    expect(secrets.records.length).toBe(1)
+    expect(secrets.records[0].data.title).toBe('Orphaned Record')
```
mgallego-keeper, line 498: Optional: adding `expect(secrets.records[0].folderUid).toBeUndefined()` here would lock the tagging condition, i.e. `folderUid` is only set when the folder key was actually found and used.

mgallego-keeper (approved):
Approving. The fix is correct for the reported scenario, mirrors the Java KSM-753 implementation that already shipped in `java-sdk-v17.3.0` (folder-key map built before the records loop, per-record key selection by `innerFolderUid`, app-key fallback, `folderUid` tagged only when the folder key was found), and every degraded case fails closed to the pre-fix behavior (record skipped and logged) rather than throwing out of `getSecrets`.

**Verification I ran locally**, since `test.js.yml` only triggers on PRs to master and no test CI runs on this release-branch PR:

- Full suite at 95a33a81: 6 suites, 57 tests, all green (the rollup pretest build type-checks src, ts-jest type-checks the tests).
- Mutation check: with `src/keeper.ts` reverted to the base commit (75c63bfa) and the new tests kept, the first new test fails with the exact original symptom (`Record record-uid-1 skipped due to error: Error, Unsupported state or unable to authenticate data`) and the fallback test still passes. Both regression tests genuinely guard the fix.

**Other checks that came back clean:**

- Field choice: `innerFolderUid` is confirmed three ways: the shipped Java fix, the Python DTO wire parsing, and the live vault verification described above. The earlier attempt on this ticket (PR #907, keyed on `record.folderUid`) never shipped in any tag after the eede2533 removal, so there is nothing to unwind.
- Browser platform parity: WebCrypto unwraps folder keys with the `unwrapKey` usage, so they work as unwrapping keys for record keys; GCM on both hops (the `appKey` CBC special case is not in play); `memoryOnly=true` keeps folder keys out of persisted storage, same as the folders loop.
- Placement is right: the pre-loop runs after the `encryptedAppKey` bind block, so the app key is always available to it.

**Non-blocking notes without an inline anchor:**

- `SecretsManagerResponseRecord.innerFolderUid` (line 228) is declared required but is optional on the wire; the new truthiness guard is correct, and `innerFolderUid?: string` would let the types say so. Pre-existing, zero runtime impact.
- If the server ever returned the same record both flat and nested under `folders[].records`, it would now surface twice, where the flat copy previously self-deduped by failing decryption. Duplicates are already a tolerated condition (KSM-731 notation handling) and Java behaves the same; FYI only.
- Optional extra test scenarios if you feel like it: a mixed batch (one app-key flat record plus one folder-key flat record) to pin per-record selection, and a corrupt folder key proving `getSecrets` still succeeds.
- Process, for follow-ups rather than this PR: consider adding the release branches to `test.js.yml` triggers so release-branch PRs run the suite, and make sure KSM-748 lands in the 17.6.0 CHANGELOG entry at release cut (REL-5278).

The inline comments are two questions (subfolder case, filtered fetch) and small nits; none of them block the merge.

PR #1077, opened 2026-07-31
fix(javascript): KSM-1035 one-sided throttle jitter and retry_after cap

stas-schaller:
## Summary
JavaScript SDK: applies the throttle-backoff hardening already shipped for Ruby (KSM-883) to the JS retry loop, tightening the jitter to one-sided and capping a server-supplied `retry_after`.

### Fixed
- `throttleJitter()` narrowed from `[-0.25, 0.25)` to `[0, 0.25)`. The old symmetric jitter could shave up to 25% off the computed backoff floor, so a retry could fire before the server's throttle window cleared and immediately re-trigger the same throttle (KSM-1035)
- `parseThrottle()` now caps a server-supplied `retry_after` at `MAX_THROTTLE_DELAY_SEC` (176s) — the same ceiling the exponential backoff branch already reaches on its last retry attempt, so a server-supplied value can no longer exceed what the client would otherwise wait

## Testing
```
```
`test/throttle.test.ts`:
- new: `throttleJitter` never returns a negative value across 200 samples
- new: `parseThrottle` clamps `retry_after` at 176s
- tightened the existing `honors retry_after from the response body` e2e assertion's lower bound from 2250ms to 3000ms, since jitter can no longer subtract time

## Security Impact
None — this is timing/backoff logic only; no change to how secrets are transmitted, encrypted, or stored.

## Breaking Changes
None.

## Related Issues
- Jira: KSM-1035

### DIFF
--- a/sdk/javascript/packages/core/src/keeper.ts
+++ b/sdk/javascript/packages/core/src/keeper.ts
@@ -96,7 +96,7 @@ export const parseThrottle = (body: string): number | null => {
 /**
  * Computes the backoff delay (milliseconds) for a 0-based `attempt`: `retryAfter` seconds when
  * provided (> 0), otherwise exponential backoff (BASE_THROTTLE_DELAY_SEC * 2**attempt -> 11, 22,
- * 44, 88, 176s). The `jitter` fraction (typically in [-0.25, 0.25)) is then applied.
+ * 44, 88, 176s). The `jitter` fraction (typically in [0, 0.25)) is then applied.
  */
 export const throttleDelay = (attempt: number, retryAfter: number, jitter: number = throttleJitter()): number => {
     const baseSec = retryAfter > 0 ? retryAfter : BASE_THROTTLE_DELAY_SEC * Math.pow(2, attempt)
--- a/sdk/javascript/packages/core/src/keeper.ts
+++ b/sdk/javascript/packages/core/src/keeper.ts
@@ -20,6 +20,7 @@ const KEY_PRIVATE_KEY = 'privateKey' // The client's private key
 // request, so the counter only clears after 10s of silence).
 const MAX_THROTTLE_RETRIES = 5
 const BASE_THROTTLE_DELAY_SEC = 11 // 1s safety margin over the backend's 10s memcached TTL
+const MAX_THROTTLE_DELAY_SEC = 176 // same ceiling the exponential branch reaches at the last retry (11 * 2**4)
 const CLIENT_ID_HASH_TAG = 'KEEPER_SECRETS_MANAGER_CLIENT_ID' // Tag for hashing the client key to client id
 
 let keeperPublicKeys: Record<number, Uint8Array>
@@ -62,14 +63,17 @@ export type SecretManagerOptions = {
 // utils.ts/platform code that throws them; re-exported here so the public API is unchanged.
 export {KeeperError, KeeperThrottleError} from './errors'
 
-// Returns a jitter multiplier in [-0.25, 0.25). Kept separate so concurrent clients
-// desynchronize their retries; unit tests exercise throttleDelay with a pinned jitter.
-export const throttleJitter = (): number => Math.random() * 0.5 - 0.25
+// Returns a jitter multiplier in [0, 0.25). One-sided so the delay never drops below the
+// computed floor (retrying too soon just re-triggers the throttle); kept separate so
+// concurrent clients desynchronize their retries. Unit tests exercise throttleDelay with a
+// pinned jitter.
+export const throttleJitter = (): number => Math.random() * 0.25
 
 /**
  * If `body` is a backend throttle error (`result_code`/`error` === "throttled") returns its
- * `retry_after` in seconds (>= 0); otherwise returns `null` so the caller falls through to
- * normal error handling. Non-JSON / non-object bodies return `null`.
+ * `retry_after` in seconds (>= 0, capped at MAX_THROTTLE_DELAY_SEC); otherwise returns `null`
+ * so the caller falls through to normal error handling. Non-JSON / non-object bodies return
+ * `null`.
  */
 export const parseThrottle = (body: string): number | null => {
     let obj: any
@@ -86,7 +90,7 @@ export const parseThrottle = (body: string): number | null => {
         return null
     }
     const retryAfter = Number(obj.retry_after)
-    return Number.isFinite(retryAfter) && retryAfter > 0 ? retryAfter : 0
+    return Number.isFinite(retryAfter) && retryAfter > 0 ? Math.min(retryAfter, MAX_THROTTLE_DELAY_SEC) : 0
 }
 
 /**
--- a/sdk/javascript/packages/core/test/throttle.test.ts
+++ b/sdk/javascript/packages/core/test/throttle.test.ts
@@ -60,8 +60,8 @@ describe('throttleDelay (unit)', () => {
         expect(throttleDelay(0, 0, 0)).toBe(11000)
         expect(throttleDelay(1, -5, 0)).toBe(22000)
     })
-    test('jitter bounds keep the first delay in [8.25s, 13.75s]', () => {
-        expect(throttleDelay(0, 0, -0.25)).toBe(8250)
+    test('jitter bounds keep the first delay in [11s, 13.75s)', () => {
+        expect(throttleDelay(0, 0, 0)).toBe(11000)
         expect(throttleDelay(0, 0, 0.25)).toBe(13750)
     })
 })
--- a/sdk/javascript/packages/core/test/throttle.test.ts
+++ b/sdk/javascript/packages/core/test/throttle.test.ts
@@ -9,6 +9,7 @@ import {
     KeeperThrottleError,
     parseThrottle,
     throttleDelay,
+    throttleJitter,
 } from '../'
 
 // A valid one-time token (same fixture the e2e suite uses) so initializeStorage produces a
@@ -65,6 +66,16 @@ describe('throttleDelay (unit)', () => {
     })
 })
 
+describe('throttleJitter (unit)', () => {
+    test('is one-sided: never pushes the delay below its floor', () => {
+        for (let i = 0; i < 200; i++) {
+            const jitter = throttleJitter()
+            expect(jitter).toBeGreaterThanOrEqual(0)
+            expect(jitter).toBeLessThan(0.25)
+        }
+    })
+})
+
 describe('parseThrottle (unit)', () => {
     test('throttled via error / result_code with retry_after variants', () => {
         expect(parseThrottle('{"error":"throttled"}')).toBe(0)
@@ -76,6 +87,13 @@ describe('parseThrottle (unit)', () => {
         expect(parseThrottle('{"error":"key"}')).toBeNull()
         expect(parseThrottle('not json')).toBeNull()
         expect(parseThrottle('')).toBeNull()
+    })
+    test('retry_after beyond the exponential ceiling is capped at 176s', () => {
+        // 176s = BASE_THROTTLE_DELAY_SEC * 2**(MAX_THROTTLE_RETRIES - 1) = 11 * 2**4, the same
+        // ceiling the exponential branch already reaches on the last retry attempt.
+        expect(parseThrottle('{"error":"throttled","retry_after":500}')).toBe(176)
+        expect(parseThrottle('{"error":"throttled","retry_after":176}')).toBe(176)
+        expect(parseThrottle('{"error":"throttled","retry_after":175}')).toBe(175)
     })
 })
 
@@ -118,8 +136,8 @@ describe('throttle retry (e2e via getSecrets)', () => {
             call++ === 0 ? throttle403(3) : throttle403()
         )
         await expect(getSecrets(options)).rejects.toBeInstanceOf(KeeperThrottleError)
-        // retry_after = 3s with +/-25% jitter -> [2.25s, 3.75s]
-        expect(sleeps[0]).toBeGreaterThanOrEqual(2250)
+        // retry_after = 3s with one-sided [0, +25%) jitter -> [3s, 3.75s]; never below the 3s floor
+        expect(sleeps[0]).toBeGreaterThanOrEqual(3000)
         expect(sleeps[0]).toBeLessThanOrEqual(3750)
     })
 
--- a/sdk/javascript/packages/core/CHANGELOG.md
+++ b/sdk/javascript/packages/core/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Log
 
+## 17.6.0
+- KSM-1035 - Throttle backoff hardening: retry jitter is now one-sided (0 to +25%, so delays never fall below the computed floor), and a server-supplied `retry_after` is capped at 176s (the exponential ladder's last step) so it can no longer force an excessive or unbounded wait.
+
 ## 17.5.0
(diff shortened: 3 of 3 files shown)


sdk/javascript/packages/core/src/keeper.ts
```diff
@@ -20,6 +20,7 @@ const KEY_PRIVATE_KEY = 'privateKey' // The client's private key
 // request, so the counter only clears after 10s of silence).
 const MAX_THROTTLE_RETRIES = 5
 const BASE_THROTTLE_DELAY_SEC = 11 // 1s safety margin over the backend's 10s memcached TTL
+const MAX_THROTTLE_DELAY_SEC = 176 // same ceiling the exponential branch reaches at the last retry (11 * 2**4)
```
mgallego-keeper, line 23: Nit, optional: two readers can trip on this comment. The cap applies before jitter, so the actual sleep on a capped value reaches 220s; someone who reads 176 as the per-attempt sleep bound (say, to size an outer timeout) gets burned. Kotlin's comment wording avoids implying a sleep bound. Separately, deriving the value (`BASE_THROTTLE_DELAY_SEC * 2 ** (MAX_THROTTLE_RETRIES - 1)`) would keep it in lockstep if the ladder is ever retuned, though Ruby and Kotlin both hardcode 176, so textual parity argues for keeping the literal.

```suggestion
const MAX_THROTTLE_DELAY_SEC = 176 // exponential-branch ceiling at the last retry (11 * 2**4); jitter can still push the actual sleep to 220s
```

sdk/javascript/packages/core/src/keeper.ts
```diff
@@ -62,14 +63,17 @@ export type SecretManagerOptions = {
...
 export {KeeperError, KeeperThrottleError} from './errors'
 
-// Returns a jitter multiplier in [-0.25, 0.25). Kept separate so concurrent clients
-// desynchronize their retries; unit tests exercise throttleDelay with a pinned jitter.
-export const throttleJitter = (): number => Math.random() * 0.5 - 0.25
+// Returns a jitter multiplier in [0, 0.25). One-sided so the delay never drops below the
+// computed floor (retrying too soon just re-triggers the throttle); kept separate so
+// concurrent clients desynchronize their retries. Unit tests exercise throttleDelay with a
+// pinned jitter.
+export const throttleJitter = (): number => Math.random() * 0.25
```
mgallego-keeper, line 70: The jitter contract changed here, but `throttleDelay`'s JSDoc (line 99, just outside this diff) still says the jitter fraction is "typically in [-0.25, 0.25)". Left as-is, it documents symmetric jitter as the contract on an exported function, which invites reintroducing exactly the sub-floor delay this PR removes. Ruby and Kotlin both updated the equivalent sentence when they shipped (SecretsManager.kt line 1770 reads "(typically in [0, 0.25))"). Please update it to match.

sdk/javascript/packages/core/src/keeper.ts
```diff
@@ -62,14 +63,17 @@ export type SecretManagerOptions = {
...
 
 /**
  * If `body` is a backend throttle error (`result_code`/`error` === "throttled") returns its
- * `retry_after` in seconds (>= 0); otherwise returns `null` so the caller falls through to
- * normal error handling. Non-JSON / non-object bodies return `null`.
+ * `retry_after` in seconds (>= 0, capped at MAX_THROTTLE_DELAY_SEC); otherwise returns `null`
```
mgallego-keeper, line 74: Nano-nit: consider stating the value, e.g. "capped at 176s (`MAX_THROTTLE_DELAY_SEC`)". The constant is module-private and the generated typings do not carry this JSDoc, so nothing outside this file states the actual number.

sdk/javascript/packages/core/src/keeper.ts
```diff
@@ -86,7 +90,7 @@ export const parseThrottle = (body: string): number | null => {
         return null
     }
     const retryAfter = Number(obj.retry_after)
-    return Number.isFinite(retryAfter) && retryAfter > 0 ? retryAfter : 0
+    return Number.isFinite(retryAfter) && retryAfter > 0 ? Math.min(retryAfter, MAX_THROTTLE_DELAY_SEC) : 0
```
mgallego-keeper, line 93: Optional defense-in-depth, fine to punt since this placement is exactly what KSM-1035 prescribes: with the clamp living here, neither of the PR's two invariants is structural inside `throttleDelay` (lines 101-105), and unlike Kotlin's `internal` equivalent, `throttleDelay` is exported from the package index. Its `retryAfter` parameter is uncapped, a negative `jitter` still produces a sub-floor delay, and the `Math.max(sec, 0)` guard at line 104 clamps at the wrong bound (0, not the floor) and is dead code for all real inputs. `Math.max(sec, baseSec) * 1000` would make the floor structural, and a `Math.min(retryAfter, MAX_THROTTLE_DELAY_SEC)` inside the function would make the cap structural, which is where Ruby enforces it (core.rb lines 1557-1558). Note the pre-existing test at throttle.test.ts lines 63-66 would fail on the floor clamp, so it needs its rework first.

One more small thing in the same vein: when this clamp engages, the warning at line 809 logs only the capped delay, so the server's stated window is unobservable. Logging "server requested Ns, capped to 176s" when clamping would preserve the signal for telemetry.

sdk/javascript/packages/core/test/throttle.test.ts
```diff
@@ -65,6 +66,16 @@ describe('throttleDelay (unit)', () => {
     })
```
mgallego-keeper, line 66: The test ending here (lines 63-66) still pins `jitter = -0.25` and presents 8.25s, below the backend's 10s window, as a legal first delay, so the suite now asserts two contradictory jitter contracts (this one vs the new `throttleJitter` block below). KSM-1030's acceptance criteria pin the first-retry floor at the base delay, not base minus 25%, and the Ruby spec was reworked accordingly (floor 11.0 with rand pinned to 0, ceiling 13.75 at 0.25). Suggested mirror:

```ts
    test('keeps the first delay within the one-sided 0..+25% jitter bounds', () => {
        expect(throttleDelay(0, 0, 0)).toBe(11000)    // floor: no jitter added
        expect(throttleDelay(0, 0, 0.25)).toBe(13750) // ceiling: +25%
    })
```

sdk/javascript/packages/core/test/throttle.test.ts
```diff
@@ -65,6 +66,16 @@ describe('throttleDelay (unit)', () => {
...
 })
 
+describe('throttleJitter (unit)', () => {
+    test('is one-sided: never pushes the delay below its floor', () => {
+        for (let i = 0; i < 200; i++) {
+            const jitter = throttleJitter()
+            expect(jitter).toBeGreaterThanOrEqual(0)
+            expect(jitter).toBeLessThan(0.25)
+        }
```
mgallego-keeper, line 75: Optional: these assertions are range-only, so a regression that collapses `throttleJitter` to a constant in-range value (say `() => 0`) stays green while silently losing the desynchronization property the function's own comment states as its purpose. One deterministic extra assertion closes it:

```suggestion
        const draws: number[] = []
        for (let i = 0; i < 200; i++) {
            const jitter = throttleJitter()
            expect(jitter).toBeGreaterThanOrEqual(0)
            expect(jitter).toBeLessThan(0.25)
            draws.push(jitter)
        }
        expect(new Set(draws).size).toBeGreaterThan(1)
```

sdk/javascript/packages/core/test/throttle.test.ts
```diff
@@ -77,6 +88,13 @@ describe('parseThrottle (unit)', () => {
         expect(parseThrottle('not json')).toBeNull()
         expect(parseThrottle('')).toBeNull()
     })
+    test('retry_after beyond the exponential ceiling is capped at 176s', () => {
```
mgallego-keeper, line 91: Optional: the cap is pinned only at the `parseThrottle` unit level; a refactor that moves the clamping around could restore uncapped sleeps with every e2e test staying green. The recording harness makes the end-to-end lock cheap, something like this in the e2e describe:

```ts
    test('caps a huge retry_after through the real retry path', async () => {
        let call = 0
        const { options, sleeps } = await makeOptions(async () =>
            call++ === 0 ? throttle403(500) : throttle403()
        )
        await expect(getSecrets(options)).rejects.toBeInstanceOf(KeeperThrottleError)
        // 500s capped to 176s, then one-sided [0, +25%) jitter -> [176s, 220s)
        expect(sleeps[0]).toBeGreaterThanOrEqual(176000)
        expect(sleeps[0]).toBeLessThan(220000)
    })
```

sdk/javascript/packages/core/test/throttle.test.ts
```diff
@@ -118,8 +136,8 @@ describe('throttle retry (e2e via getSecrets)', () => {
             call++ === 0 ? throttle403(3) : throttle403()
         )
         await expect(getSecrets(options)).rejects.toBeInstanceOf(KeeperThrottleError)
-        // retry_after = 3s with +/-25% jitter -> [2.25s, 3.75s]
-        expect(sleeps[0]).toBeGreaterThanOrEqual(2250)
+        // retry_after = 3s with one-sided [0, +25%) jitter -> [3s, 3.75s]; never below the 3s floor
+        expect(sleeps[0]).toBeGreaterThanOrEqual(3000)
```
mgallego-keeper, line 140: For the record: KSM-1035 routes test-assertion changes through the test-review gate before implementation, and this tightens an existing assertion. The change itself is exactly right (3000ms is the new lower distribution edge, and 3750 stays valid since jitter is strictly below 0.25), so flagging only to confirm the gate ran; happy to treat this review as it otherwise.

mgallego-keeper (changes requested):
Reviewed at fef68e26. Ran the suite locally since release-branch PRs only trigger the Socket checks: **6/6 suites, 57/57 tests pass** (`npm test`: pretest rollup build + jest, Node 22).

The change itself is correct and I could not break it. The one-sided jitter keeps every delay at or above its floor (first retry 11s, clearing the backend's 10s reset-on-request window), the 176s cap matches the exponential ladder's last step (11 * 2**4), boundary coverage at 175/176/500 is right, and the tightened e2e lower bound is exactly the new distribution edge for `retry_after: 3` (3750 stays valid since jitter is strictly below 0.25). `Infinity`/`NaN`/negative/absent `retry_after` still route safely to the exponential branch via the `Number.isFinite` guard, so the cap cannot be bypassed. Worst-case total wait is now bounded at 5 sleeps of [176s, 220s), roughly 15 to 18 minutes, where it was previously unbounded. A bonus the PR does not claim: the cap also eliminates a latent 32-bit `setTimeout` overflow, where a `retry_after` above ~2.1M seconds fired the retry after ~1ms, turning absurd server values into rapid-fire retries against an already-throttled endpoint. Parity checks out too: Kotlin caps in the parse path like this PR does, Ruby caps inside `throttle_delay`; observable behavior is identical across the three.

Requesting changes for three small items. Two are inline (the stale `throttleDelay` JSDoc, the stale symmetric-jitter unit test); the third cannot anchor inline because the file is not in the diff:

**Missing 17.6.0 CHANGELOG entry.** Convention is one entry per KSM ticket (the Ruby equivalent has one), and it matters more than usual here: the newest entry (17.5.0 / KSM-880) still advertises "plus ±25% jitter" with unconditional `retry_after` honoring, and CHANGELOG.md is the only user-facing documentation of throttle behavior. Your PR description's Fixed section is already the right text; suggested entry:

```markdown
## 17.6.0
- KSM-1035 - Throttle backoff hardening: retry jitter is now one-sided (0 to +25%, so delays never fall below the computed floor), and a server-supplied `retry_after` is capped at 176s (the exponential ladder's last step) so it can no longer force an excessive or unbounded wait.
```

Also inline, optional and non-blocking: making both invariants structural inside the exported `throttleDelay` (plus a logging nit for when the cap engages), an e2e test driving the cap through `getSecrets`, a variability assertion on the new jitter test, and two wording nits.

Noting for the record, no action needed:
- The cap deliberately undershoots a large server-stated window (five 176-220s waits, then `KeeperThrottleError`, where the old code slept once for the full value). That trade-off is explicit in KSM-1030, and honest values should be tiny given the backend's 10s window, so this is the right call.
- Sub-floor `retry_after` values (e.g. `0.5`) are still honored as-is and can burn all five retries in a few seconds. Pre-existing cross-SDK policy, out of scope here.
- `Number()` coercion accepts junk the sibling SDKs route to the exponential branch (`true` becomes 1s, `[500]` becomes 500, then 176). Pre-existing and JS-unique; follow-up material at most.
- Python and Rust still carry the old symmetric uncapped backoff pending their KSM-1030 children; Go and .NET have no merged throttle code yet.

sdk/javascript/packages/core/src/keeper.ts
mgallego-keeper: Fixed in 5c99bfa5, wording now matches the Kotlin sentence. Resolving.

sdk/javascript/packages/core/test/throttle.test.ts
mgallego-keeper: Reworked in 5c99bfa5: floor pinned at throttleDelay(0, 0, 0) = 11000, ceiling kept at 13750, and the name now matches the one-sided distribution. The suite no longer carries two contradictory jitter contracts. Resolving.

sdk/javascript/packages/core/test/throttle.test.ts
mgallego-keeper: No separate gate ran, so treating this review cycle as the gate, as offered. The tightened bound is verified again at 5c99bfa5 (59/59 on the merge with the current release head). Resolving.

mgallego-keeper (approved):
Re-reviewed at 5c99bfa5, including the merge with the current release head. All three requested changes land correctly:

- CHANGELOG: `## 17.6.0` section added with the KSM-1035 entry; the 17.5.0 section is correctly left as historical record.
- `throttleDelay` JSDoc now documents the one-sided `[0, 0.25)` jitter, matching Kotlin and Ruby.
- The jitter-bounds unit test now pins the floor at `throttleDelay(0, 0, 0) = 11000` and keeps the 13750 ceiling, so the suite no longer asserts two contradictory jitter contracts.

Verification: the base moved under this PR (#1076 / KSM-748, a different region of keeper.ts, no semantic interaction); GitHub reports a clean merge, and the full suite passes locally on the merge of 5c99bfa5 with the release head: 6/6 suites, 59/59 tests. Socket checks are green on the new head.

The optional inline suggestions (structural clamps in `throttleDelay`, the e2e cap lock, the jitter variability assertion, two wording nits) stay non-blocking: punt or take any later, and feel free to resolve those threads either way. If you do touch the tests again, the two cheapest wins are the e2e cap test and one value assertion on `sleeps[1]` in the honors-retry_after e2e (expected range [22000, 27500)), which would keep an attempt-counter regression in the retry loop from staying green.

For the record, pre-existing and non-blocking:
- `retry_after` now has a max cap but still no minimum floor, so a sub-10s server value (say 0.5) still drives retries inside the backend's reset-on-request window and can burn all five attempts in seconds. Ruby and Kotlin behave the same, so if this is ever worth changing it's an all-SDK follow-up under KSM-1030, not this PR.
- Throttle detection reads only the first 1000 bytes of the 403 body; a throttle body padded past 1KB would truncate mid-JSON and skip the retry path entirely. Real throttle bodies are ~50 bytes, so speculative; unchanged since 17.5.0.
- Release staging: the new 17.6.0 section carries only KSM-1035 so far, while the branch already has KSM-748, KSM-1073, KSM-1079, and KSM-1084 unlogged; they'll need the usual backfill docs commit at the cut (as 37b4b280 did for 17.5.0).

Approving.

PR #1078, opened 2026-07-31
fix(javascript): KSM-1128 bound server-key-rotation retries in postQuery

stas-schaller:
## Summary

JavaScript SDK: the \`postQuery\` key-rotation branch (\`{"error":"key"}\`, default path when no custom server public key is pinned) had no retry cap and no validation on the server-suggested \`key_id\`. A server that kept suggesting the same key would retry forever. A server that suggested an out-of-table id would persist it to storage and break every subsequent call.

### Fixed
- Added a \`keyRotationAttempt\` counter bounded by \`MAX_KEY_ROTATION_RETRIES\` (3). After 3 rotation attempts the SDK throws a typed \`KeeperError\` instead of looping forever (KSM-1128).
- The server-suggested \`key_id\` is validated before use. The shape guard rejects any value that is not a positive integer. The membership guard rejects any id not present in the bundled key table (keys 7-18 as of this version). An unsupported id now throws a typed \`KeeperError\` and is never persisted to storage.
- When the server sends \`{"error":"key"}\` to a client with a pinned custom server public key, the existing custom-key diagnostic always fires regardless of what \`key_id\` contains. The shape and membership guards only run on the standard rotation path.

## Testing
\`\`\`
\`\`\`
New tests in \`test/keeper.test.ts\`:
- \`key rotation - retries are bounded, not infinite\` — asserts the SDK throws after exactly \`MAX_KEY_ROTATION_RETRIES + 1\` calls rather than looping.
- \`key rotation - suggested key id is adopted and persisted\` — verifies that a valid suggested id is adopted and saved to storage (mutation-verified: deleting the \`saveString\` call fails this test).
- \`key rotation - unsupported suggested key id is rejected, not persisted\` — sends \`key_id: 99\`, asserts exactly 1 request, verifies \`serverPublicKeyId\` is never written to storage (mutation-verified: removing the membership check fails this test with the config-poisoning path itself).

## Security Impact
The membership check closes a configuration-poisoning vector: before this fix, a server could suggest an out-of-table \`key_id\` and the SDK would persist it, making the client unable to connect on every subsequent call. No change to how keys are transmitted, wrapped, or decrypted.

## Breaking Changes
None.

## Related Issues
- Jira: KSM-1128

### DIFF
--- a/sdk/javascript/packages/core/CHANGELOG.md
+++ b/sdk/javascript/packages/core/CHANGELOG.md
@@ -6,6 +6,7 @@
 - KSM-1084 - Fixed `deleteSecret()` and `deleteFolder()` silently reporting success when the server rejected some UIDs. The SDK now surfaces per-item error messages from the server to the caller.
 - KSM-748 - Fixed `getSecrets()` silently dropping records created by Commander or the Vault UI inside shared folders. The SDK now uses the folder key to decrypt the record key for any flat record that has `innerFolderUid` set. This matches the behavior for records in `folders[].records[]`.
 - KSM-1035 - Fixed throttle retry jitter being two-sided, which could reduce a retry delay below the computed floor. Jitter is now one-sided (0 to +25%). The SDK also caps a server-supplied `retry_after` at 176s to prevent an arbitrarily long wait.
+- KSM-1128 - Bounded the server key-rotation retry in `postQuery`. When the server sends `{"error":"key"}`, the code retries at most 3 times before throwing a typed `KeeperError`, instead of retrying forever. Before storing a suggested `key_id`, the code validates its shape (positive integer) and its membership in the bundled key table (keys 7-18). An unsupported key id can no longer corrupt the configuration. The pinned custom-key path does not change.
 - Maintenance: Updated `minimatch`, `@babel/core`, and `handlebars` dev dependencies.
 
 ## 17.5.0
--- a/sdk/javascript/packages/core/src/keeper.ts
+++ b/sdk/javascript/packages/core/src/keeper.ts
@@ -19,6 +19,10 @@ const KEY_PRIVATE_KEY = 'privateKey' // The client's private key
 // per clientId+endpoint (100 requests / 10s window; memcached TTL 10s that resets on every
 // request, so the counter only clears after 10s of silence).
 const MAX_THROTTLE_RETRIES = 5
+// Bounds the server-key-rotation retry (postQuery's `error === 'key'` branch, no custom key
+// pinned): one legitimate rotation should resolve it, so this only needs to tolerate a little
+// slack, not act as a real retry budget.
+const MAX_KEY_ROTATION_RETRIES = 3
 const BASE_THROTTLE_DELAY_SEC = 11 // 1s safety margin over the backend's 10s memcached TTL
 const MAX_THROTTLE_DELAY_SEC = 176 // same ceiling the exponential branch reaches at the last retry (11 * 2**4)
 const CLIENT_ID_HASH_TAG = 'KEEPER_SECRETS_MANAGER_CLIENT_ID' // Tag for hashing the client key to client id
@@ -788,6 +792,7 @@ const postQuery = async (options: SecretManagerOptions, path: string, payload: A
     const url = `https://${hostName}/api/rest/sm/v1/${path}`
     const sleep = options.throttleSleep || ((ms: number) => new Promise<void>(resolve => setTimeout(resolve, ms)))
     let throttleAttempt = 0
+    let keyRotationAttempt = 0
     while (true) {
         const transmissionKey = await generateTransmissionKey(options.storage)
         const encryptedPayload = await encryptAndSignPayload(options.storage, transmissionKey, payload)
@@ -820,7 +825,11 @@ const postQuery = async (options: SecretManagerOptions, path: string, payload: A
                         const currentKeyId = await options.storage.getString(KEY_SERVER_PUBLIC_KEY_ID)
                         throw new Error(`Server rejected the custom server public key (id ${currentKeyId}). The server suggested key id ${errorObj.key_id}. Please update your IL5 KSM configuration.`)
                     }
+                    if (keyRotationAttempt >= MAX_KEY_ROTATION_RETRIES) {
+                        throw new Error(`Server key rotation exhausted ${MAX_KEY_ROTATION_RETRIES} retries; suggested key id ${errorObj.key_id} was not accepted`)
+                    }
                     await options.storage.saveString(KEY_SERVER_PUBLIC_KEY_ID, errorObj.key_id!.toString())
+                    keyRotationAttempt++
                     continue
                 }
             } else {
--- a/sdk/javascript/packages/core/src/keeper.ts
+++ b/sdk/javascript/packages/core/src/keeper.ts
@@ -821,13 +821,13 @@ const postQuery = async (options: SecretManagerOptions, path: string, payload: A
                 try { errorObj = JSON.parse(errorMessage) } catch {}
                 if (errorObj?.error === 'key') {
                     const suggestedKeyId = errorObj.key_id
-                    if (typeof suggestedKeyId !== 'number' || !Number.isInteger(suggestedKeyId) || suggestedKeyId <= 0) {
-                        throw new KeeperError(`Server key error response contains invalid key_id: ${JSON.stringify(suggestedKeyId)}`)
-                    }
                     const customKey = await options.storage.getString(KEY_SERVER_PUBLIC_KEY)
                     if (customKey) {
                         const currentKeyId = await options.storage.getString(KEY_SERVER_PUBLIC_KEY_ID)
                         throw new KeeperError(`Server rejected the custom server public key (id ${currentKeyId}). The server suggested key id ${suggestedKeyId}. Please update your IL5 KSM configuration.`)
+                    }
+                    if (typeof suggestedKeyId !== 'number' || !Number.isInteger(suggestedKeyId) || suggestedKeyId <= 0) {
+                        throw new KeeperError(`Server key error response contains invalid key_id: ${JSON.stringify(suggestedKeyId)}`)
                     }
                     if (!(suggestedKeyId in keeperPublicKeys)) {
                         throw new KeeperError(`Server suggested unsupported key id ${suggestedKeyId}; this SDK version supports key ids 7-18`)
--- a/sdk/javascript/packages/core/src/keeper.ts
+++ b/sdk/javascript/packages/core/src/keeper.ts
@@ -829,8 +829,11 @@ const postQuery = async (options: SecretManagerOptions, path: string, payload: A
                         const currentKeyId = await options.storage.getString(KEY_SERVER_PUBLIC_KEY_ID)
                         throw new KeeperError(`Server rejected the custom server public key (id ${currentKeyId}). The server suggested key id ${suggestedKeyId}. Please update your IL5 KSM configuration.`)
                     }
+                    if (!(suggestedKeyId in keeperPublicKeys)) {
+                        throw new KeeperError(`Server suggested unsupported key id ${suggestedKeyId}; this SDK version supports key ids 7-18`)
+                    }
                     if (keyRotationAttempt >= MAX_KEY_ROTATION_RETRIES) {
-                        throw new KeeperError(`Server key rotation exhausted ${MAX_KEY_ROTATION_RETRIES} retries; suggested key id ${suggestedKeyId} was not accepted`)
+                        throw new KeeperError(`Server key rotation exhausted ${MAX_KEY_ROTATION_RETRIES} retries; transmission key id ${transmissionKey.publicKeyId} was not accepted`)
                     }
                     await options.storage.saveString(KEY_SERVER_PUBLIC_KEY_ID, suggestedKeyId.toString())
                     keyRotationAttempt++
--- a/sdk/javascript/packages/core/src/keeper.ts
+++ b/sdk/javascript/packages/core/src/keeper.ts
@@ -1,7 +1,7 @@
 import {EncryptedPayload, KeeperHttpResponse, KeyValueStorage, platform, TransmissionKey} from './platform'
 import {webSafe64FromBytes, webSafe64ToBytes, tryParseInt} from './utils'
 import {parseNotation} from './notation'
-import {KeeperThrottleError} from './errors'
+import {KeeperError, KeeperThrottleError} from './errors'
 
 export {KeyValueStorage} from './platform'
 
@@ -820,15 +820,19 @@ const postQuery = async (options: SecretManagerOptions, path: string, payload: A
                 let errorObj: KeeperApiError | null = null
                 try { errorObj = JSON.parse(errorMessage) } catch {}
                 if (errorObj?.error === 'key') {
+                    const suggestedKeyId = errorObj.key_id
+                    if (typeof suggestedKeyId !== 'number' || !Number.isInteger(suggestedKeyId) || suggestedKeyId <= 0) {
+                        throw new KeeperError(`Server key error response contains invalid key_id: ${JSON.stringify(suggestedKeyId)}`)
+                    }
                     const customKey = await options.storage.getString(KEY_SERVER_PUBLIC_KEY)
                     if (customKey) {
                         const currentKeyId = await options.storage.getString(KEY_SERVER_PUBLIC_KEY_ID)
-                        throw new Error(`Server rejected the custom server public key (id ${currentKeyId}). The server suggested key id ${errorObj.key_id}. Please update your IL5 KSM configuration.`)
+                        throw new KeeperError(`Server rejected the custom server public key (id ${currentKeyId}). The server suggested key id ${suggestedKeyId}. Please update your IL5 KSM configuration.`)
                     }
                     if (keyRotationAttempt >= MAX_KEY_ROTATION_RETRIES) {
-                        throw new Error(`Server key rotation exhausted ${MAX_KEY_ROTATION_RETRIES} retries; suggested key id ${errorObj.key_id} was not accepted`)
+                        throw new KeeperError(`Server key rotation exhausted ${MAX_KEY_ROTATION_RETRIES} retries; suggested key id ${suggestedKeyId} was not accepted`)
                     }
-                    await options.storage.saveString(KEY_SERVER_PUBLIC_KEY_ID, errorObj.key_id!.toString())
+                    await options.storage.saveString(KEY_SERVER_PUBLIC_KEY_ID, suggestedKeyId.toString())
                     keyRotationAttempt++
                     continue
                 }
--- a/sdk/javascript/packages/core/src/keeper.ts
+++ b/sdk/javascript/packages/core/src/keeper.ts
@@ -824,13 +824,14 @@ const postQuery = async (options: SecretManagerOptions, path: string, payload: A
                     const customKey = await options.storage.getString(KEY_SERVER_PUBLIC_KEY)
                     if (customKey) {
                         const currentKeyId = await options.storage.getString(KEY_SERVER_PUBLIC_KEY_ID)
-                        throw new KeeperError(`Server rejected the custom server public key (id ${currentKeyId}). The server suggested key id ${suggestedKeyId}. Please update your IL5 KSM configuration.`)
+                        throw new KeeperError(`Server rejected the custom server public key (id ${currentKeyId ?? transmissionKey.publicKeyId}). The server suggested key id ${suggestedKeyId}. Please update your IL5 KSM configuration.`)
                     }
                     if (typeof suggestedKeyId !== 'number' || !Number.isInteger(suggestedKeyId) || suggestedKeyId <= 0) {
                         throw new KeeperError(`Server key error response contains invalid key_id: ${JSON.stringify(suggestedKeyId)}`)
                     }
                     if (!(suggestedKeyId in keeperPublicKeys)) {
-                        throw new KeeperError(`Server suggested unsupported key id ${suggestedKeyId}; this SDK version supports key ids 7-18`)
+                        const supported = Object.keys(keeperPublicKeys)
+                        throw new KeeperError(`Server suggested unsupported key id ${suggestedKeyId}; this SDK version supports key ids ${supported[0]}-${supported[supported.length - 1]}`)
                     }
                     if (keyRotationAttempt >= MAX_KEY_ROTATION_RETRIES) {
                         throw new KeeperError(`Server key rotation exhausted ${MAX_KEY_ROTATION_RETRIES} retries; transmission key id ${transmissionKey.publicKeyId} was not accepted`)
--- a/sdk/javascript/packages/core/test/keeper.test.ts
+++ b/sdk/javascript/packages/core/test/keeper.test.ts
@@ -321,7 +321,7 @@ test('IL5 dynamic key - rotation suppression: server key_id hint ignored when se
     expect(await storage.getString('serverPublicKeyId')).toBe('20')
 })
 
-test('IL5 dynamic key - server key rotation retries are bounded, not infinite', async () => {
+test('key rotation - retries are bounded, not infinite', async () => {
     const storage = inMemoryStorage({})
     await initializeStorage(storage, FAKE_ONE_TIME_TOKEN, 'fake.keepersecurity.com')
     let calls = 0
@@ -341,7 +341,7 @@ test('IL5 dynamic key - server key rotation retries are bounded, not infinite', 
     expect(calls).toBe(4)
 })
 
-test('IL5 dynamic key - single key rotation succeeds on retry', async () => {
+test('key rotation - suggested key id is adopted and persisted', async () => {
     const storage = inMemoryStorage({})
     await initializeStorage(storage, FAKE_ONE_TIME_TOKEN, 'fake.keepersecurity.com')
     let calls = 0
@@ -369,7 +369,7 @@ test('IL5 dynamic key - single key rotation succeeds on retry', async () => {
     expect(await storage.getString('serverPublicKeyId')).toBe('8')
 })
 
-test('IL5 dynamic key - unsupported suggested key id is rejected, not persisted', async () => {
+test('key rotation - unsupported suggested key id is rejected, not persisted', async () => {
     const storage = inMemoryStorage({})
     await initializeStorage(storage, FAKE_ONE_TIME_TOKEN, 'fake.keepersecurity.com')
     let calls = 0
--- a/sdk/javascript/packages/core/test/keeper.test.ts
+++ b/sdk/javascript/packages/core/test/keeper.test.ts
@@ -321,27 +321,38 @@ test('IL5 dynamic key - server key rotation retries are bounded, not infinite', 
     const storage = inMemoryStorage({})
     await initializeStorage(storage, 'YyIhK5wXFHj36wGBAOmBsxI3v5rIruINrC8KXjyM58c', 'fake.keepersecurity.com')
     let calls = 0
+    const enc = new TextEncoder()
+    const keyError = { statusCode: 400, data: enc.encode(JSON.stringify({ error: 'key', key_id: 7 })), headers: [] }
     const options: SecretManagerOptions = {
         storage,
-        queryFunction: async () => {
-            calls++
-            // Safety net only: without a bound, this branch never throws a matching error on its
-            // own and the loop would otherwise run until the test times out.
-            if (calls > 50) {
-                throw new Error('TEST_GUARD_UNBOUNDED_LOOP')
-            }
-            // key_id 7 is a real supported key number so the loop keeps re-entering the
-            // key-rotation branch itself, rather than tripping over an unrelated "unsupported
-            // key number" error from generateTransmissionKey.
-            return {
-                statusCode: 400,
-                data: new TextEncoder().encode(JSON.stringify({ error: 'key', key_id: 7 })),
-                headers: []
-            }
-        }
+        // key_id 7 is a real supported key number so each retry re-enters the key-rotation branch
+        // rather than tripping over an unrelated "unsupported key number" error.
+        queryFunction: async () => { calls++; return keyError }
     }
     await expect(getSecrets(options)).rejects.toThrow(/key rotation exhausted/i)
-    expect(calls).toBeLessThanOrEqual(10)
+    // MAX_KEY_ROTATION_RETRIES = 3: initial attempt + 3 retries = 4 total calls.
+    expect(calls).toBe(4)
+})
+
+test('IL5 dynamic key - single key rotation succeeds on retry', async () => {
+    const storage = inMemoryStorage({})
+    await initializeStorage(storage, 'YyIhK5wXFHj36wGBAOmBsxI3v5rIruINrC8KXjyM58c', 'fake.keepersecurity.com')
+    let calls = 0
+    const enc = new TextEncoder()
+    const emptyResponse = enc.encode(JSON.stringify({ records: [], folders: [], expiresOn: 0, warnings: [] }))
+    const options: SecretManagerOptions = {
+        storage,
+        queryFunction: async (_url, tk) => {
+            calls++
+            if (calls === 1) {
+                return { statusCode: 400, data: enc.encode(JSON.stringify({ error: 'key', key_id: 7 })), headers: [] }
+            }
+            return { statusCode: 200, data: await platform.encryptWithKey(emptyResponse, tk.key), headers: [] }
+        }
+    }
+    const secrets = await getSecrets(options)
+    expect(secrets.records).toEqual([])
+    expect(calls).toBe(2)
 })
 
 test('stale pinned server key: diagnostic message propagates to caller, key preserved', async () => {
--- a/sdk/javascript/packages/core/test/keeper.test.ts
+++ b/sdk/javascript/packages/core/test/keeper.test.ts
@@ -317,6 +317,33 @@ test('IL5 dynamic key - rotation suppression: server key_id hint ignored when se
     expect(await storage.getString('serverPublicKeyId')).toBe('20')
 })
 
+test('IL5 dynamic key - server key rotation retries are bounded, not infinite', async () => {
+    const storage = inMemoryStorage({})
+    await initializeStorage(storage, 'YyIhK5wXFHj36wGBAOmBsxI3v5rIruINrC8KXjyM58c', 'fake.keepersecurity.com')
+    let calls = 0
+    const options: SecretManagerOptions = {
+        storage,
+        queryFunction: async () => {
+            calls++
+            // Safety net only: without a bound, this branch never throws a matching error on its
+            // own and the loop would otherwise run until the test times out.
+            if (calls > 50) {
+                throw new Error('TEST_GUARD_UNBOUNDED_LOOP')
+            }
+            // key_id 7 is a real supported key number so the loop keeps re-entering the
+            // key-rotation branch itself, rather than tripping over an unrelated "unsupported
+            // key number" error from generateTransmissionKey.
+            return {
+                statusCode: 400,
+                data: new TextEncoder().encode(JSON.stringify({ error: 'key', key_id: 7 })),
+                headers: []
+            }
+        }
+    }
+    await expect(getSecrets(options)).rejects.toThrow(/key rotation exhausted/i)
+    expect(calls).toBeLessThanOrEqual(10)
+})
+
 test('stale pinned server key: diagnostic message propagates to caller, key preserved', async () => {
     const fakeKey = 'BK9w6TZFxE6nFNbMfIpULCup2a8xc6w2tUTABjxny7yFmxW0dAEojwC6j6zb5nTlmb1dAx8nwo3qF7RPYGmloRM'
     const storage = inMemoryStorage({})
--- a/sdk/javascript/packages/core/test/keeper.test.ts
+++ b/sdk/javascript/packages/core/test/keeper.test.ts
@@ -10,6 +10,10 @@ import {
 
 import * as fs from 'fs'
 
+const FAKE_ONE_TIME_TOKEN = 'YyIhK5wXFHj36wGBAOmBsxI3v5rIruINrC8KXjyM58c'
+
+const keyErrorResponse = (keyId: number) => JSON.stringify({ error: 'key', key_id: keyId })
+
 test('Get secrets e2e', async () => {
 
     const responses: { transmissionKey: string, data: string, statusCode: number } [] = JSON.parse(fs.readFileSync('../../../fake_data.json').toString())
@@ -319,15 +323,18 @@ test('IL5 dynamic key - rotation suppression: server key_id hint ignored when se
 
 test('IL5 dynamic key - server key rotation retries are bounded, not infinite', async () => {
     const storage = inMemoryStorage({})
-    await initializeStorage(storage, 'YyIhK5wXFHj36wGBAOmBsxI3v5rIruINrC8KXjyM58c', 'fake.keepersecurity.com')
+    await initializeStorage(storage, FAKE_ONE_TIME_TOKEN, 'fake.keepersecurity.com')
     let calls = 0
     const enc = new TextEncoder()
-    const keyError = { statusCode: 400, data: enc.encode(JSON.stringify({ error: 'key', key_id: 7 })), headers: [] }
     const options: SecretManagerOptions = {
         storage,
-        // key_id 7 is a real supported key number so each retry re-enters the key-rotation branch
-        // rather than tripping over an unrelated "unsupported key number" error.
-        queryFunction: async () => { calls++; return keyError }
+        queryFunction: async () => {
+            calls++
+            if (calls > 50) {
+                throw new Error('runaway loop detected in key rotation retry')
+            }
+            return { statusCode: 400, data: enc.encode(keyErrorResponse(7)), headers: [] }
+        }
     }
     await expect(getSecrets(options)).rejects.toThrow(/key rotation exhausted/i)
     // MAX_KEY_ROTATION_RETRIES = 3: initial attempt + 3 retries = 4 total calls.
@@ -336,7 +343,7 @@ test('IL5 dynamic key - server key rotation retries are bounded, not infinite', 
 
 test('IL5 dynamic key - single key rotation succeeds on retry', async () => {
     const storage = inMemoryStorage({})
-    await initializeStorage(storage, 'YyIhK5wXFHj36wGBAOmBsxI3v5rIruINrC8KXjyM58c', 'fake.keepersecurity.com')
+    await initializeStorage(storage, FAKE_ONE_TIME_TOKEN, 'fake.keepersecurity.com')
     let calls = 0
     const enc = new TextEncoder()
     const emptyResponse = enc.encode(JSON.stringify({ records: [], folders: [], expiresOn: 0, warnings: [] }))
@@ -344,31 +351,45 @@ test('IL5 dynamic key - single key rotation succeeds on retry', async () => {
         storage,
         queryFunction: async (_url, tk) => {
             calls++
-            if (calls === 1) {
-                return { statusCode: 400, data: enc.encode(JSON.stringify({ error: 'key', key_id: 7 })), headers: [] }
+            if (calls > 50) {
+                throw new Error('runaway loop detected in key rotation retry')
             }
+            if (calls === 1) {
+                return { statusCode: 400, data: enc.encode(keyErrorResponse(8)), headers: [] }
+            }
+            // Verify the rotation was adopted: second request should use key_id 8.
+            expect(tk.publicKeyId).toBe(8)
             return { statusCode: 200, data: await platform.encryptWithKey(emptyResponse, tk.key), headers: [] }
         }
     }
     const secrets = await getSecrets(options)
     expect(secrets.records).toEqual([])
     expect(calls).toBe(2)
+    // Verify the suggested key_id 8 was persisted to storage.
+    expect(await storage.getString('serverPublicKeyId')).toBe('8')
 })
 
 test('stale pinned server key: diagnostic message propagates to caller, key preserved', async () => {
     const fakeKey = 'BK9w6TZFxE6nFNbMfIpULCup2a8xc6w2tUTABjxny7yFmxW0dAEojwC6j6zb5nTlmb1dAx8nwo3qF7RPYGmloRM'
     const storage = inMemoryStorage({})
-    await initializeStorage(storage, 'YyIhK5wXFHj36wGBAOmBsxI3v5rIruINrC8KXjyM58c', 'fake.keepersecurity.com')
+    await initializeStorage(storage, FAKE_ONE_TIME_TOKEN, 'fake.keepersecurity.com')
     await storage.saveString('serverPublicKey', fakeKey)
     await storage.saveString('serverPublicKeyId', '20')
-    const keyError = JSON.stringify({ error: 'key', key_id: 7 })
+    let calls = 0
+    const enc = new TextEncoder()
     const options: SecretManagerOptions = {
         storage,
-        queryFunction: async () => ({
-            statusCode: 400,
-            data: new TextEncoder().encode(keyError),
-            headers: []
-        })
+        queryFunction: async () => {
+            calls++
+            if (calls > 50) {
+                throw new Error('runaway loop detected')
+            }
+            return {
+                statusCode: 400,
+                data: enc.encode(keyErrorResponse(7)),
+                headers: []
+            }
+        }
     }
     await expect(getSecrets(options)).rejects.toThrow(/Server rejected the custom server public key/)
     await expect(getSecrets(options)).rejects.toThrow(/Please update your IL5 KSM configuration/)
--- a/sdk/javascript/packages/core/test/keeper.test.ts
+++ b/sdk/javascript/packages/core/test/keeper.test.ts
@@ -369,6 +369,27 @@ test('IL5 dynamic key - single key rotation succeeds on retry', async () => {
     expect(await storage.getString('serverPublicKeyId')).toBe('8')
 })
 
+test('IL5 dynamic key - unsupported suggested key id is rejected, not persisted', async () => {
+    const storage = inMemoryStorage({})
+    await initializeStorage(storage, FAKE_ONE_TIME_TOKEN, 'fake.keepersecurity.com')
+    let calls = 0
+    const enc = new TextEncoder()
+    const options: SecretManagerOptions = {
+        storage,
+        queryFunction: async () => {
+            calls++
+            if (calls > 50) {
+                throw new Error('runaway loop detected in key rotation retry')
+            }
+            return { statusCode: 400, data: enc.encode(keyErrorResponse(99)), headers: [] }
+        }
+    }
+    await expect(getSecrets(options)).rejects.toThrow(/unsupported key id 99/)
+    // Rejected before the retry loop persists anything: one request, config untouched.
+    expect(calls).toBe(1)
+    expect(await storage.getString('serverPublicKeyId')).toBeUndefined()
+})
+
 test('stale pinned server key: diagnostic message propagates to caller, key preserved', async () => {
     const fakeKey = 'BK9w6TZFxE6nFNbMfIpULCup2a8xc6w2tUTABjxny7yFmxW0dAEojwC6j6zb5nTlmb1dAx8nwo3qF7RPYGmloRM'
     const storage = inMemoryStorage({})


sdk/javascript/packages/core/src/keeper.ts
```diff
@@ -816,7 +821,11 @@ const postQuery = async (options: SecretManagerOptions, path: string, payload: A
                         const currentKeyId = await options.storage.getString(KEY_SERVER_PUBLIC_KEY_ID)
                         throw new Error(`Server rejected the custom server public key (id ${currentKeyId}). The server suggested key id ${errorObj.key_id}. Please update your IL5 KSM configuration.`)
                     }
+                    if (keyRotationAttempt >= MAX_KEY_ROTATION_RETRIES) {
+                        throw new Error(`Server key rotation exhausted ${MAX_KEY_ROTATION_RETRIES} retries; suggested key id ${errorObj.key_id} was not accepted`)
+                    }
                     await options.storage.saveString(KEY_SERVER_PUBLIC_KEY_ID, errorObj.key_id!.toString())
```
mgallego-keeper: **[blocking] Validate `errorObj.key_id` before persisting it.** Two failure modes flow through this line, both reproduced on the built bundle, and both live in the exact branch this PR hardens:

1. **Config poisoning, and the new bound never engages.** With `{"error":"key","key_id":19}` (or `0`, `8.5`, `"abc"`), `'19'` is saved here; the next iteration throws `Key number 19 is not supported` inside `generateTransmissionKey`, before any request is sent, so the exhaustion check above is never reached. First call: 1 network request, then failure. Every later call: 0 network requests, failure, forever. With file-backed `localConfigStorage` the poisoned id survives restarts until the JSON is hand-edited.
2. **Raw `TypeError` on a missing id.** `{"error":"key"}` without `key_id` makes this line throw `TypeError: Cannot read properties of undefined (reading 'toString')` out of `getSecrets` (the `!` has no runtime effect; `KeeperApiError.key_id` is optional). The new exhaustion message has the same soft spot: it prints `suggested key id undefined was not accepted` when the final response lacks the field.

Python guards both before persisting (raises on a blank `key_id`, checks membership in `keeper_public_keys`). The same shape here closes both:

```ts
if (!errorObj.key_id || !keeperPublicKeys[errorObj.key_id]) {
    throw new Error(`Server suggested an unsupported public key id ${errorObj.key_id}`)
}
await options.storage.saveString(KEY_SERVER_PUBLIC_KEY_ID, errorObj.key_id.toString())
```

The new test's own comment (lines 333-335) picks `key_id: 7` specifically to route around the unsupported-id variant, so that variant currently ships untested; a companion test with `key_id: 99` would pin the validation.

sdk/javascript/packages/core/src/keeper.ts
```diff
@@ -816,7 +821,11 @@ const postQuery = async (options: SecretManagerOptions, path: string, payload: A
                         const currentKeyId = await options.storage.getString(KEY_SERVER_PUBLIC_KEY_ID)
                         throw new Error(`Server rejected the custom server public key (id ${currentKeyId}). The server suggested key id ${errorObj.key_id}. Please update your IL5 KSM configuration.`)
                     }
+                    if (keyRotationAttempt >= MAX_KEY_ROTATION_RETRIES) {
+                        throw new Error(`Server key rotation exhausted ${MAX_KEY_ROTATION_RETRIES} retries; suggested key id ${errorObj.key_id} was not accepted`)
```
mgallego-keeper: **Type this error.** The parallel exhaustion case in this same loop throws `KeeperThrottleError` (line 807), `errors.ts` documents `KeeperError` as the base for SDK-raised errors (with dedicated `instanceof` catchability tests in throttle.test.ts), and the 17.5.0 changelog tracks migrating the remaining `throw new Error(...)` sites under KSM-1024; this adds a brand-new one. It lands on a terminal condition callers will want to classify: a retry wrapper keyed on error class will blind-retry a call that fails identically every time (each fresh call gets a new 4-request budget), and message-regex matching is the only alternative. `KeeperError`, or a `KeeperKeyRotationError extends KeeperError` mirroring `KeeperThrottleError`, keeps the pattern consistent.

sdk/javascript/packages/core/src/keeper.ts
```diff
@@ -816,7 +821,11 @@ const postQuery = async (options: SecretManagerOptions, path: string, payload: A
                         const currentKeyId = await options.storage.getString(KEY_SERVER_PUBLIC_KEY_ID)
                         throw new Error(`Server rejected the custom server public key (id ${currentKeyId}). The server suggested key id ${errorObj.key_id}. Please update your IL5 KSM configuration.`)
                     }
+                    if (keyRotationAttempt >= MAX_KEY_ROTATION_RETRIES) {
```
mgallego-keeper, line 836: Because this check runs before the save, the thrown message blames the current response's `key_id`, which was never sent in any request, and drops it, while storage keeps the id the server just rejected. Reproduced with progressing suggestions 8, 9, 10, 11: requests go out with ids [7, 8, 9, 10], the throw says "suggested key id 11 was not accepted" though 11 was never attempted, and storage retains '10', so the caller's next call re-sends the known-bad 10 and burns a round trip rediscovering 11. In the same-id ping-pong case the wording is coincidentally right, so this bites exactly the cases where a progressing chain needs diagnosing. Either persist the last suggestion before throwing, or word the message around the id actually rejected (`transmissionKey.publicKeyId` is in scope).

sdk/javascript/packages/core/src/keeper.ts
```diff
@@ -816,7 +821,11 @@ const postQuery = async (options: SecretManagerOptions, path: string, payload: A
...
+                        throw new Error(`Server key rotation exhausted ${MAX_KEY_ROTATION_RETRIES} retries; suggested key id ${errorObj.key_id} was not accepted`)
+                    }
                     await options.storage.saveString(KEY_SERVER_PUBLIC_KEY_ID, errorObj.key_id!.toString())
+                    keyRotationAttempt++
```
mgallego-keeper, line 840: Three non-blocking observations on the counting approach:

1. **No-progress detection is free here.** When the response re-suggests the id that was just used (`errorObj.key_id === transmissionKey.publicKeyId`, which is exactly what the new test simulates), the loop is proven on the first response, yet the client still burns 3 more full encrypt/sign/POST round trips (each also counting toward the backend's 100-requests-per-10s throttle window). A same-id fast-fail, or a visited-id set, with the counter kept as a backstop, is strictly more precise.
2. **The counter never resets on progress**, so more than 3 legitimate, distinct rotations within a single call now hard-fail where the old code converged. Edge territory (mixed server pools mid-migration, or rotations landing during the multi-minute throttle backoffs possible in this same loop), but worth a conscious decision.
3. **Silent retries.** The throttle path warns per attempt (line 810); rotation says nothing until the terminal throw, so an exhausted run leaves no trail of which ids were tried. A mirrored `attempt N/3, key id X to Y` line would make this diagnosable from logs.

sdk/javascript/packages/core/src/keeper.ts
```diff
@@ -784,6 +788,7 @@ const postQuery = async (options: SecretManagerOptions, path: string, payload: A
     const url = `https://${hostName}/api/rest/sm/v1/${path}`
     const sleep = options.throttleSleep || ((ms: number) => new Promise<void>(resolve => setTimeout(resolve, ms)))
     let throttleAttempt = 0
+    let keyRotationAttempt = 0
```
mgallego-keeper, line 795: Adjacent interaction worth a look while in this function (pre-existing, fine as a follow-up ticket): lines 781-783 re-save `options.serverPublicKeyId` at the top of every call, so for configs that pin only the id (type-legal per `SecretManagerOptions`, and used by this suite's base64-config test with `serverPublicKeyId: '10'`) a completed rotation never sticks. Reproduced: with pinned '8' while the server wants 7, every single `getSecrets` call sends ids [8, 7], forever (storage ends at '7', then is clobbered back to '8' at the next call's start). That is permanent 2x request amplification, and with this PR one bounded retry burned per call; a pathological chain of more than 3 hops can now never converge across any number of calls, where pre-PR code converged within one call. The custom-key guard at line 819 checks `KEY_SERVER_PUBLIC_KEY` only, so none of this is covered by the pinned-key suppression.

sdk/javascript/packages/core/src/keeper.ts
```diff
@@ -816,7 +821,11 @@ const postQuery = async (options: SecretManagerOptions, path: string, payload: A
                         const currentKeyId = await options.storage.getString(KEY_SERVER_PUBLIC_KEY_ID)
                         throw new Error(`Server rejected the custom server public key (id ${currentKeyId}). The server suggested key id ${errorObj.key_id}. Please update your IL5 KSM configuration.`)
```
mgallego-keeper: Small diagnostic gap in the branch this PR hardens: when `serverPublicKey` is configured without `serverPublicKeyId`, the request goes out with the defaulted key id 7, but `getString` returns undefined for the never-saved id and this message prints "(id undefined)" (reproduced on the built bundle). Falling back to `transmissionKey.publicKeyId` (in scope) names the id actually used, so support chases the real mismatch instead of a phantom missing-id bug.

sdk/javascript/packages/core/test/keeper.test.ts
```diff
@@ -317,6 +317,33 @@ test('IL5 dynamic key - rotation suppression: server key_id hint ignored when se
...
+            // key number" error from generateTransmissionKey.
+            return {
+                statusCode: 400,
+                data: new TextEncoder().encode(JSON.stringify({ error: 'key', key_id: 7 })),
+                headers: []
+            }
+        }
+    }
+    await expect(getSecrets(options)).rejects.toThrow(/key rotation exhausted/i)
+    expect(calls).toBeLessThanOrEqual(10)
```
mgallego-keeper: As written, the two assertions cannot catch the regressions that matter: the call count is deterministically 4, so `toBeLessThanOrEqual(10)` lets a counter-reset bug that doubles requests to 8 pass, and a bound of 0 (rotation entirely broken) also passes, because "exhausted 0 retries" still matches `/key rotation exhausted/i` with `calls === 1`. Suggest `expect(calls).toBe(4)` (initial plus `MAX_KEY_ROTATION_RETRIES`; the constant is module-private, so hardcode with a comment), plus a happy-path companion (rotate once, then 200) so a legitimate single rotation staying functional is pinned too. That also lets the `> 50` guard and the `10` threshold collapse into one shared number.

sdk/javascript/packages/core/test/keeper.test.ts
```diff
@@ -317,6 +317,33 @@ test('IL5 dynamic key - rotation suppression: server key_id hint ignored when se
...
+            // own and the loop would otherwise run until the test times out.
+            if (calls > 50) {
+                throw new Error('TEST_GUARD_UNBOUNDED_LOOP')
+            }
+            // key_id 7 is a real supported key number so the loop keeps re-entering the
+            // key-rotation branch itself, rather than tripping over an unrelated "unsupported
+            // key number" error from generateTransmissionKey.
+            return {
+                statusCode: 400,
+                data: new TextEncoder().encode(JSON.stringify({ error: 'key', key_id: 7 })),
```
mgallego-keeper: Third copy in this file of the identical `{ error: 'key', key_id: 7 }` 400-response literal (siblings at lines 305-313 and 353-361), and third fake one-time-token literal (lines 34, 322, 350). throttle.test.ts already models the cleanup with its file-top `throttle403()` / `FAKE_TOKEN` helpers; a `keyErrorResponse(keyId = 7)` builder plus a hoisted token constant would keep the three from drifting.

mgallego-keeper (changes requested):
## Review summary

The fix is correct for what it targets and I verified it empirically before the detailed pass:

- Full core suite green on the PR head (56/56).
- The new test genuinely fails on unfixed code. One nuance for anyone re-running that check: the tests import the rollup-built `dist/` (package `main`), so the failure only reproduces via `npm test` (whose `pretest` rebuilds); a bare `npx jest` against a stale `dist` still passes on unfixed sources.
- With the fix, the exhaustion path throws after exactly 4 transport calls (initial plus 3 retries), matching the description.
- Mergeable-clean against the current base (post #1079).

Requesting changes for three items, detailed inline:

1. Validate `errorObj.key_id` before persisting it: config poisoning and a raw `TypeError` both live in the exact branch this PR hardens, and the new bound never engages on the poisoning path.
2. Throw a typed error (`KeeperError` or a `KeeperKeyRotationError`), matching `KeeperThrottleError` in the same loop and the KSM-1025/KSM-1024 direction.
3. Tighten the new test so it can catch bound regressions, plus a happy-path companion.

### Findings without an inline anchor (lines outside the diff)

- **No status gate on the rotation branch** (keeper.ts:818). The throttle branch is deliberately gated on 403, and its comment explains why; the key branch accepts any non-200 whose body parses to `{"error":"key"}`. Both platform `post` implementations resolve rather than throw on 5xx, so a gateway echoing a cached or fabricated key-shaped body on a 502 triggers spurious rotation, and with the unvalidated save above, persistent poisoning. Consider gating on the status the backend actually uses for key errors.
- **1000-byte truncation feeds the parse** (keeper.ts:799). A key error carried in a body over 1000 bytes fails `JSON.parse` in the bare catch, and rotation (bound included) silently deactivates; verified mechanically (calls=1, raw truncated text thrown). Real key-error bodies are tiny today, so this is future-proofing, not a blocker.
- **The new terminal error is swallowed on two caller paths.** `getSecrets2`'s `justBound` refetch (~keeper.ts:1073) only `console.error`s it and returns the first fetch's data as success (the refetch also gets a fresh retry budget, so up to 8 key-error round trips total), and `tryGetNotationResults` maps it to `[]`, indistinguishable from a field with no values. Terminating beats the old infinite hang; flagging so the `justBound` swallow is a decision rather than an accident.
- **Changelog**: no 17.6.0 section exists yet (KSM-1079/KSM-1084 also lack entries, so presumably batched at release), and the 17.5.0 KSM-880 entry ends with "Existing key-rotation retry behavior is unchanged", which stops being true once this merges. Worth catching at release cut.

### Cross-SDK

The same loop is live elsewhere; I filed siblings so this does not get rediscovered one SDK at a time: KSM-1247 (.NET: unbounded, and persists the suggested id unvalidated), KSM-1248 (Java/Kotlin: same two defects, despite bounding throttle retries in the same function), KSM-1249 (Python: unbounded; it already validates ids). Rust's variant was KSM-998 (deployed). All three are linked to KSM-1128.

### Minor, no action needed to merge

The test name says "IL5 dynamic key" but it exercises the default non-IL5 path; `MAX_KEY_ROTATION_RETRIES` lands between the two throttle constants and splits that group; the new test follows the file's existing pattern of not restoring `platform.getRandomBytes` stubs (file-wide, fine to punt).

sdk/javascript/packages/core/src/keeper.ts
mgallego-keeper: f62fb4ce addresses half of this. Fixed and verified against the rebuilt bundle: the shape guard rejects non-number / non-integer / non-positive ids with a typed `KeeperError` before anything is written, and the missing-`key_id` `TypeError` is gone (`{"error":"key"}` now yields `KeeperError` with `invalid key_id: undefined`, storage untouched).

Still open: membership. The original repro reproduces unchanged on f62fb4ce: `{"error":"key","key_id":99}` passes the shape guard, `'99'` is persisted, and the client is bricked (first `getSecrets` makes exactly 1 network request; every later one makes 0 and throws `Key number 99 is not supported`, a plain `Error`, with storage stuck at `'99'`). Carried into the round-2 review with a placement-corrected suggestion: membership belongs after the pinned-custom-key check, since isolated deployments legitimately run ids outside the public table.

sdk/javascript/packages/core/src/keeper.ts
mgallego-keeper: Resolved by f62fb4ce: all three throws in this branch are now `KeeperError`, and I verified `instanceof KeeperError` against the rebuilt bundle on the exhaustion path (throws after exactly 4 transport calls). Thanks.

sdk/javascript/packages/core/src/keeper.ts
mgallego-keeper: Not picked up in f62fb4ce; still stands but non-blocking. The exhaustion message still names the final suggestion, which was never sent in any request, and drops it while storage keeps the id the server just rejected. Cheap to fold into the membership fix; also fine to punt.

sdk/javascript/packages/core/src/keeper.ts
mgallego-keeper: Not picked up in f62fb4ce; these were explicitly non-blocking and I am fine punting all three (same-id fast-fail, reset-on-progress, per-attempt log line). One extra data point for whenever this is revisited: each silent rotation retry also rewrites the whole config file, so the per-attempt log line would pair naturally with that write. Resolving.

sdk/javascript/packages/core/src/keeper.ts
mgallego-keeper: Out of this PR's scope as noted in round 1; keeping the thread open until it has a home. Happy to file a follow-up ticket mirroring the KSM-1247/KSM-1248/KSM-1249 series.

sdk/javascript/packages/core/src/keeper.ts
mgallego-keeper: Not picked up in f62fb4ce; still stands, non-blocking. Now that the branch computes `suggestedKeyId` up front and `transmissionKey` is in scope, the fallback is a one-liner: `id ${currentKeyId ?? transmissionKey.publicKeyId}`.

sdk/javascript/packages/core/test/keeper.test.ts
mgallego-keeper: The tightening ask is resolved by f62fb4ce: exact `toBe(4)` plus the new single-rotation happy-path test, 57/57 green in a clean worktree. Resolving this thread, but two new test gaps surfaced while verifying the rework, continued as round-2 inline comments: dropping the 50-call guard turned a bound regression into a silent jest hang (the retry loop is microtask-only, so `testTimeout` can never fire), and neither test can detect a broken rotation adoption (both use `key_id: 7`, the default).

sdk/javascript/packages/core/test/keeper.test.ts
mgallego-keeper: Partially picked up in f62fb4ce (the response literal is hoisted within the bounded test); the file-wide helper is punted, which is fine on a release branch. Waiving and resolving.

sdk/javascript/packages/core/src/keeper.ts
```diff
@@ -811,12 +816,20 @@ const postQuery = async (options: SecretManagerOptions, path: string, payload: A
                 let errorObj: KeeperApiError | null = null
                 try { errorObj = JSON.parse(errorMessage) } catch {}
                 if (errorObj?.error === 'key') {
+                    const suggestedKeyId = errorObj.key_id
+                    if (typeof suggestedKeyId !== 'number' || !Number.isInteger(suggestedKeyId) || suggestedKeyId <= 0) {
```
mgallego-keeper: **[blocking, carried from round 1]** This guard needs a membership check as well, but placed correctly rather than merged into this line: isolated/IL5 deployments legitimately run key ids outside the public table, so membership must sit after the pinned-custom-key check (Python gates it the same way: `str(key_id) not in keeper_public_keys and not custom_key`). Simplest shape: move this whole guard below the `customKey` branch and extend it:

```ts
if (typeof suggestedKeyId !== 'number' || !Number.isInteger(suggestedKeyId) || suggestedKeyId <= 0 || !(suggestedKeyId in keeperPublicKeys)) {
    throw new KeeperError(`Server key error response contains an invalid or unsupported key_id: ${JSON.stringify(suggestedKeyId)}`)
}
```

Shape alone still admits the round-1 repro, re-verified on the rebuilt f62fb4ce bundle: `key_id: 99` passes, `'99'` is persisted at the save below, and `generateTransmissionKey` then throws `Key number 99 is not supported` (a plain `Error`) before any request, on this call and every later one, storage stuck at `'99'` (a fractional-exponent id like `1e21` even round-trips as the string `'1e+21'`). `keeperPublicKeys` is safe to read here: `generateTransmissionKey` already dereferenced it earlier in this same iteration.

Moving the guard below the `customKey` throw also matters on its own: as committed, a pinned IL5 deployment whose server omits or malforms `key_id` now gets the generic `invalid key_id` error instead of the actionable `Please update your IL5 KSM configuration` diagnostic it got before this commit, so the PR description's "custom-key-pinned path is unchanged" note no longer holds (the pinned throw's class also changed to `KeeperError`, which is fine, but worth reflecting in the description). Optional while here: coercing numeric strings (`Number(suggestedKeyId)`) before validating would keep tolerating a quoted `"key_id": "8"`, which the pre-PR code accepted and rotated on; cross-SDK evidence says the wire format is a JSON number, so this is belt-and-suspenders.

sdk/javascript/packages/core/test/keeper.test.ts
```diff
@@ -317,6 +317,44 @@ test('IL5 dynamic key - rotation suppression: server key_id hint ignored when se
...
+    const storage = inMemoryStorage({})
+    await initializeStorage(storage, 'YyIhK5wXFHj36wGBAOmBsxI3v5rIruINrC8KXjyM58c', 'fake.keepersecurity.com')
+    let calls = 0
+    const enc = new TextEncoder()
+    const keyError = { statusCode: 400, data: enc.encode(JSON.stringify({ error: 'key', key_id: 7 })), headers: [] }
+    const options: SecretManagerOptions = {
+        storage,
+        // key_id 7 is a real supported key number so each retry re-enters the key-rotation branch
+        // rather than tripping over an unrelated "unsupported key number" error.
+        queryFunction: async () => { calls++; return keyError }
```
mgallego-keeper: Restore a runaway guard in this stub (and the same in the happy-path test's mock below): with the 50-call guard gone, a regression of the retry bound no longer fails this test, it hangs the whole jest run silently. Verified by patching `MAX_KEY_ROTATION_RETRIES` to `Infinity` in the built bundle: zero output, full CPU, and `--testTimeout` never fires, because every await in the retry loop (sync crypto wrappers, resolved-promise storage, this stub) settles as a microtask, so the timer that would enforce the timeout is starved; the run had to be killed. A two-line guard keeps the failure mode a clean assertion:

```ts
if (calls > 10) throw new Error('runaway rotation retry loop')
```

sdk/javascript/packages/core/test/keeper.test.ts
```diff
@@ -317,6 +317,44 @@ test('IL5 dynamic key - rotation suppression: server key_id hint ignored when se
...
+    await initializeStorage(storage, 'YyIhK5wXFHj36wGBAOmBsxI3v5rIruINrC8KXjyM58c', 'fake.keepersecurity.com')
+    let calls = 0
+    const enc = new TextEncoder()
+    const emptyResponse = enc.encode(JSON.stringify({ records: [], folders: [], expiresOn: 0, warnings: [] }))
+    const options: SecretManagerOptions = {
+        storage,
+        queryFunction: async (_url, tk) => {
+            calls++
+            if (calls === 1) {
+                return { statusCode: 400, data: enc.encode(JSON.stringify({ error: 'key', key_id: 7 })), headers: [] }
```
mgallego-keeper: Precision gap, verified by mutation: deleting the adoption line (`saveString` in the rotation branch) leaves all 57 tests green, because `key_id: 7` equals the default key id, so this test proves the loop converges but not that the suggested key is adopted. Suggesting `8` and asserting adoption locks it in: the mock already receives `tk`, so `expect(tk.publicKeyId).toBe(8)` when `calls === 2`, plus `expect(await storage.getString('serverPublicKeyId')).toBe('8')` after. Adoption does work today (probed on the rebuilt bundle: suggest 8, second request uses `publicKeyId` 8, storage ends `'8'`); this is about making the suite able to notice if it ever stops.

mgallego-keeper (changes requested):
## Re-review of f62fb4ce

Same verification method as round 1: isolated worktree at f62fb4ce, `npm ci` + `npm test` (which rebuilds `dist/` via `pretest`), behavioral probes against the rebuilt bundle, plus mutation checks on the new tests.

**Verified fixed:**

- Missing/malformed `key_id`: `{"error":"key"}` now throws a typed `KeeperError` (`invalid key_id: undefined`) with storage untouched; the raw `TypeError` is gone.
- Typed errors: all three throws in the branch are `KeeperError`, `instanceof` verified on the exhaustion path, still exactly 4 transport calls. With both counters in play the loop is now fully bounded on every `continue` path (worst case 9 requests alternating throttle and key errors).
- Tests: `toBe(4)` and the rotate-once happy path are in; 57/57 green. Adoption itself works end to end (probe: suggest key 8, the second request goes out with `publicKeyId` 8, storage ends at `'8'`).

**Still open (the round-1 blocking item):** the suggested id is persisted with shape validation only, no membership check, and the round-1 repro reproduces unchanged on f62fb4ce: `{"error":"key","key_id":99}` passes the guard, `'99'` is persisted, and the client is bricked: 1 network request on the first `getSecrets`, 0 on every later one (`Key number 99 is not supported`, a plain `Error`, so the one path that stays broken also dodges the new typing), until the stored `serverPublicKeyId` is hand-edited. File-backed configs carry it across restarts.

The realistic trigger is worth spelling out: the supported table today is ids 7-18 (twelve keys in `initialize()`), and key 18 was itself added later (KSM-739), so "the backend rotates to an id this SDK version does not ship" is a when, not an if, for long-lived deployments. With a membership check, that day produces a clean, typed, recoverable upgrade-the-SDK error on every call; without it, config surgery. The inline comment has the placement-corrected fix: the guard must sit after the pinned-custom-key branch (mirroring Python's `and not custom_key` gating), which also repairs a small diagnostic regression this commit introduced for pinned IL5 configs.

**Two new test gaps found while verifying the rework** (inline comments): a bound regression now hangs jest silently instead of failing (demonstrated by patching the bound to `Infinity`: the retry loop is microtask-only, so the timer-based `testTimeout` never fires; the run sat at full CPU until killed), and neither rotation test can detect a broken adoption (mutation check: deleting the `saveString` at the rotation branch leaves all 57 tests green, because `key_id: 7` equals the default key).

**Pre-existing, noting only (no action needed here):** `if (response.data)` at the top of the error handling treats an empty `Uint8Array` as a present body (empty typed arrays are truthy), so a non-200 with a zero-length body throws `Error('')` with an empty message on the browser platform; and the branch's generic fallback `throw new Error(errorMessage)` stays untyped, which is fine to leave for the KSM-1024 migration.

Everything else from round 1 is resolved or explicitly waived; per-thread dispositions posted on the round-1 comments.

sdk/javascript/packages/core/src/keeper.ts
mgallego-keeper: Complete as of c3519324. The membership half landed at `keeper.ts:828`, and the round-1 repro is closed: `{"error":"key","key_id":99}` now makes exactly 1 network request, throws a typed `KeeperError`, and storage is never written (probed on the rebuilt bundle: `calls=1 stored=undefined`), where before it persisted `'99'` and bricked the client on every later call.

Boundaries check out too: `6` and `1e21` are both rejected (the latter closes the `'1e+21'` string round-trip I flagged in round 2), and `18` is still accepted and adopted, so nothing at the top of the table regressed. Resolving.

sdk/javascript/packages/core/src/keeper.ts
mgallego-keeper: Both halves are addressed in c3519324, and the placement is the part I most wanted to confirm. Verified on the rebuilt bundle:

| suggested id | result |
|---|---|
| 99 | 1 call, typed `KeeperError`, storage untouched |
| 6 | rejected at the lower boundary |
| 1e21 | rejected (closes the `'1e+21'` round-trip from round 2) |
| 18 | accepted, adopted, `publicKeyId` numeric: no false positive at the top of the table |
| 99, **pinned IL5 config** | still `Please update your IL5 KSM configuration`, untouched by the new check |

That last row is the one that mattered: putting membership below the `customKey` branch means isolated deployments running out-of-table ids are unaffected, which is what the Python parity (`and not custom_key`) buys. `keeperPublicKeys` reduces to a plain `{7..18}` object, so `in` is an exact match, and the shape guard above guarantees the operand is a positive integer, which keeps the number-to-string coercion safe. Resolving.

One leftover from this thread, non-blocking, split into its own comment on the shape guard: that guard is still *above* the `customKey` branch, so a pinned config whose server omits or malforms `key_id` still gets the generic error rather than the IL5 diagnostic.

sdk/javascript/packages/core/src/keeper.ts
mgallego-keeper: Picked up in c3519324, and it resolves the misleading-id problem. The message now names `transmissionKey.publicKeyId`. Verified with alternating suggestions 8 and 9: requests go out as ids `[7, 8, 9, 8]` and the throw reads `Server key rotation exhausted 3 retries; transmission key id 8 was not accepted`, so it names an id that was actually sent rather than one that never left the process.

Storage still retains the last suggestion (`'8'`), which no longer worries me now that membership is validated: the retained id is guaranteed to be one the SDK can actually use. Resolving.

sdk/javascript/packages/core/src/keeper.ts
mgallego-keeper: Filed as **KSM-1255** (Bug, Major, unassigned, linked A-Relates to KSM-1128), so this finally has a home. Resolving the thread on that basis.

Re-verified against c3519324 while writing it up, and one new detail went into the ticket: this path also **bypasses the membership check this PR just added**. Passing `serverPublicKeyId: '99'` in options persists it unvalidated at 781-783, and the next `generateTransmissionKey` throws the untyped `Error("Key number 99 is not supported")` with zero network requests. The new validation guards the rotation branch only, so the caller-supplied route into a poisoned config is still open. That is KSM-1255's problem, not this PR's.

sdk/javascript/packages/core/src/keeper.ts
mgallego-keeper: Still not picked up as of c3519324, still non-blocking. The one-liner is unchanged if you want it while you are in this branch:

```ts
`Server rejected the custom server public key (id ${currentKeyId ?? transmissionKey.publicKeyId}).`
```

Leaving this open as the only cosmetic item outstanding. Equally fine to punt it to a follow-up if you would rather not touch the pinned path again in this PR.

sdk/javascript/packages/core/test/keeper.test.ts
mgallego-keeper: Restored in c3519324 (`calls > 50`, in all three rotation stubs), and it does exactly what it needed to. I re-ran the round-2 mutation to confirm: with `MAX_KEY_ROTATION_RETRIES` patched to `Infinity` and the bundle rebuilt, this test now **fails in 45ms** with `runaway loop detected in key rotation retry`. In round 2 that same mutation produced no output, pinned a core, never tripped `--testTimeout`, and had to be killed. The bound now has a regression test that reports rather than hangs. Resolving.

sdk/javascript/packages/core/test/keeper.test.ts
mgallego-keeper: Locked in by c3519324. Verified by mutation: deleting the `saveString` adoption line now fails this test with `Expected: 8, Received: 7`, where in round 2 that same deletion left all 57 tests green.

Switching the suggestion to `8` is what made it observable, and the trailing storage assertion (`'8'`) is a useful belt-and-braces in case the in-stub `expect` ever gets swallowed by a future change to how `queryFunction` errors are handled. Resolving.

sdk/javascript/packages/core/test/keeper.test.ts
```diff
@@ -317,20 +321,75 @@ test('IL5 dynamic key - rotation suppression: server key_id hint ignored when se
...
+            expect(tk.publicKeyId).toBe(8)
+            return { statusCode: 200, data: await platform.encryptWithKey(emptyResponse, tk.key), headers: [] }
+        }
+    }
+    const secrets = await getSecrets(options)
+    expect(secrets.records).toEqual([])
+    expect(calls).toBe(2)
+    // Verify the suggested key_id 8 was persisted to storage.
+    expect(await storage.getString('serverPublicKeyId')).toBe('8')
+})
```
mgallego-keeper, line 370: **[blocking] The membership check needs a regression test.** It is the fix for the blocker carried through both previous rounds, and nothing currently holds it in place: deleting the three lines at `src/keeper.ts:828-830` leaves this file at 18/18 green (mutation-verified on the rebuilt bundle). Someone tidying that branch later can drop it and the suite will not notice, which is roughly how the original defect survived.

This is the probe I used to verify your fix, and it drops in next to the two tests above. I confirmed it passes on c3519324 and fails with the check removed (it then reports `Key number 99 is not supported`, which is the poisoned-config path itself, so the test documents the bug it prevents):

```ts
test('IL5 dynamic key - unsupported suggested key id is rejected, not persisted', async () => {
    const storage = inMemoryStorage({})
    await initializeStorage(storage, FAKE_ONE_TIME_TOKEN, 'fake.keepersecurity.com')
    let calls = 0
    const enc = new TextEncoder()
    const options: SecretManagerOptions = {
        storage,
        queryFunction: async () => {
            calls++
            if (calls > 50) {
                throw new Error('runaway loop detected in key rotation retry')
            }
            return { statusCode: 400, data: enc.encode(keyErrorResponse(99)), headers: [] }
        }
    }
    await expect(getSecrets(options)).rejects.toThrow(/unsupported key id 99/)
    // Rejected before the retry loop persists anything: one request, config untouched.
    expect(calls).toBe(1)
    expect(await storage.getString('serverPublicKeyId')).toBeUndefined()
})
```

The `serverPublicKeyId` assertion is the half that matters most: it is what would have caught the original config-poisoning bug, and it is the property that cannot be observed from the thrown message alone.

sdk/javascript/packages/core/src/keeper.ts
```diff
@@ -811,12 +816,23 @@ const postQuery = async (options: SecretManagerOptions, path: string, payload: A
...
+                        throw new KeeperError(`Server key error response contains invalid key_id: ${JSON.stringify(suggestedKeyId)}`)
+                    }
                     const customKey = await options.storage.getString(KEY_SERVER_PUBLIC_KEY)
                     if (customKey) {
                         const currentKeyId = await options.storage.getString(KEY_SERVER_PUBLIC_KEY_ID)
-                        throw new Error(`Server rejected the custom server public key (id ${currentKeyId}). The server suggested key id ${errorObj.key_id}. Please update your IL5 KSM configuration.`)
+                        throw new KeeperError(`Server rejected the custom server public key (id ${currentKeyId}). The server suggested key id ${suggestedKeyId}. Please update your IL5 KSM configuration.`)
+                    }
+                    if (!(suggestedKeyId in keeperPublicKeys)) {
+                        throw new KeeperError(`Server suggested unsupported key id ${suggestedKeyId}; this SDK version supports key ids 7-18`)
```
mgallego-keeper: Non-blocking, and now tracked as **KSM-1257**, so none of this has to happen here.

This is the second membership check over the same table. `generateTransmissionKey:757-758` already rejects unbundled ids, as an untyped plain `Error("Key number 99 is not supported")`, and it is still reachable on this head (a stored or caller-supplied out-of-table id reaches it with zero network requests). So one condition now produces two messages and two error types, which is a small tax on support triage.

Separately, the literal `7-18` restates what `keeperPublicKeys` already knows (12 entries, built by the `reduce` in `initialize()`). That range has moved before: KSM-739 appended #18 in 17.4.0, and the next append silently makes this sentence wrong. Deriving it keeps it true by construction, since integer-like keys enumerate in ascending order:

```ts
const supported = Object.keys(keeperPublicKeys)
throw new KeeperError(`Server suggested unsupported key id ${suggestedKeyId}; this SDK version supports key ids ${supported[0]}-${supported[supported.length - 1]}`)
```

Worth noting the check itself is sound as written: `keeperPublicKeys` reduces to a plain object, and the shape guard above guarantees a positive integer, so `in` cannot pick up an inherited `Object.prototype` member.

sdk/javascript/packages/core/src/keeper.ts
```diff
@@ -811,12 +816,23 @@ const postQuery = async (options: SecretManagerOptions, path: string, payload: A
                 let errorObj: KeeperApiError | null = null
                 try { errorObj = JSON.parse(errorMessage) } catch {}
                 if (errorObj?.error === 'key') {
+                    const suggestedKeyId = errorObj.key_id
+                    if (typeof suggestedKeyId !== 'number' || !Number.isInteger(suggestedKeyId) || suggestedKeyId <= 0) {
```
mgallego-keeper: Non-blocking leftover from the round-2 placement thread, and the reason the PR body needs a small edit.

The membership check moved below the `customKey` branch, which is correct. This *shape* guard is still above it, so a pinned IL5 deployment whose server answers with a missing or malformed `key_id` gets the generic error instead of the actionable one. Verified on the rebuilt bundle, pinned config with `serverPublicKeyId: '20'`:

| server sends | resulting message |
|---|---|
| `key_id: 7` | `Server rejected the custom server public key (id 20)... Please update your IL5 KSM configuration.` |
| `key_id: 99` | same IL5 diagnostic (correct: membership does not intercept the pinned path) |
| no `key_id` | `Server key error response contains invalid key_id: undefined` |
| `key_id: 'abc'` | `Server key error response contains invalid key_id: "abc"` |

Storage is preserved in all four, so nothing is harmed here; it is purely which diagnostic support ends up chasing. Moving this guard below the `customKey` throw as well would make the pinned path genuinely untouched.

That is also the body edit: **"The already-bounded custom-key-pinned path (`stale pinned server key` diagnostic) is unchanged"** does not hold for the bottom two rows. While you are in there, the Changes section is missing the shape guard, the typed `KeeperError`s, and the membership check, and that last one is a genuine behavior change (an unsupported suggestion is now a hard typed failure rather than a silent persist) that belongs in front of whoever reads this at release time.

mgallego-keeper (changes requested):
## Round 3: every blocking ask is fixed, two release-hygiene items left

c3519324 takes all three round-2 asks. I verified each by mutation against a rebuilt bundle rather than by reading the diff, since this suite runs against `dist` and a stale bundle will happily lie to you. 57/57 green in a clean worktree.

**Verified fixed**

1. **Membership validation** (`keeper.ts:828`), the blocker carried from round 1. `key_id: 99` now makes exactly 1 request, throws a typed `KeeperError`, and never touches storage; `6` and `1e21` are rejected; `18` is still accepted and adopted; and a pinned IL5 config with an out-of-table id still gets the IL5 diagnostic. That last point is what the placement below the `customKey` branch buys, and it was the main risk in my own ask. Evidence in the thread.
2. **Runaway guard restored** (`test:333`). With `MAX_KEY_ROTATION_RETRIES` patched to `Infinity`, the bounded test now fails in **45ms** with `runaway loop detected in key rotation retry`. In round 2 that same mutation produced no output, pinned a core, never tripped `--testTimeout`, and had to be killed.
3. **Adoption asserted** (`test:361`, `test:369`). Deleting the `saveString` adoption line now fails with `Expected: 8, Received: 7`. In round 2 it left all 57 tests green.

The exhaustion message also now names an id that was actually sent, which closes the round-1 wording thread unprompted.

**Requesting changes for two things, both small and neither about code behavior**

1. **The membership check has no test.** Deleting those three lines leaves `keeper.test.ts` at 18/18 green (mutation-verified). The fix for the blocker is currently held in place by nothing. Inline comment has a ready-to-paste test that I confirmed passes on this head and fails with the check removed.
2. **No changelog entry.** #1077 merged since your last push and created a `## 17.6.0` section on the base branch, so there is now a place for this and no conflict risk (this PR does not touch `CHANGELOG.md`, so it will merge clean and ship undocumented). Suggested line:

```markdown
- KSM-1128 - Bounded the server key-rotation retry in `postQuery`: the `{"error":"key"}` branch now retries at most 3 times before throwing a typed `KeeperError` instead of retrying forever, and the server's suggested `key_id` is validated (shape, and membership in the bundled key table) before it is persisted, so an unsupported suggestion can no longer poison the stored configuration. The pinned custom-key path is unaffected.
```

**Follow-ups filed** (all Bug, unassigned, Triage, linked A-Relates to KSM-1128), so none of the below needs to happen in this PR:

- **KSM-1255**: `options.serverPublicKeyId` is re-saved on every call, so a completed rotation never sticks (permanent 2x request amplification); it is also the still-open route to persisting an unvalidated id, bypassing the check this PR adds. Closes the two-round-old `:791` thread.
- **KSM-1256**: a stored id outside the bundled table is unrecoverable (zero requests, no self-heal, survives restarts under `localConfigStorage`). This PR stops the SDK from *creating* that state; recovering configs already in it is separate. Cross-linked to KSM-1247/1248/1249, since the same unvalidated-persist defect shipped in those SDKs, and 17.5.0 and earlier could create it here.
- **KSM-1257**: `generateTransmissionKey:757` still throws an untyped `Error` for the same condition, now with a divergent message. KSM-1024 candidate.

**Still open and purely cosmetic:** the `:822` thread (pinned diagnostic prints `(id undefined)`).

Nice work on the rework: the adoption assertion and the runaway guard are the kind of thing that only pays off years later, when someone refactors this branch and the suite tells them instead of hanging.

sdk/javascript/packages/core/test/keeper.test.ts
mgallego-keeper: Taken verbatim in bb3c6205, and verified at this head rather than assumed: 58/58 green, and removing the three lines at `src/keeper.ts:828-830` now fails exactly this test, with `Received message: "Key number 99 is not supported"`, which is the config-poisoning path itself. So the test documents the bug it prevents.

That closes the blocker that ran through rounds 1, 2 and 3, and closes it properly: the check is now held in place by the suite rather than by anyone remembering it is load-bearing. Resolving.

sdk/javascript/packages/core/src/keeper.ts
mgallego-keeper: bb3c6205 did not pick this up, which was fine while it was non-blocking. It is no longer purely cosmetic though: the claim it concerns has now been copied into the changelog, so I have raised it there as blocking and am leaving this thread open as the code half of the same point.

Since I was asserting a behavior change, I measured it instead of reasoning about it. Same probe run against the base branch and against this head, pinned config with `serverPublicKeyId: '20'`:

| server sends | base (pre-PR) | bb3c6205 |
|---|---|---|
| `key_id: 7` | IL5 diagnostic | IL5 diagnostic (unchanged) |
| `key_id: 99` | IL5 diagnostic | IL5 diagnostic (unchanged) |
| no `key_id` | IL5 diagnostic | `Server key error response contains invalid key_id: undefined` |
| `key_id: 'abc'` | IL5 diagnostic | `Server key error response contains invalid key_id: "abc"` |

The configured key survives in every case, so nothing is broken; it is only which diagnostic support ends up chasing. Moving this guard below the `customKey` throw restores the old message for the bottom two rows and makes the "pinned path unchanged" claim true in both the PR body and the changelog, which is why I would rather see it fixed here than have the docs reworded around it.

sdk/javascript/packages/core/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Change Log
 
+## 17.6.0
+- KSM-1128 - Bounded the server key-rotation retry in `postQuery`. When the server sends `{"error":"key"}`, the code retries at most 3 times before throwing a typed `KeeperError`, instead of retrying forever. Before storing a suggested `key_id`, the code validates its shape (positive integer) and its membership in the bundled key table (keys 7-18). An unsupported key id can no longer corrupt the configuration. The pinned custom-key path does not change.
```
mgallego-keeper, line 9: **[blocking] Two things on this entry.**

**1. This heading conflicts with the base branch.** #1077 merged earlier today and created a `## 17.6.0` section already, so this adds a second one and the PR is now `CONFLICTING`. `CHANGELOG.md` is the only conflicted file; `keeper.ts` and `keeper.test.ts` auto-merge cleanly.

The trap in resolving it is that "take ours" drops #1077's line entirely, so here is the whole block, merged, with KSM-1035 first since it landed first:

```markdown
# Change Log

## 17.6.0
- KSM-1035 - Throttle backoff hardening: retry jitter is now one-sided (0 to +25%, so delays never fall below the computed floor), and a server-supplied `retry_after` is capped at 176s (the exponential ladder's last step) so it can no longer force an excessive or unbounded wait.
- KSM-1128 - Bounded the server key-rotation retry in `postQuery`. When the server sends `{"error":"key"}`, the code retries at most 3 times before throwing a typed `KeeperError`, instead of retrying forever. Before storing a suggested `key_id`, the code validates its shape (positive integer) and its membership in the bundled key table (keys 7-18). An unsupported key id can no longer corrupt the configuration.
```

(I left the final sentence off the KSM-1128 bullet there, pending your call on point 2.)

**2. The final sentence is not accurate.** "The pinned custom-key path does not change" is contradicted by measurement. Pinned config with `serverPublicKeyId: '20'`, base branch versus this head:

| server sends | base (pre-PR) | bb3c6205 |
|---|---|---|
| `key_id: 7` | IL5 diagnostic | IL5 diagnostic (unchanged) |
| `key_id: 99` | IL5 diagnostic | IL5 diagnostic (unchanged) |
| no `key_id` | IL5 diagnostic | `Server key error response contains invalid key_id: undefined` |
| `key_id: 'abc'` | IL5 diagnostic | `Server key error response contains invalid key_id: "abc"` |

The configured key is preserved in all four cases, so this is a diagnostic-quality change rather than a functional one. Two ways out, and I would take the first:

- **Make it true.** Move the shape guard at `src/keeper.ts:820` below the `customKey` branch, as suggested in the thread there. Four lines, restores the IL5 message for the bottom two rows, and then this sentence and the PR body bullet are both correct as written.
- **Reword it.** Drop the sentence, or replace it with something like: "The pinned custom-key path continues to preserve the configured key; a missing or malformed `key_id` now reports the validation error instead of the pinned-key diagnostic."

One smaller note while you are in here, non-blocking: "An unsupported key id can no longer corrupt the configuration" holds for the rotation path this PR fixes, but `options.serverPublicKeyId` still persists an unvalidated id (KSM-1255, verified on this head: zero network requests, untyped error). Scoping the sentence to the server-suggested id would make it exact.

mgallego-keeper (changes requested):
## Round 4: the test is exactly right, the changelog needs one more pass

The membership test landed verbatim, and I verified it at this head instead of taking the diff's word for it: 58/58 green, and deleting the three-line check now fails exactly that test with `Key number 99 is not supported`. That closes the blocker that ran through rounds 1, 2 and 3, and it closes it the right way, held by the suite rather than by memory.

Requesting changes for the changelog only, for two reasons, both on that single entry.

**1. It conflicts with the base branch.** This PR went from `MERGEABLE` to `CONFLICTING`. #1077 merged earlier today and already created a `## 17.6.0` section, so this adds a second one. `CHANGELOG.md` is the only conflicted file (`keeper.ts` and `keeper.test.ts` auto-merge clean), and the trap is that resolving it as "take ours" silently deletes #1077's KSM-1035 line. The exact merged block is in the inline comment so that cannot happen by accident.

**2. Its last sentence is not accurate,** and this is the third round that claim has appeared in. "The pinned custom-key path does not change" is contradicted by a base-versus-head comparison I ran for this round:

| server sends | base (pre-PR) | bb3c6205 |
|---|---|---|
| `key_id: 7` | IL5 diagnostic | IL5 diagnostic (unchanged) |
| `key_id: 99` | IL5 diagnostic | IL5 diagnostic (unchanged) |
| no `key_id` | IL5 diagnostic | `Server key error response contains invalid key_id: undefined` |
| `key_id: 'abc'` | IL5 diagnostic | `Server key error response contains invalid key_id: "abc"` |

The configured key is preserved in all four cases, so this is diagnostic quality rather than a functional problem. But it was in the PR body in round 2, still there in round 3 when I commented on it inline, and it has now been promoted into a user-facing changelog, which is the one place it will outlive this PR. Both fix options are inline; the code fix is four lines and makes the sentence true rather than working around it.

For completeness: the PR body is still unedited, so it carries the same sentence and still omits the shape guard, the typed `KeeperError`s, and the membership check. Not blocking, but it is what a release reader sees.

The three non-blocking round-3 items are untouched, which is fine, they are all tracked: KSM-1257 for the duplicate membership logic and the hardcoded range, plus the `:820` and `:822` threads.

Once the two `## 17.6.0` sections are merged into one and that sentence is either dropped or made true, this has my approval. The code has been correct since c3519324 and nothing this round changed that.

sdk/javascript/packages/core/src/keeper.ts
```diff
@@ -815,12 +820,24 @@ const postQuery = async (options: SecretManagerOptions, path: string, payload: A
...
                 if (errorObj?.error === 'key') {
+                    const suggestedKeyId = errorObj.key_id
                     const customKey = await options.storage.getString(KEY_SERVER_PUBLIC_KEY)
                     if (customKey) {
                         const currentKeyId = await options.storage.getString(KEY_SERVER_PUBLIC_KEY_ID)
-                        throw new Error(`Server rejected the custom server public key (id ${currentKeyId}). The server suggested key id ${errorObj.key_id}. Please update your IL5 KSM configuration.`)
+                        throw new KeeperError(`Server rejected the custom server public key (id ${currentKeyId ?? transmissionKey.publicKeyId}). The server suggested key id ${suggestedKeyId}. Please update your IL5 KSM configuration.`)
                     }
-                    await options.storage.saveString(KEY_SERVER_PUBLIC_KEY_ID, errorObj.key_id!.toString())
+                    if (typeof suggestedKeyId !== 'number' || !Number.isInteger(suggestedKeyId) || suggestedKeyId <= 0) {
```
mgallego-keeper, line 829: **The placement is right, and I checked it end to end rather than from the diff.** Moving this below the `customKey` branch restored the IL5 diagnostic for every malformed-`key_id` shape, byte-identical to base, so the pinned path is genuinely untouched now and the changelog sentence holds.

One note for whoever touches this next, no action now. This guard is behaviorally near-redundant with the membership check below it: everything it rejects, membership also rejects, with a single exception. A numeric string `"8"` passes membership, because object keys are strings, and is caught only here. Its other job is type narrowing, since it is what turns `number | undefined` into `number` and therefore what lets `suggestedKeyId in keeperPublicKeys` and `suggestedKeyId.toString()` compile.

That combination is worth knowing because it shapes how this fails if someone tidies it later: deleting the guard outright is a compile error, but reducing it to "the membership check already covers this" would quietly re-admit the string form, which base accepted and adopted. Rejecting the string is correct, and the cross-SDK evidence now settles the round-2 question I left open: `sdk/rust/src/core/core.rs:1194` says "Server returns key_id as a number, not a string", and Java's `val key_id: Int` and .NET's `public int key_id` cannot deserialize a string at all. Python's `str(key_id)` is the outlier, not the contract.

sdk/javascript/packages/core/test/keeper.test.ts
```diff
@@ -317,20 +321,96 @@ test('IL5 dynamic key - rotation suppression: server key_id hint ignored when se
...
+                throw new Error('runaway loop detected in key rotation retry')
+            }
+            return { statusCode: 400, data: enc.encode(keyErrorResponse(99)), headers: [] }
+        }
+    }
+    await expect(getSecrets(options)).rejects.toThrow(/unsupported key id 99/)
+    // Rejected before the retry loop persists anything: one request, config untouched.
+    expect(calls).toBe(1)
+    expect(await storage.getString('serverPublicKeyId')).toBeUndefined()
+})
```
mgallego-keeper, line 391: **Two regression tests, non-blocking.** Mutation battery run against this file at this head, each mutation rebuilt and re-run:

| mutation | caught? |
|---|---|
| drop the membership check | yes, fails with `Key number 99 is not supported` |
| `MAX_KEY_ROTATION_RETRIES = Infinity` | yes, runaway guard fires in ms |
| drop the `saveString` adoption | yes |
| bound off-by-one (`>=` to `>`) | yes, via `expect(calls).toBe(4)` |
| revert `87a945b1`, shape guard back above `customKey` | **no, 21/21 green** |
| delete the shape guard entirely | **no, 21/21 green** |

The four that are caught are exactly the ones we built tests for in rounds 2 through 4, and they hold up after the rename. The two that are missed are this round's commit: every pinned-key test here uses a valid `key_id: 7`, which passes the shape guard from either position, so nothing in the suite observes the placement.

Both tests below pass at this head and each fails under exactly its matching mutation and no other, verified rather than assumed. They need one change to the helper at the top of the file so the missing-`key_id` case is expressible:

```ts
const keyErrorResponse = (keyId?: unknown) => JSON.stringify({ error: 'key', key_id: keyId })
```

`JSON.stringify` drops `undefined` values, so `keyErrorResponse()` emits `{"error":"key"}` and the existing three call sites are unaffected.

```ts
test('key rotation - pinned custom key takes precedence over key_id validation', async () => {
    const fakeKey = 'BK9w6TZFxE6nFNbMfIpULCup2a8xc6w2tUTABjxny7yFmxW0dAEojwC6j6zb5nTlmb1dAx8nwo3qF7RPYGmloRM'
    const storage = inMemoryStorage({})
    await initializeStorage(storage, FAKE_ONE_TIME_TOKEN, 'fake.keepersecurity.com')
    await storage.saveString('serverPublicKey', fakeKey)
    await storage.saveString('serverPublicKeyId', '20')
    let calls = 0
    const enc = new TextEncoder()
    const options: SecretManagerOptions = {
        storage,
        queryFunction: async () => {
            calls++
            if (calls > 50) {
                throw new Error('runaway loop detected in key rotation retry')
            }
            // No key_id at all: the pinned-key diagnostic must still win over the shape guard.
            return { statusCode: 400, data: enc.encode(keyErrorResponse()), headers: [] }
        }
    }
    await expect(getSecrets(options)).rejects.toThrow(/Please update your IL5 KSM configuration/)
    expect(calls).toBe(1)
    expect(await storage.getString('serverPublicKeyId')).toBe('20')
})

test('key rotation - malformed key_id is rejected, not persisted', async () => {
    const storage = inMemoryStorage({})
    await initializeStorage(storage, FAKE_ONE_TIME_TOKEN, 'fake.keepersecurity.com')
    let calls = 0
    const enc = new TextEncoder()
    const options: SecretManagerOptions = {
        storage,
        queryFunction: async () => {
            calls++
            if (calls > 50) {
                throw new Error('runaway loop detected in key rotation retry')
            }
            return { statusCode: 400, data: enc.encode(keyErrorResponse('8')), headers: [] }
        }
    }
    await expect(getSecrets(options)).rejects.toThrow(/invalid key_id: "8"/)
    expect(calls).toBe(1)
    expect(await storage.getString('serverPublicKeyId')).toBeUndefined()
})
```

The first fails under the placement revert with the generic `invalid key_id: undefined` in place of the IL5 message; the second fails under the shape-guard deletion, since membership accepts the string `"8"` and persists it. Fine to punt both to KSM-1257 if you would rather not touch this file again.

mgallego-keeper (approved):
## Round 5: both blockers closed, approving

`87a945b1` took the code option rather than rewording, which is the outcome I hoped for, and the rebase resolved the changelog cleanly. I verified both at the head `21bc228e` by measurement rather than by reading the diff.

**Blocker 1, the changelog conflict.** Resolved by the rebase onto `8ccb7afd`, and resolved the right way: one `## 17.6.0` section, #1077's KSM-1035 line intact, KSM-1128 appended in order. The whole-file diff against base is a single added line, which is the check that would have caught a "take ours" resolution silently dropping KSM-1035.

**Blocker 2, the pinned-path claim.** Now true rather than reworded. Same probe as round 4, run against both bundles, pinned config with `serverPublicKeyId: '20'`:

| server sends | base `8ccb7afd` | head `21bc228e` |
|---|---|---|
| `key_id: 7` | IL5 diagnostic | byte-identical |
| `key_id: 99` | IL5 diagnostic | byte-identical |
| `key_id: 'abc'` | IL5 diagnostic | byte-identical |
| no `key_id` | IL5 diagnostic | byte-identical |

All four rows that diverged in round 4 now match base exactly. Two deltas remain on that path and both are things I asked for: the error class is `KeeperError` instead of plain `Error` (a subclass, so `catch (e: Error)` is unaffected), and a pin-without-id config now reports `(id 7)` instead of `(id undefined)`, which closes the last open cosmetic thread from round 1. The changelog sentence is accurate as written and I am done pushing on it.

The standard rotation path is unchanged from round 4 and still correct across 21 probed scenarios: `99`, `6`, `0`, `-1`, `8.5`, `'abc'`, `'8'`, `1e21`, `true`, `null` and a missing `key_id` all throw a typed `KeeperError` after exactly 1 request with storage untouched, where base persisted a poisoned id in seven of those cases. `7`, `8` and `18` bound at 4 requests where base ran to my 61-call runaway guard. `8200b424` also picked up the derived supported-range message, which closes the hardcoded `7-18` half of KSM-1257: `keeperPublicKeys` is built by a `reduce` from a contiguous array starting at 7, so `Object.keys` enumerates ascending and the message stays true by construction when key 19 is appended.

**One new gap, non-blocking, raised inline.** I ran a six-mutation battery against the suite this round. It holds four of the six things it should, and misses exactly the two that this round's commit represents: reverting `87a945b1` leaves 21/21 green, and so does deleting the shape guard. Every pinned-key test uses a valid `key_id: 7`, which passes the shape guard from either position, so nothing observes the placement that took three rounds to settle. Two verified tests are in the inline comment. I am explicitly not blocking on this: losing the membership check corrupted a config, which is why round 4 blocked, while losing this one degrades an error message. Different stakes, different call.

**Three notes that need nothing from you.**

The round-2 question about tolerating a quoted `"key_id": "8"` is now settled by evidence, and the strict reading here is the right one. `sdk/rust/src/core/core.rs:1194` carries the comment "Server returns key_id as a number, not a string", and Java declares `val key_id: Int` while .NET declares `public int key_id`, neither of which can deserialize a string at all.

The unbounded rotation `continue` this PR fixes exists verbatim in Java (`SecretsManager.kt:1832`), .NET (`SecretsManagerClient.cs:1577`) and Python (`core.py:661`). Java and .NET also lack the membership check, so the configuration-poisoning path closed here is still open in both. Worth parity tickets alongside KSM-1255 and KSM-1257 whenever there is room.

`test.js.yml` triggers only on `pull_request: branches: [master]`, so none of these tests run in CI on this PR. That is the repo-wide convention (python, java, rust and cli workflows are all master-only) and not something to change here; they will run on the release-to-master PR. Locally this head is 62/62 green across 6 suites.

Approving. The code has been correct since `c3519324` and this round made the documentation match it.

PR #1096, opened 2026-08-05
feat(python): add full PAM connection settings fields to CLI helper schema (KSM-1140)

stas-schaller:
## Summary

Expands the Python CLI helper's `v3/field_type.py` to carry the complete set of PAM connection settings fields, matching the Java SDK reference implementation and cross-checked against vault TypeScript type definitions.

### New Features
- KSM-1140: `PamSettings.connection.connection` expanded from 8 fields to the full `PamSettingsConnection` schema — RDP (30 fields), SSH/Terminal (11 fields), VNC (8 fields), database (4 fields), Telnet (4 fields), Kubernetes (7 fields), plus a nested `sftp` object (`SFTPConnection`: 6 fields)
- KSM-1140: `PamSettings.connection.portForward` gains `useSpecifiedLocalPort`, `localPort`, and `allowKeeperDBProxy` (vault TS source)
- KSM-1140: `allowSupplyHost` moved to the `PamSetting` wrapper level (matches vault TypeScript `PamSettingsFieldData`; Java duplicates it at both levels — vault TS is the authoritative source)
- KSM-1140: `PamRemoteBrowserSettings.connection` gains `recordingIncludeKeys`, `ignoreInitialSslCert`, audio control fields (`disableAudio`, `disableCopy`, `disablePaste`, `audioChannels`, `audioBps`, `audioSampleRate`), and browser session fields (`sessionPersistence`, `allowFileUploads`, `allowFileDownloads`)

## Testing

```bash
cd sdk/python/helper && python -m pytest -q
cd sdk/python/core  && python -m pytest -q
```

## Breaking Changes

None. All additions are new keys in existing schema dicts; existing field names and types are unchanged.

## Related Issues

- Closes KSM-1140
- Jira: KSM-1140, KSM-1138


sdk/python/helper/keeper_secrets_manager_helper/v3/field_type.py
```diff
@@ -876,20 +887,110 @@ class PamSettings(FieldType):
...
                             "reusePort": {"value_type": bool},
                             "port": {"value_type": str},
+                            "useSpecifiedLocalPort": {"value_type": bool},
+                            "localPort": {"value_type": str},
+                            "allowKeeperDBProxy": {"value_type": bool},
                         }
                     },
+                    "allowSupplyHost": {"value_type": bool},
                     "connection": {
                         "value_type": list,
```
mgallego-keeper, line 894: This is the big one: everything added under this node is dead metadata.

`value_type: list` is a terminal leaf for all three consumers of these schema dicts, and none of them look at the nested `schema`:

- `build_value` (`field_type.py:325`): `issubclass(list, FieldType)` and `issubclass(list, BaseEnum)` are False and `list is dict` is False, so it falls through to the simple data type `else` at `:365` and returns the value unrecursed and unfiltered.
- `is_value_valid` (`field_type.py:281`): no branch matches `list`, so zero validation.
- `_expand_value_type` (`field_type.py:50`, and the near duplicate at `record_type.py:33`): `issubclass(list, dict)` is False, so it emits a flat placeholder and never walks in.

Running the PR schema:

```
ksm secret template field pamSettings
{"connection": {"portForward":     "<#ADD: Port Settings>",
                "allowSupplyHost": "<#ADD: Insert a bool>",
                "connection":      "<#ADD: Connection Settings>"}}
```

and user input is unfiltered both before and after the PR:

```python
PamSettings(value={"connection": {"connection": [{"protocol": "rdp", "BOGUS": "yes"}]}}).to_dict()
# value[0].connection.connection[0] == {"protocol": "rdp", "BOGUS": "yes"}
```

So `hostKey`, `width`, `disableCsvExport`, `sftp{}`, `rbiSettings{}`, `useSpecifiedLocalPort`, `allowKeeperDBProxy` and the rest change nothing: they were already accepted before this PR, and junk is still accepted after. The same applies to the `database` and `dbConnectionMethod` pair from KSM-1127, which means that changelog line ("so template generation and record construction can use them") is already inaccurate.

Two ways to go, and I do not have a strong preference:

(a) Make it real: have `build_value`, `is_value_valid` and `_expand_value_type` recurse per item when `value_type is list` and a `schema` is present. That delivers what the PR describes and retroactively makes KSM-1127 true, but it switches key dropping on for `pamSettings`, so it is a behavior change that wants tests and might be happier outside a release branch.

(b) Treat these as documentation and reword the changelog so it claims only what `allowSupplyHost` and the `pamRemoteBrowserSettings` fields actually deliver.

Your call which fits the release better.

sdk/python/helper/keeper_secrets_manager_helper/v3/field_type.py
```diff
@@ -876,20 +887,110 @@ class PamSettings(FieldType):
                         "schema": {
                             "reusePort": {"value_type": bool},
                             "port": {"value_type": str},
+                            "useSpecifiedLocalPort": {"value_type": bool},
+                            "localPort": {"value_type": str},
+                            "allowKeeperDBProxy": {"value_type": bool},
                         }
                     },
+                    "allowSupplyHost": {"value_type": bool},
```
mgallego-keeper: Two things here.

The description says this moves `allowSupplyHost` to "the `PamSetting` wrapper level", but placed here it serializes to `value[0].connection.allowSupplyHost`, one level deeper than that. Worth comparing the four implementations of `pamSettings.value[i]`:

- Java `PamSetting` (`RecordData.kt:769`): `{connection, portForward, allowSupplyHost}`
- TS (`sdk/javascript/packages/core/src/keeper.ts:1667`), .NET (`RecordData.cs:607`), Rust (`field_structs.rs:2311`): `{portForward: [...], connection: [...]}`
- this helper: `{connection: {portForward: [...], allowSupplyHost, connection: [...]}}`

The helper has an extra `connection` wrapper that none of the other four have, so Java would try to parse `value[0].connection` as a `PamSettingsConnection` and choke on its `portForward` and `connection` keys. That wrapper predates this PR, but since the premise here is parity with Java and vault TS, it seems worth confirming what the vault actually accepts before adding 71 more fields underneath it.

Minor: the description says "moved", but the diff only adds; nothing was removed. Java also keeps `allowSupplyHost` on `PamSettingsConnection` itself, which you deliberately omit. The reasoning in the body makes sense, I am just noting it is a conscious divergence from the cited reference.

This one is live, incidentally, since it sits under a `dict` node, so it does reach template output.

sdk/python/helper/keeper_secrets_manager_helper/v3/field_type.py
```diff
@@ -850,11 +850,22 @@ class PamRemoteBrowserSettings(FieldType):
...
+                    "ignoreInitialSslCert": {"value_type": bool},
+                    "disableAudio": {"value_type": bool},
+                    "disableCopy": {"value_type": bool},
+                    "disablePaste": {"value_type": bool},
+                    "audioChannels": {"value_type": int},
+                    "audioBps": {"value_type": int},
+                    "audioSampleRate": {"value_type": int},
+                    "sessionPersistence": {"value_type": str},
+                    "allowFileUploads": {"value_type": bool},
+                    "allowFileDownloads": {"value_type": bool},
```
mgallego-keeper, line 868: These three are the highest risk additions in the PR, because they are the only ones that are both unverifiable and functionally live.

Java's `PamRbiConnection` (`RecordData.kt:614`) has 15 fields and this PR matches all 15 exactly, but these three are not among them, and they do not appear anywhere else in this repo or in the record-templates checkout. Everything under `pamSettings.connection` is inert, so a wrong name there is harmless, but this block sits under a `dict` node and does get emitted, so a wrong name here ships into user records.

Where do these come from, and should Java's `PamRbiConnection` be updated to match?

Worth saying explicitly that the rest of this block is the most valuable change in the PR. Before it, these keys were silently destroyed:

```python
# base
PamRemoteBrowserSettings(value={"connection": {"protocol": "http", "disableAudio": True, "allowFileUploads": True}}).to_dict()
# -> {"connection": {"protocol": "http"}}
# with this PR
# -> {"connection": {"protocol": "http", "disableAudio": true, "allowFileUploads": true}}
```

sdk/python/helper/keeper_secrets_manager_helper/v3/field_type.py
```diff
@@ -876,20 +887,110 @@ class PamSettings(FieldType):
...
+                            "clientCert": {"value_type": str},
+                            "clientKey": {"value_type": str},
+                            "useSSL": {"value_type": bool},
+                            "driveRedirectionPath": {"value_type": str},
+                            "rbiSettings": {
+                                "value_type": dict,
+                                "schema": {
+                                    "sessionPersistence": {"value_type": str},
+                                }
+                            },
```
mgallego-keeper: Neither of these is mentioned in the PR description, and I cannot find a source for either: not in Java's 78 field `PamSettingsConnection`, not in the JS, .NET or Rust SDKs, and zero hits in the record-templates checkout.

`rbiSettings` is the one I would push back on hardest: an RBI settings object nested inside a PAM *connection*, containing only `sessionPersistence`, which is also being added as a flat field on `PamRemoteBrowserSettings.connection` in this same PR. If both are real, the duplication deserves a comment explaining which one the vault reads.

Both are inert per the other comment, so there is no runtime risk today. This is about not baking unsourced names into the schema.

sdk/python/helper/keeper_secrets_manager_helper/v3/field_type.py
```diff
@@ -876,20 +887,110 @@ class PamSettings(FieldType):
                         "schema": {
                             "reusePort": {"value_type": bool},
                             "port": {"value_type": str},
+                            "useSpecifiedLocalPort": {"value_type": bool},
+                            "localPort": {"value_type": str},
+                            "allowKeeperDBProxy": {"value_type": bool},
```
mgallego-keeper: Java's `PamSettingsPortForward` (`RecordData.kt:762`) has exactly four fields: `reusePort`, `port`, `useSpecifiedLocalPort` and `localPort`. The first three you add match it; `allowKeeperDBProxy` is not there and does not appear anywhere else in the repo. The description attributes it to vault TS, which I cannot verify from here. Can you point at the source?

sdk/python/core/README.md
```diff
@@ -34,6 +34,7 @@ see the official docs link above.
 ## Change Log
 
 ### 17.4.0
+* KSM-1140 - Added full PAM connection settings field coverage to the CLI helper schema. `PamSettings.connection` now includes all `PamSettingsConnection` fields (RDP, SSH/Terminal, VNC, database, Telnet, Kubernetes, and SFTP sub-settings), `PamSettingsPortForward` gains `useSpecifiedLocalPort`, `localPort`, and `allowKeeperDBProxy`, and `PamRemoteBrowserSettings` gains audio control and browser session fields (`disableAudio`, `disableCopy`, `disablePaste`, `audioChannels`, `audioBps`, `audioSampleRate`, `sessionPersistence`, `allowFileUploads`, `allowFileDownloads`).
```
mgallego-keeper: This belongs in `sdk/python/helper/README.md` instead. The change is entirely in the helper package, and every other entry under `### 17.4.0` here is a `keeper_secrets_manager_core` change. The only "helper" mention anywhere in this file is KSM-458, which was itself a core change (removing core's dependency on the helper).

The helper's changelog is the `## Recent Changes` section of `sdk/python/helper/README.md`, and it already has a `### Version 1.1.3` heading on this branch with the two most recent helper changes under it:

```markdown
### Version 1.1.3
- KSM-1119 - Fixed `FieldType.__init__` crashing with `IndexError` ...
- KSM-1127 - Fixed `PamSettings.connection` schema missing `database` and `dbConnectionMethod` ...
```

`5da81237` (KSM-1127) is the direct precedent: same file, same class, six days ago.

While you are in there: `sdk/python/helper/setup.py` is still at `version="1.1.2"` and 1.1.2 is already on PyPI, so `validate-versions` in `publish.pypi.sdk.yml` will abort the 17.4.0 publish. Bumping it to 1.1.3 alongside this would unblock that.

mgallego-keeper (changes requested):
Thorough pass on this one. The field level work is solid: I AST diffed the new schema against Java's `PamSettingsConnection` (`sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/RecordData.kt:663`) and all 76 shared fields match on both name and type, including the ones that are easy to get wrong (`scrollback`, `serverAliveInterval`, `width`/`height`/`dpi`/`colorDepth` as int while `fontSize`/`port`/`destPort` stay str). `SFTPConnection`'s six fields are exact and there are no duplicate keys. Nice.

Three things I think need to change before this merges, plus a few questions. Details inline; the big one is that most of the new keys are never read at runtime.

**1. The changelog bullet is in the wrong file.**

The code change is entirely in `sdk/python/helper`, but the bullet went into `sdk/python/core/README.md` under `### 17.4.0`, where the other ten entries are all `keeper_secrets_manager_core` changes. The helper's changelog is the `## Recent Changes` section of `sdk/python/helper/README.md`, which on this branch already has a `### Version 1.1.3` heading holding KSM-1119 and KSM-1127. `5da81237` (KSM-1127) is the closest precedent: same file, same class, six days ago, logged there. More in the inline comment.

**2. `sdk/python/helper/setup.py` still says `version="1.1.2"`, and 1.1.2 is already on PyPI.**

The helper README documents 1.1.3 but `setup.py` was never bumped. `publish.pypi.sdk.yml` has a `validate-versions` job that hard aborts the whole Python publish when the helper version already exists on PyPI, so the 17.4.0 release will fail as things stand. This predates your PR (it arrived with KSM-1119 and KSM-1127), but since this adds a third 1.1.3 eligible change it looks like the right moment to bump it.

**3. Most of the new keys are inert.**

See the inline comment on the `connection` list node. Short version: `value_type: list` is a terminal leaf for every consumer of these schema dicts, so the roughly 71 keys added under `portForward` and the inner `connection` are never validated, never filtered, and never reach template output. `allowSupplyHost` and the `pamRemoteBrowserSettings` block *are* live, and the latter fixes real silent data loss, so there is genuine value here; it is just concentrated in about 12 of the 82 new keys.

**Tests and CI**

There is no coverage for `pamSettings` or `pamRemoteBrowserSettings` today (neither class is even imported in `tests/v3/v3_field_type_all_fields_test.py`), and `.github/workflows/test.python.helper.yml` triggers on `branches: [ master ]` only, so Test-Python-Helper will not run on a PR into a release branch. A test asserting the new `pamRemoteBrowserSettings` keys survive `to_dict()` would cover the part that actually changes behavior. Widening that workflow's `branches:` to include `release/sdk/python/**` might be worth its own ticket.

**Smaller notes**

- Java groups these fields with `// RDP specific settings` style comment headers. 79 keys in one flat block will be painful to re-diff against Java next time; copying the headers over would help a lot.
- The 11 live `pamRemoteBrowserSettings` keys render as `<#ADD: Insert a bool>` in `ksm secret template field` because they have no `desc`. Consistent with their neighbours, but since these are the ones users will actually see, a short `desc` on each would be a real win.
- `resizeMethod` and `colorScheme` got reordered, which adds churn to the diff without changing anything.
- Description counts are slightly off: Java's VNC group is 7 fields, not 8, and "database (4 fields)" counts `database` and `dbConnectionMethod`, which already shipped in KSM-1127, so this PR adds 2 there.
- Adjacent gaps if full PAM coverage is the goal, both of which would take effect immediately since they sit under dict nodes: `PamHostname` is missing `allowSupplyUser` and `allowSupplyHost` (Java's `Host` at `RecordData.kt:369` has both, and `pamHostname` is used by the built in `pamDatabase`, `pamDirectory` and `pamMachine` record types, so those two keys are being dropped from user input today), and `PamResources` is missing `adminCredentialRef` (Java's `PamResource` at `RecordData.kt:426`).

sdk/python/helper/keeper_secrets_manager_helper/v3/field_type.py
mgallego-keeper: You took (b) plus a strip, and the changelog now claims only what it delivers, so the ask here is settled. Two notes for whoever picks up the rest.

The root cause is more specific than "list nodes are terminal leaves". `value_type: list` is itself the wrong model. The vault has both of these as single objects, not arrays:

```ts
// vault/js/lib/record-types/field-data-pam-settings.ts:293
export type PamSettingsFieldData = {
  allowSupplyHost?: boolean
  connection?: PamConnectionSettings      // one object
  portForward?: PamPortForwardSettings    // one object
}
```

and it reads straight through them: `pamSettings?.connection?.port` (`record-types.ts:385`), `settings.connection?.protocol` (`api-dag-pam-link.ts:234`). Java agrees: `PamSetting = {connection: PamSettingsConnection?, portForward: PamSettingsPortForward?, allowSupplyHost: Boolean?}` (`RecordData.kt:773`). Only the JS, .NET and Rust SDKs type them as arrays (`keeper.ts:1966`, `RecordData.cs:607`, `field_structs.rs:2310`), and this schema follows those three.

So the fix for KSM-1140's remaining fields is not "teach the helper to recurse into list schemas", which was my option (a) last round. It is "change these two nodes to `value_type: dict`". Then the nested schemas are live with no engine change at all, filtering and validation come for free, and `allowSupplyHost` lands where the vault reads it. It is a real behavior change though (the emitted shape flips from `connection: [...]` to `connection: {...}`, and unknown keys start being dropped), so it wants its own ticket and its own tests rather than a release branch.

Second note: the strip also took `database` and `dbConnectionMethod`, which came from KSM-1127. Details in the other comment.

sdk/python/helper/keeper_secrets_manager_helper/v3/field_type.py
mgallego-keeper: Following up on the placement question, this time with an answer rather than a question. The vault TypeScript is unambiguous:

```ts
// vault/js/lib/record-types/field-data-pam-settings.ts:293
export type PamSettingsFieldData = {
  allowSupplyHost?: boolean
  connection?: PamConnectionSettings
  portForward?: PamPortForwardSettings
}
```

and `field-data.ts:45` maps `'pamSettings' -> PamSettingsFieldData`, so that is the type of `value[i]`. The two read sites confirm the path:

```ts
// record-types.ts:377
const pamSettings = getRecordField(record, 'pamSettings')?.value?.[0]
const allowSupplyHost = !!pamSettings?.allowSupplyHost

// api-dag-pam-link.ts:227,233
settings: PamSettingsFieldData
const allowSupplyHost = settings.allowSupplyHost
```

The vault reads `value[0].allowSupplyHost`. This schema emits it one level deeper:

```python
PamSettings(value={"connection": {"allowSupplyHost": True, "connection": [{"protocol": "rdp"}]}}).to_dict()
# {'type': 'pamSettings', 'value': [{'connection': {'allowSupplyHost': True, 'connection': [{'protocol': 'rdp'}]}}]}
```

and no connection-settings type in the vault has an `allowSupplyHost` member: `BaseConnectionSettings` (:93) is `{port, allowSupplyUser, recordingIncludeKeys}`, and `ConnectionSettingsHTTP` re-declares `allowSupplyUser` at :158 but never `allowSupplyHost`. Java's `PamSettingsConnection.allowSupplyHost` (`RecordData.kt:672`) is exactly the duplicate your description called out as the one to ignore, and you were right: it has no vault counterpart.

Net effect as written: the only live key this PR adds to `pamSettings` lands where nothing reads it. Your description says "moved to the `PamSetting` wrapper level (matches vault TypeScript `PamSettingsFieldData`)", which is the correct target. The code is one level short of it.

Fix is to move the key out of the inner `connection` schema and up to the field's top-level schema, as a sibling of `connection`:

```python
class PamSettings(FieldType):
    name = "pamSettings"
    schema = {
        "value_type": dict,
        "schema": {
            "allowSupplyHost": {"value_type": bool},
            "connection": {
                "value_type": dict,
                "desc": "PAM Settings",
                "schema": {
                    "portForward": {...},
                    "connection": {...},
                }
            },
        }
    }
```

I patched that in locally and ran it:

```python
PamSettings(value={"allowSupplyHost": True, "connection": {"connection": [{"protocol": "rdp"}]}}).to_dict()
# {'type': 'pamSettings', 'value': [{'allowSupplyHost': True, 'connection': {'connection': [{'protocol': 'rdp'}]}}]}

# ksm secret template field pamSettings
# {"allowSupplyHost": "<#ADD: Insert a bool>",
#  "connection": {"connection": "<#ADD: Connection Settings>", "portForward": "<#ADD: Port Settings>"}}
```

Both paths pick it up: the top-level dict branch of `build_value`, and `FieldType.__init__`'s kwargs branch, so `PamSettings(allowSupplyHost=False)` works too. Worth noting the current path is unreachable from the other direction as well: pass `allowSupplyHost` as a sibling of `connection` today and it is silently dropped, because the top-level schema has only the one key.

sdk/python/helper/keeper_secrets_manager_helper/v3/field_type.py
mgallego-keeper: Answering my own question: all three are real, and all three are named correctly. Sorry for the noise last round.

```ts
// vault/js/lib/record-types/field-data-pam-settings.ts:150,156
export enum SessionPersistence { NONE = 'none', USER = 'user', RESOURCE = 'resource' }

export type ConnectionSettingsHTTP = BaseConnectionSettings & ClipboardConnectionSettings & {
  protocol: ConnectionProtocol.HTTP
  allowSupplyUser?: boolean
  allowUrlManipulation?: boolean
  allowedUrlPatterns?: string
  allowedResourceUrlPatterns?: string
  httpCredentialsUid?: string
  autofillConfiguration?: string
  ignoreInitialSslCert?: boolean
  sessionPersistence?: SessionPersistence
  disableAudio?: boolean
  audioChannels?: 1 | 2
  audioBps?: 8 | 16
  audioSampleRate?: number
  allowFileUploads?: boolean
  allowFileDownloads?: boolean
}
```

`field-data.ts:43` maps `'pamRemoteBrowserSettings' -> PamRemoteBrowserSettingsFieldData = {connection?: ConnectionSettingsHTTP}`, which is the exact shape this class already has. `disableCopy` and `disablePaste` come from `ClipboardConnectionSettings` (:37), and HTTP is in the `clipboardConnectionSettings` switch (:54). `recordingIncludeKeys` comes from `BaseConnectionSettings` (:96). So all 18 keys in this block check out, types included: `int` is right for the three audio fields, `str` for `sessionPersistence`.

Independent corroboration for the file-transfer pair: QATA-3687 automates "Allow File Uploads" as a real RBI connection setting (`allowFileUploads: true`), and PG-418 documents the Gateway forwarding the per-record `rbiSettings.sessionPersistence` into KeeperDB's `CredentialPayload.session_persistence`.

Three optional things while you are in here. All of them sit under this dict node, so unlike the `pamSettings` side they would take effect immediately.

1. `allowSupplyUser` is missing. `ConnectionSettingsHTTP` re-declares it explicitly at :158 and the launch UI branches on it (it is what selects the `credentialSelect` affordance), so RBI records carry it today and this schema drops it on input.

2. `acceptLanguage` and `suppressJsDialogs` are missing. They are not in my checkout of the vault, which is from April, but the current model has them: the [Web Vault PAM Sessions over the Guacamole/WebRTC Channel implementation analysis](https://keeper.atlassian.net/wiki/spaces/IOS/pages/5202280454/Web+Vault+PAM+Sessions+over+the+Guacamole+WebRTC+Channel+Implementation+Analysis) (analyzed 2026-08-10 against `~/vault`) lists `ConnectionSettingsHTTP` as ending in "`allowFileUploads`/`allowFileDownloads`, `acceptLanguage`, `suppressJsDialogs`". Worth a look at `~/vault` HEAD before adding them.

3. `sessionPersistence` is a closed set of three values. The helper has enum support (`BaseEnum` in `v3/enum.py`, used by `AccountTypeEnum`, `CountryEnum`, `PhoneTypeEnum`), and `_expand_value_type` renders an enum via `build_example()`, so the template would read `<#ADD: none | resource | user>` instead of `<#ADD: Insert a str>`. Since this is the only new field with a fixed value set, that is a cheap win for `ksm secret template field pamRemoteBrowserSettings`. Note it also makes `is_value_valid` reject typos for that key, case-insensitively, which is a small behavior change worth calling out if you take it.

sdk/python/helper/keeper_secrets_manager_helper/v3/field_type.py
mgallego-keeper: Correction: `rbiSettings` is real, and your description was right to cite vault TS. I found the source:

```ts
// vault/js/lib/record-types/field-data-pam-settings.ts:235
export type KeeperDbRbiSettings = {
  sessionPersistence?: SessionPersistence
}

export type KeeperDBSettings = {
  dbConnectionMethod?: DBConnectionMethod
  rbiSettings?: KeeperDbRbiSettings
}
```

reached from `DatabaseConnectionSettings` (:244), which is why it belongs on a database connection. PG-418 covers the runtime half: the Gateway is supposed to forward the per-record `rbiSettings.sessionPersistence` into KeeperDB's `CredentialPayload.session_persistence`, and today it sends none of `user_id`, `session_persistence` or `resource_id`, which is why KeeperDB notebooks do not survive cross-gateway routing in HA.

So my "cannot find a source" call was wrong, and the duplication I asked you to justify is not duplication: flat `sessionPersistence` on `ConnectionSettingsHTTP` is the RBI record setting, `rbiSettings.sessionPersistence` is the database record setting for KeeperDB mode. Both correct, different records.

Both are stripped now along with the rest of the inert block. Worth re-adding under the `value_type: dict` ticket. Resolving.

sdk/python/helper/keeper_secrets_manager_helper/v3/field_type.py
mgallego-keeper: Correction: `allowKeeperDBProxy` is real, and vault TS has it exactly where your description said:

```ts
// vault/js/lib/record-types/field-data-pam-settings.ts:285
export type PamPortForwardSettings = {
  port?: string
  reusePort?: boolean
  useSpecifiedLocalPort?: boolean
  localPort?: string
  allowKeeperDBProxy?: boolean
}
```

The Confluence analysis adds the semantics: a separate `protocolSupportsKeeperDbProxy` predicate (MySQL, MariaDB, PostgreSQL, SQL Server, Oracle) governs `PamPortForwardSettings.allowKeeperDBProxy`, and it applies to the port-forward path rather than to `dbConnectionMethod`. You pointed at vault TS, vault TS has it, and my inability to verify it was on my side.

Worth noting for the tracking ticket that Java is the one behind here: `PamSettingsPortForward` is still 4 fields (`RecordData.kt:766`) with no `allowKeeperDBProxy`, and the vault also has `port` on this type, which neither Java nor this schema carries. Stripped now as inert. Resolving.

sdk/python/core/README.md
mgallego-keeper: Done. The entry is out of `sdk/python/core/README.md` (no `KSM-1140` left anywhere in that file) and into the helper's `### Version 1.1.3` alongside KSM-1119 and KSM-1127, and `setup.py` is at 1.1.3.

Verified the release side while I was here: 1.1.3 is not on PyPI (latest is 1.1.2, uploaded 2026-06-15), and `validate-versions` derives the version with `python3 setup.py --version`, so it picks up the bump and the 17.4.0 publish is unblocked. Resolving.

sdk/python/helper/keeper_secrets_manager_helper/v3/field_type.py
```diff
@@ -888,8 +900,6 @@ class PamSettings(FieldType):
                             "ignoreCert": {"value_type": bool},
                             "resizeMethod": {"value_type": str},
                             "colorScheme": {"value_type": str},
-                            "database": {"value_type": str},
-                            "dbConnectionMethod": {"value_type": str},
```
mgallego-keeper, line 892: These two lines revert 5da81237 (KSM-1127), your own fix from 2026-08-05, which is on the base branch. The README hunk in this same PR keeps its changelog entry, one line above the new KSM-1140 one:

> KSM-1127 - Fixed `PamSettings.connection` schema missing `database` and `dbConnectionMethod` fields. Both fields are now present in the `pamDatabase` connection schema, so template generation and record construction can use them.

After this commit they are not present, so 1.1.3 would ship a changelog entry that its own code contradicts. That is the blocking part: the file and the code have to agree.

On intent, the commit message says "strip inert sub-schemas from list-typed portForward and connection fields", but that is not what the edit does. `protocol`, `userRecords`, `security`, `ignoreCert`, `resizeMethod` and `colorScheme` all stay under the same list node, and `reusePort`/`port` stay under `portForward`. What the edit actually produces is the exact JS, .NET and Rust shape: `PamSettingsConnection` as those 6 fields (`keeper.ts:1952`, `RecordData.cs:591`, `field_structs.rs:2274`) and `PamSettingsPortForward` as `reusePort` plus `port`. Whether that alignment was deliberate or the two KSM-1127 lines came along for the ride, I cannot tell from here.

Either resolution works for me:

**(a) Restore the two lines.** Smallest diff, keeps the KSM-1127 entry honest, and both fields are genuinely real: Java has them on `PamSettingsConnection` (`RecordData.kt:685-686`) and so does the vault, as `DatabaseConnectionSettings.database` (`field-data-pam-settings.ts:246`) and `KeeperDBSettings.dbConnectionMethod` (:240).

**(b) Keep the removal** and rewrite the KSM-1127 line to say what it actually delivered, which is nothing observable, plus a note that the keys were dropped again.

For the record, the removal is behavior-neutral. I diffed `get_field_type_schema('pamSettings')` and `PamSettings(...).to_dict()` between the base branch and ef1fd54, and the only difference in either is `allowSupplyHost`:

```python
# both base and ef1fd54
# ksm secret template field pamSettings -> connection.connection == "<#ADD: Connection Settings>"
PamSettings(value={"connection": {"connection": [
    {"protocol": "mysql", "database": "db1", "dbConnectionMethod": "tls", "JUNK": "kept?"}]}}).to_dict()
# value[0].connection.connection[0] keeps all four keys, JUNK included, before and after
```

So this is a docs-consistency call, not a functional one. Which also means KSM-1127 itself never delivered what its ticket and changelog claim, and is worth reopening or closing as a no-op once the `value_type: dict` change gives it somewhere real to land.

mgallego-keeper (changes requested):
Re-reviewed at `ef1fd54`. All three asks from round 1 are done, and done correctly. Before the new findings, I owe you corrections: I got the field sourcing wrong last round, and your PR description was right.

I found the vault TypeScript locally (`vault/js/lib/record-types/field-data-pam-settings.ts` plus `field-data.ts`), and it confirms every name I questioned.

| Round-1 doubt | Verdict | Vault source |
| --- | --- | --- |
| `sessionPersistence`, `allowFileUploads`, `allowFileDownloads` | all real, correctly named | `ConnectionSettingsHTTP` :156-171. `sessionPersistence` is the `SessionPersistence` enum (`none`/`user`/`resource`), `audioChannels` is `1 \| 2`, `audioBps` is `8 \| 16` |
| `allowKeeperDBProxy` | real | `PamPortForwardSettings` :285-291 |
| `rbiSettings: {sessionPersistence}` | real, and not a duplicate | `KeeperDbRbiSettings` / `KeeperDBSettings` :235-242, reached via `DatabaseConnectionSettings` :244 |

Two corrections that cut the other way:

- I said Java "would choke" on the helper's extra keys. It will not. Record data is decoded with `nonStrictJson`, which sets `ignoreUnknownKeys = true` (`SecretsManager.kt:1698`, used for `KeeperRecordData` at `:1385`). Unknown keys are silently ignored, which is harder to debug than an exception but is not a crash.
- I said `pamHostname` is missing `allowSupplyUser` and `allowSupplyHost`. Drop that one: the vault's `HostFieldData` is `{hostName, port}` only (`field-data.ts:132`), and Java's `Host` is the outlier. In the vault, `allowSupplyUser` lives on the connection settings and `allowSupplyHost` on the field data, not on the host field. The `PamResources` note does stand: `PamResourcesFieldData` has `adminCredentialRef` (`field-data.ts:156`) with a dedicated UI component (`AdminCredentialField.tsx`), and this schema drops it.

So the names in this PR are sound. Two things still need to change, both one-liners, details inline.

**1. `allowSupplyHost` is one level too deep, so the only live `pamSettings` addition in this PR is inert.** The vault reads `value[0].allowSupplyHost` (`record-types.ts:378`, `api-dag-pam-link.ts:233`); this schema emits `value[0].connection.allowSupplyHost`, and no vault connection-settings type has that member. Your description names the right target (`PamSettingsFieldData`), the code just sits one level below it. Moving the key to the field's top-level schema fixes it; I patched and ran it to confirm.

**2. The commit removes `database` and `dbConnectionMethod`, reverting KSM-1127, while this PR's own README hunk keeps the KSM-1127 line saying both fields are present.** Behavior-neutral either way, but 1.1.3 cannot ship a changelog that its code contradicts. Restore the lines or retract the entry, either is fine.

### What is confirmed working

The `pamRemoteBrowserSettings` block is the real content of this PR and it delivers. Base versus `ef1fd54`:

```python
PamRemoteBrowserSettings(value={"connection": {
    "protocol": "http", "disableAudio": True, "disableCopy": False,
    "allowFileDownloads": False, "sessionPersistence": "persistent", "audioChannels": 2}}).to_dict()

# base (release/sdk/python/core/v17.4.0)
# value[0].connection == {"protocol": "http"}                       <- 5 keys silently destroyed
# ef1fd54
# value[0].connection == {"protocol": "http", "disableAudio": true, "disableCopy": false,
#                         "allowFileDownloads": false, "sessionPersistence": "persistent",
#                         "audioChannels": 2}
```

`False`, `0` and `""` survive as well, since `build_value` only drops `None`. And it is not just the Python API: the same 11 keys now survive `Parser` plus `Record` via `f.pamRemoteBrowserSettings={...}`, and they appear in both `ksm secret template field pamRemoteBrowserSettings` and the record template for any record type carrying the field. Junk keys are still filtered. That is the silent-data-loss fix, and it is why this should land.

Release plumbing also checks out: helper 1.1.3 is not on PyPI (latest 1.1.2, 2026-06-15), `validate-versions` reads `python3 setup.py --version` so it sees the bump, the changelog is in the right file, and every claim in the new changelog line matches a key the code actually emits. Helper suite: 63 passed. `git diff --check` clean.

### Not asking for these, but they are the interesting follow-ups

- **`portForward` and `connection` should be `value_type: dict`, not `list`.** This is the actual fix for KSM-1140's remaining fields, and it is a sourcing bug rather than an engine limitation: the vault and Java both model these as single objects (`PamSettingsFieldData`, `PamSetting`), and only the JS, .NET and Rust SDKs use arrays. Flip the two nodes to `dict` and the whole nested field set becomes live, validated and filtered with no change to `build_value`, `is_value_valid` or `_expand_value_type`. It is a behavior change to the emitted shape, so it wants its own ticket and tests, not a release branch. Worth deciding whether the three array-typed SDKs or the vault is wrong before anyone writes it.
- **KSM-1140 is not closed by this PR.** The ticket is "Python SDK: Add PAM connection settings fields (KSM-738 parity)" and its body is about the Python SDK having no typed PAM classes at all, 0 of 61 fields, with the CLI helper schema as an additional site. `sdk/python/core` still has no PAM types (`dtos.py:798` lists the field type names and nothing more), and this PR now adds 12 helper keys. Also `tracking/features/KSM-738-SDK-TRACKING.md`, cited by the ticket, is not in the repo on any branch. The PR description still describes the pre-strip change (RDP 30 fields, SSH/Terminal 11, VNC 8, nested `sftp`, the `portForward` trio) and says "Breaking Changes: None. All additions are new keys", which is no longer accurate now that the diff also removes two. Worth updating the description and splitting the ticket.
- **`pamRemoteBrowserSettings.connection` is still missing live fields:** `allowSupplyUser`, plus `acceptLanguage` and `suppressJsDialogs` from the current vault model. Details inline.
- **Still no test coverage for either class,** and `.github/workflows/test.python.helper.yml` triggers on `branches: [ master ]`, so `Test-Python-Helper` did not run on this PR: only the two Socket Security checks did. A single test asserting the 11 RBI keys survive `to_dict()` would pin exactly the regression class this PR fixes. Widening that workflow to `release/sdk/python/**` deserves its own ticket.
- **Unrelated pre-existing bug, worth its own ticket,** since it will bite anyone using the fields this PR adds: the documented sub-key syntax silently produces an empty field when the value is a JSON object. `Field.__init__` runs the value setter first, which JSON-decodes the string into a dict, and then skips the `{value_key: value}` wrapping because of the `isinstance(self.value, dict) is False` guard (`field.py:30`). So `c.pamRemoteBrowserSettings.connection={"disableAudio":true}` yields `{"type": "pamRemoteBrowserSettings", "value": []}` with no error, on both base and `ef1fd54`. The field-level form `c.pamRemoteBrowserSettings={"connection":{...}}` works fine.

Happy to file the tickets (`value_type: dict` change, KSM-1140 split, workflow branches, `Field` value-key bug) if useful.

sdk/python/helper/keeper_secrets_manager_helper/v3/field_type.py
mgallego-keeper: Fixed in `dfd4f720`, and this is the right fix: `allowSupplyHost` is now a sibling of `connection` in `PamSettings.schema["schema"]`, so it serializes to `value[0].allowSupplyHost`, which is the path vault TS reads.

Verified on `dfd4f720` by running it, not by reading the schema.

`to_dict()`:

```python
PamSettings(value={"allowSupplyHost": True,
                   "connection": {"portForward": [{"port": "22", "reusePort": True}],
                                  "connection": [{"protocol": "ssh"}]}}).to_dict()
```

```json
{"type": "pamSettings",
 "value": [{"allowSupplyHost": true,
            "connection": {"portForward": [{"port": "22", "reusePort": true}],
                           "connection": [{"protocol": "ssh"}]}}]}
```

The full `Parser` -> `Record` -> `get_record_create_obj()` path (what `ksm secret add` actually runs) agrees: the emitted `RecordField` has `value[0]["allowSupplyHost"] is True`.

Template generation puts it at the correct level too:

```
$ ksm secret template field pamSettings
{"type": "pamSettings",
 "value": {"allowSupplyHost": "<#ADD: Insert a bool>",
           "connection": {"portForward": "<#ADD: Port Settings>",
                          "connection": "<#ADD: Connection Settings>"}},
 "privacyScreen": false}
```

`allowSupplyHost: False` survives too, so there is no falsy-drop, and the 63 helper tests still pass.

One consequence worth knowing about, which is the reason for the single comment I left on the README: the old nested location is now dropped rather than rejected, and it takes the whole field with it. `PamSettings(value={"connection": {"allowSupplyHost": True}}).to_dict()` returns `"value": []`, an empty field, with no error.

The outer `connection` wrapper is unchanged, so `value[0].connection` is still `{portForward, connection}` rather than vault's `PamConnectionSettings`. That is the pre-existing shape problem and stays out of scope here, per the other thread. Resolving.

sdk/python/helper/keeper_secrets_manager_helper/v3/field_type.py
mgallego-keeper: Fixed in `dfd4f720`. `database` and `dbConnectionMethod` are back under the inner `connection.schema`, byte identical to `5da81237`:

```python
"colorScheme": {"value_type": str},
"database": {"value_type": str},
"dbConnectionMethod": {"value_type": str},
```

The cleanest confirmation is that they no longer appear in the PR diff at all. `git diff origin/release/sdk/python/core/v17.4.0...dfd4f720 -- sdk/python/helper/keeper_secrets_manager_helper/v3/field_type.py` is now exactly two hunks: the 11 new `PamRemoteBrowserSettings.connection` keys, and the single `allowSupplyHost` line. Net zero change on those two lines, so the accidental KSM-1127 revert is fully undone and the 1.1.3 changelog entry for it is accurate again.

Thanks for turning both of these around quickly. Resolving.

sdk/python/helper/README.md
```diff
@@ -9,6 +9,7 @@ For more information see our official documentation page https://docs.keeper.io/
 ### Version 1.1.3
 - KSM-1119 - Fixed `FieldType.__init__` crashing with `IndexError` when a complex field (address, name, host, paymentCard, etc.) is returned by the server with an empty value list. The SDK now leaves attribute variables as `None` instead of indexing into `[]`.
 - KSM-1127 - Fixed `PamSettings.connection` schema missing `database` and `dbConnectionMethod` fields. Both fields are now present in the `pamDatabase` connection schema, so template generation and record construction can use them.
+- KSM-1140 - Added `allowSupplyHost` to the `PamSettings.connection` schema. Added audio control and browser session fields to `PamRemoteBrowserSettings.connection`: `disableAudio`, `disableCopy`, `disablePaste`, `audioChannels`, `audioBps`, `audioSampleRate`, `sessionPersistence`, `allowFileUploads`, `allowFileDownloads`, `ignoreInitialSslCert`, `recordingIncludeKeys`.
```
mgallego-keeper: One phrase left over from `dfd4f720`. This line still says `allowSupplyHost` was added to the `PamSettings.connection` schema, but that commit moved it out of `connection` up to the field level, which was the entire point of the fix.

This matters more than a normal doc nit, because the old location is not rejected, it is silently swallowed, and it takes the whole field with it. A user following this line writes:

```
ksm secret add ... 'pamSettings={"connection": {"allowSupplyHost": true}}'
```

and gets:

```python
>>> PamSettings(value={"connection": {"allowSupplyHost": True}}).to_dict()
{"type": "pamSettings", "value": []}
```

An empty field, no error, no warning. So as written the changelog points users at the one path that loses the value, and this is the changelog that ships to PyPI with 1.1.3.

Suggested replacement for the first sentence only:

```suggestion
- KSM-1140 - Added `allowSupplyHost` to the `PamSettings` schema at the field level, so it serializes to `value[0].allowSupplyHost`. Added audio control and browser session fields to `PamRemoteBrowserSettings.connection`: `disableAudio`, `disableCopy`, `disablePaste`, `audioChannels`, `audioBps`, `audioSampleRate`, `sessionPersistence`, `allowFileUploads`, `allowFileDownloads`, `ignoreInitialSslCert`, `recordingIncludeKeys`.
```

The second sentence needed no change: I checked all 11 names against vault TS `field-data-pam-settings.ts` and every one matches, across `BaseConnectionSettings` (`recordingIncludeKeys`), `ClipboardConnectionSettings` (`disableCopy`, `disablePaste`) and `ConnectionSettingsHTTP` (the other eight). The list is also complete, matching the diff exactly with nothing over- or under-claimed.

mgallego-keeper (changes requested):
Round 3. Both blockers from round 2 are fixed in `dfd4f720`, and I verified both by running the code rather than reading the schema. One phrase in the changelog is left over from the fix itself, and it is the only thing standing between this and an approval.

### Both blockers, confirmed fixed

**`allowSupplyHost` depth.** Now a sibling of `connection` at the field level, so it emits `value[0].allowSupplyHost`, which is what vault TS reads. Confirmed through `to_dict()`, through template generation, and through the full `Parser` -> `Record` -> `get_record_create_obj()` path that `ksm secret add` actually runs. `False` survives as well, so no falsy-drop.

**KSM-1127 revert.** `database` and `dbConnectionMethod` are restored byte identical to `5da81237`. They no longer appear in the PR diff at all, so the revert is fully undone and that changelog line is true again.

### What the PR is now

The net diff against `release/sdk/python/core/v17.4.0` is 3 files, +15/-2, and every added key is live and name-verified against vault TS `field-data-pam-settings.ts`:

| Where | Keys | Live? |
|---|---|---|
| `PamRemoteBrowserSettings.connection` | `recordingIncludeKeys`, `ignoreInitialSslCert`, `disableAudio`, `disableCopy`, `disablePaste`, `audioChannels`, `audioBps`, `audioSampleRate`, `sessionPersistence`, `allowFileUploads`, `allowFileDownloads` | yes, `dict` node |
| `PamSettings` (field level) | `allowSupplyHost` | yes, `dict` node |

That is 12 keys, all of which survive serialization, down from the 82 claimed in round 1 of which roughly 71 were inert. This is a much better PR than the one I first reviewed. Helper suite: 63 passed.

### The one remaining item

The KSM-1140 changelog line still says `allowSupplyHost` was added to `PamSettings.connection`. After `dfd4f720` that is the wrong location, and it is the location where the value is silently discarded (`value` collapses to `[]`, no error). It is a one-phrase fix and I left a `suggestion` block on the line with exact replacement text.

I am requesting changes only to keep the changelog from shipping guidance that loses user data on PyPI. Push that one line and I will approve immediately; nothing else is outstanding on my side.

### Not blocking, for the tracking ticket

Unchanged from round 2, all still unfiled and none of them belong in this PR:

1. `portForward` and `connection` should be `value_type: dict`, not `list`. The vault and Java both model them as single objects. This is the real KSM-1140 fix and it is a breaking change, so it needs its own ticket.
2. The outer `connection` wrapper means `value[0].connection` is `{portForward, connection}` rather than vault's `PamConnectionSettings`. Same root cause as 1.
3. KSM-1140's actual scope is typed PAM classes in `sdk/python/core`, still 0 of 61 fields, so this PR does not close the ticket.
4. `Test-Python-Helper` is `branches: [ master ]`, so it did not run on this release-branch PR. I ran the suite locally instead.
5. `Field.__init__` dropping the `value_key` wrapper for a JSON-object sub-key (`field.py:30`) is what turns a wrong sub-key into a silent empty field, which is exactly the failure mode in the changelog comment above.

One small pre-existing thing I noticed and am deliberately not asking you to change: no built-in record type in `default_record_types.yml` carries `pamSettings` or `pamRemoteBrowserSettings`, so these schemas are only reachable as custom fields or via a server-supplied record type. Related, the KSM-1127 line's "present in the `pamDatabase` connection schema" is loose, since the `pamDatabase` record type has no `pamSettings` field. Both predate this PR.

sdk/python/helper/README.md
mgallego-keeper: Fixed in `73cbe807`. Verified by running the helper at this head, not by reading the schema.

Your wording came out better than the suggestion I supplied: "field level, alongside `connection`" tells the reader where the key goes without making them parse a JSON path. And it is accurate:

```python
>>> PamSettings(value={"allowSupplyHost": True}).to_dict()
{"type": "pamSettings", "value": [{"allowSupplyHost": true}]}

>>> PamSettings(value={"connection": {"allowSupplyHost": True}}).to_dict()
{"type": "pamSettings", "value": []}
```

`allowSupplyHost` sits at `field_type.py:880` as a sibling of `connection` at 881, so the location this line now documents is the one that serializes, and the location the old wording pointed at is still the one that silently empties the whole field. That was the entire concern here, and it is closed.

Second sentence re-checked at this head as well: all 11 names are present under `PamRemoteBrowserSettings.connection` (`field_type.py:853-868`) and they serialize where the line says they do:

```python
>>> PamRemoteBrowserSettings(value={"connection": {"disableAudio": True, "audioChannels": 2}}).to_dict()
{"type": "pamRemoteBrowserSettings", "value": [{"connection": {"disableAudio": true, "audioChannels": 2}}]}
```

Resolving.

mgallego-keeper (approved):
## Round 4: approved

`73cbe807` takes the one blocker left from round 3, and that was the only thing between this PR and an approval. Promise kept.

Verified at this head rather than from the diff:

- `allowSupplyHost` is at the field level (`field_type.py:880`, sibling of `connection` at 881) and serializes to `value[0].allowSupplyHost`. The location the previous wording pointed at still returns `{"value": []}`, silently, so this correction was worth making and it now reads correctly.
- All 11 `PamRemoteBrowserSettings.connection` fields named in the entry exist (`field_type.py:853-868`) and serialize under `connection`.
- `database` and `dbConnectionMethod` are still inside the inner `connection.schema` (`field_type.py:903-904`), so the KSM-1127 entry above this one also still holds.
- Helper suite: 63 passed, 0 failed at `73cbe807`.
- Clean merge against `release/sdk/python/core/v17.4.0`. The base's STE pass (`c10a925e`) touched `sdk/python/core/README.md` and the examples but not the helper README, so this line lands exactly as written.

All 8 review threads are resolved. Nothing further from me on the code.

PR #1097, opened 2026-08-07
fix(cli): KSM-1168 require --prefix for ksm sync --type aws; redact dry-run output

stas-schaller:
## Summary

CLI security fix for two issues identified in Bugcrowd submission 0a7417ea-c744-4934-b1cf-6af91ef4a39a (P3). Both issues affect `ksm sync --type aws` with `--record`, `--folder`, or `--folder-recursive`.

### Fixed

- **Breaking / Security**: KSM-1168 — `ksm sync --type aws` with `--record`, `--folder`, or `--folder-recursive` now requires `--prefix`. The prefix is prepended to every AWS secret name derived from a record title, confining the sync to an operator-controlled namespace. Without it, any Keeper user who can add records to a synced folder could set a record title to an existing AWS secret path and overwrite it on the next sync run.
- **Security**: KSM-1168 — `--dry-run` output no longer includes the live destination secret value. It now reports `dstExists` (bool) and `dstDiffers` (bool or null) only, eliminating the CI-log exfiltration path.

## Testing

```bash
cd integration/keeper_secrets_manager_cli && pytest tests/sync_test.py
```

Key new tests: `test_sync_prefix_prepended_to_record_title`, `test_sync_map_does_not_require_prefix`, `test_dry_run_does_not_disclose_dst_value`.

## Security Impact

Both changes reduce the attack surface for an insider threat scenario (attacker has Keeper account with folder write access and can observe sync output or trigger sync runs):

- `--prefix` required: limits the blast radius of title injection to the operator-defined namespace.
- `dstValue` removed from dry-run: live secret values no longer appear in CI build logs.

## Breaking Changes

`--prefix` is now required when using `--record`, `--folder`, or `--folder-recursive` with `--type aws`. Users upgrading from 1.4.x must add `--prefix <value>` to existing sync commands (e.g. `--prefix keeper/`). `--map` mode is unaffected.

## Related Issues

- Closes KSM-1168
- Bugcrowd: 0a7417ea-c744-4934-b1cf-6af91ef4a39a


mgallego-keeper (commented):
Reviewed for KSM-1168. The core is solid: `--prefix` is required for record/folder mode, it is prepended to every derived name, dry-run redaction is applied on both the `--map` path (`sync_aws_with_client`) and the record/folder JSON path (`sync_aws_json_with_client`), and `_sanitize_title_for_prefix` covers the leading-slash and `..` cases. A few gaps to consider (two are inline):

1. **Prefix separator** (inline at sync.py:1179 and :1488): a `--prefix` without a trailing slash concatenates straight into the title with no separator. Normalize the join or validate the prefix.

2. **Overwrite within the prefix** (optional, defense in depth): `--prefix` confines the namespace, but `_set_secret_aws` (put_secret_value at sync.py:383) still overwrites any existing secret under the prefix whose value differs, including one this sync did not create. If that matters, tag secrets the sync creates and refuse to overwrite an untagged one unless `--force`, plus a `--no-overwrite`. Not required for this fix; a reasonable follow-up.

3. **Test coverage**:
   - The record-based dry-run redaction (the `mapping["json_key"] is None` branch, which is the path `--record`/`--folder` actually use) is not directly covered. `test_sync_aws_json_dry_run` uses a `kms_key+json_key` mapKey, which drives the map-based (`else`) branch, and `test_sync_prefix_prepended_to_record_title` mocks `sync_aws_json_with_client`. Worth a test that runs a record-based entry through the dry run and asserts the live value is absent.
   - `_sanitize_title_for_prefix` traversal and leading-slash (inline at sync.py:752).
   - Prefix application on the `--folder`/`--folder-recursive` path (sync.py:1488) is untested; only `--record` is.

4. **Minor**: the dry-run output now drops the `dstValue` key entirely (`pop`) instead of setting it to null. A consumer reading `entry["dstValue"]` would now KeyError; `entry.get("dstValue")` is fine. Keeping the key with a null value would be more backward compatible if output stability matters.

Azure and GCP have the same dry-run disclosure and no name validation; tracked separately in KSM-1169. I have the record-based dry-run and traversal tests from a now-closed duplicate branch and can contribute them if useful.

integration/keeper_secrets_manager_cli/keeper_secrets_manager_cli/sync.py
```diff
@@ -1166,12 +1172,13 @@ def _process_aws_records_and_folders(self, result, records, folders, folders_rec
...
 
             for record_obj in resolved_records:
-                # Validate record title for AWS compatibility
-                secret_name, error_msg = self._validate_aws_secret_name(record_obj.title)
-
+                safe_title, sanitize_error = self._sanitize_title_for_prefix(record_obj.title)
+                if sanitize_error:
+                    validation_errors.append(f"'{record_obj.title}' (UID: {record_obj.uid}): {sanitize_error}")
+                    continue
+                secret_name, error_msg = self._validate_aws_secret_name(prefix + safe_title)
```
mgallego-keeper, line 1232: With `--prefix keeper/` this yields `keeper/my-secret`, but with `--prefix keeper` (no trailing slash) it yields `keepermy-secret`: `prefix + safe_title` concatenates with no separator, and `_sanitize_title_for_prefix` has already stripped the leading slash from the title. The help text asks for a trailing slash but nothing enforces or normalizes it, so a missing slash silently merges the prefix into the title (same at the folder path, line 1488). Consider joining with a single slash (`prefix.rstrip('/') + '/' + safe_title`) or validating that `--prefix` ends with a separator.

integration/keeper_secrets_manager_cli/keeper_secrets_manager_cli/sync.py
```diff
@@ -749,6 +749,12 @@ def _validate_aws_secret_name(self, name):
         error_msg = "; ".join(errors) if errors else None
         return name, error_msg
 
+    def _sanitize_title_for_prefix(self, title):
```
mgallego-keeper, line 788: This is the check that stops a title from escaping the prefix, but there is no test for it. Worth a unit test asserting a title like `../../elsewhere` is rejected and a leading-slash title (`/foo`) is stripped, so a later refactor cannot silently drop the `..` guard.

stas-schaller:
Point 2 (ownership guard — tag on create, refuse to overwrite untagged secrets within the prefix) filed as KSM-1172 for a follow-on. The extra IAM permission and per-update API call made it too wide for this fix.

mgallego-keeper (changes requested):
Re-reviewed at 9fa94cb. Four of the five points landed cleanly. The fix for point 3a did not: it introduced a functional regression on the exact path this PR is about, and the new test for that path cannot detect it.

Thanks for filing KSM-1172 for the ownership guard, the rationale on IAM scope and the extra per-update call is fair.

### Status of the previous review

| Point | Status |
| --- | --- |
| 1. Prefix separator | **Fixed.** Verified all six accept/reject combinations: `-px keeper` rejected, `keeper/` and `myapp-` accepted, `-px ''` caught by the required-check, `-r`/`-f`/`-fr` each require it. |
| 2. Overwrite within prefix | **Deferred** to KSM-1172. Agreed, I flagged it as optional. |
| 3a. Record-based dry-run redaction test | **Test added but ineffective**, and the code it covers is now wrong. See below. |
| 3b. Sanitizer traversal / leading-slash tests | **Added.** The guard's wiring into the two loops is still uncovered. |
| 3c. Folder-path prefix test | **Added.** `test_sync_prefix_prepended_to_folder_record` asserts the concatenation. |
| 4. `dstValue` backward compatibility | **Fixed.** Key retained with a null value on both AWS paths. |

---

## Blocker: `dstDiffers` is always `true` for `--record` / `--folder` dry runs

`sync.py:1862` compares `current_value` (a `str` read back from AWS) against `mapping["srcValue"]`, which on the record-based branch is a **dict**: `_generate_record_json` returns `record.dict` for `--raw-json` and the `flattened` dict otherwise. A `str` is never equal to a `dict`, so the expression is unconditionally `True`.

Driven end to end through `sync_values`, with the AWS secret already holding exactly what the sync would write:

```
mapKey     = keeper/my-secret     <- prefix applied correctly
dstValue   = None                 <- redacted correctly
dstExists  = True
dstDiffers = True                 <- wrong
real (non-dry) run on the same input: 0 PutSecretValue calls
```

Repro:

```python
src_dict = {"login": "admin", "password": "p@ss"}          # what _generate_record_json returns
aws_current = json.dumps(src_dict, separators=(',', ':'))  # what a previous identical sync wrote

maps = [{"mapKey": "keeper/my-secret", "mapNotation": "record:rec_uid",
         "srcValue": src_dict, "dstValue": None}]
with patch.object(sync, '_get_secret_aws', return_value={"value": aws_current}):
    sync.sync_aws_json_with_client(Mock(), dry_run=True, maps=maps)

assert maps[0]["dstDiffers"] is False   # fails: it is True
```

This matters because `dstDiffers` is the field that *replaced* `dstValue`. It is the operator's only remaining signal about whether a run will change anything, and on the `--record`/`--folder` path it is pinned to "yes" for every secret that already exists. The map-based branch at line 1867 is correct, both sides are strings there; only the `json_key is None` branch is affected.

The non-dry-run path at lines 1888-1903 already does this comparison properly. Suggested fix, mirroring it:

```python
                    if mapping["json_key"] is None:
                        # Full JSON content (record-based). srcValue is a dict here,
                        # so compare it the way the real run below does rather than
                        # with a raw != against the string read back from AWS.
                        mapping["original"]["dstExists"] = current_value is not None
                        if current_value is None:
                            mapping["original"]["dstDiffers"] = None
                        else:
                            src_value = mapping["srcValue"]
                            src_json = src_value if isinstance(src_value, dict) else json.loads(src_value)
                            try:
                                current_json = json.loads(current_value)
                                current_cmp = json.dumps(
                                    {k: v for k, v in current_json.items() if k != "_preserved_plaintext"},
                                    sort_keys=True, separators=(',', ':'))
                                src_cmp = json.dumps(
                                    {k: v for k, v in src_json.items() if k != "_preserved_plaintext"},
                                    sort_keys=True, separators=(',', ':'))
                                mapping["original"]["dstDiffers"] = current_cmp != src_cmp
                            except (json.JSONDecodeError, TypeError, AttributeError):
                                # Destination holds plaintext; the real run rewrites it
                                # with _preserved_plaintext, so it does differ.
                                mapping["original"]["dstDiffers"] = True
```

---

## Test gaps

**1. The security control itself has zero coverage.** A coverage run over `tests/sync_test.py` shows `__main__.py` lines 1583, 1584, 1588 and 1589 unexecuted. Nothing tests that a record without `--prefix` is rejected, or that the separator rule fires. `test_sync_map_does_not_require_prefix` does not cover this: it calls `sync_values`, which never enforced `--prefix` at any point, so the assertion it is making is vacuous, and it has no assertions at all. As it stands the whole Bugcrowd fix could be deleted in a refactor and the suite would stay green.

`tests/conftest.py` already exports a `CliRunner` wrapper used by `config_test`, `exec_test`, `init_test` and `misc_test`, and `sync_test.py` already imports `CliRunner` (currently unused). Three cases close this:

```python
runner.invoke(sync_command, ["-c", "UID", "-t", "aws", "-r", "rec"])                    # rejected
runner.invoke(sync_command, ["-c", "UID", "-t", "aws", "-r", "rec", "-px", "keeper"])   # rejected
runner.invoke(sync_command, ["-c", "UID", "-t", "aws", "-r", "rec", "-px", "keeper/"])  # accepted
```

**2. `test_dry_run_record_based_does_not_disclose_dst_value` masks the blocker.** It passes `srcValue` as a string for a `record:` entry, but production always passes a dict there. Because it only asserts `dstDiffers is True` against a genuinely different pair, it passes both with the realistic fixture and with the current bug. It needs a dict `srcValue` plus an identical-destination case asserting `dstDiffers is False`.

**3. The traversal guard's wiring is uncovered.** `sync.py:1177-1178` (record loop, append error and `continue`) and `sync.py:1486-1487` and `1490` (folder loop) are all unexecuted. `_sanitize_title_for_prefix` is tested standalone, but nothing asserts that a `..` title actually aborts a record or folder sync, which was the point of the original comment.

---

## Minor, not blocking

- **`--type json` bypasses every prefix rule.** All the new validation sits inside `if sync_type != 'json':` at `__main__.py:1573`. Verified: `-t json -r rec` is accepted with `prefix=None`, and `-t json -r rec -px keeper` accepts the alphanumeric-terminated prefix that `--type aws` rejects. Nothing reaches AWS so it is not a hole, but it is the natural preview path and its `mapKey`s no longer match what an `aws` run would produce.
- **The prefix is never validated against the AWS charset.** `-px 'keeper:'` passes the CLI check (`:` is not alphanumeric) and then fails once per record with `'foo' (UID: U1): contains invalid characters: ':'`, attributing the failure to the record title rather than the prefix. `-px 'keeper '` silently becomes `keeper_` via the space replacement in `_validate_aws_secret_name`. The 512-character limit now includes the prefix, with the same misattribution.
- **`--prefix` is accepted and silently ignored** for map-only aws runs, where `--raw-json` warns in the equivalent situation. Conversely `-t azure -m k v -px keeper` is *rejected* with the AWS-flavoured separator message even though `--prefix` is meaningless for azure.
- **Behaviour change in json mode:** a record titled `foo/../bar` under `-t json` previously produced that mapKey and now errors with "title contains '..' path segment, which is not allowed with --prefix", referencing a flag the user did not pass. Leading slashes are now stripped there too.
- The commit message for f055a15 says the destination value "is no longer fetched or printed". It is still fetched, necessarily, to compute `dstDiffers`. Only the printing changed. Worth correcting so the Bugcrowd write-up matches the code.

---

## What is solid

- Prefix enforcement and application are correct for aws record, folder and folder-recursive modes, and the prefix genuinely reaches the AWS `SecretId` (`keeper/my-secret`).
- Containment holds independently of the `..` guard: Secrets Manager treats names as opaque strings, and `:` (needed to forge an ARN in `SecretId`) is outside the allowed charset, so the string prefix cannot be escaped. The guard is good defence in depth.
- Both AWS dry-run paths redact. GCP (`sync.py:2084`) and Azure (`sync.py:1740`) still disclose, correctly deferred to KSM-1169.
- Full CLI suite green on this branch: 179 passed, 5 skipped.
- No other consumer in the repo reads `dstValue`.

---

## One CI note for the QA deploy

`test.cli.yml` triggers on `pull_request: branches: [ master ]`. This PR targets `release/tool/cli/v1.5.0`, so the CLI unit tests never run on it. The only checks here are the two Socket Security ones. Nothing will catch the gaps above automatically until the release branch merges to master, which makes gap 1 more consequential than it would normally be. Might be worth adding the release branches to that workflow's trigger list separately.

mgallego-keeper (approved):
Re-reviewed at 59c0b0c. All blocking items from the previous review are resolved. Verified locally: full suite 183 passed / 5 skipped on the branch (Python 3.13, click 8.4.2, boto3 present).

### Status of the previous review

| Item | Status |
| --- | --- |
| Blocker: `dstDiffers` always true on record/folder dry runs | **Fixed** at sync.py:1863. Re-ran the original end-to-end repro through `sync_values`: identical destination now reports `dstExists=True, dstDiffers=False, dstValue=None`, and a real run on the same input makes 0 PutSecretValue calls, so the dry-run prediction matches the real action. Also checked: differing destination true; missing destination false/null; plaintext destination true (real run rewrites it); reordered-but-equal JSON false. The comparison mirrors the real-run path exactly, and the extra `AttributeError` catch makes the dry run strictly more robust than the real run. |
| Gap 1: CLI-level enforcement had zero coverage | **Closed** by `test_sync_aws_prefix_enforcement`; it drives `__main__.py:1583-1592` through CliRunner. The message assertions are click-version-safe: click 8.0/8.1 merge stderr into `result.output`, and click 8.2+ makes `output` an ordered mix of stdout and stderr (verified against 8.4.2). |
| Gap 2: record-based dry-run test masked the blocker | **Closed**. The fixture now uses a dict `srcValue` as production does, and `test_dry_run_record_based_dst_differs_false_when_content_matches` pins the identical-destination case. |
| Gap 3: traversal guard wiring uncovered | **Closed**. Both tests reach the real guards (record: sync.py:1189, folder: sync.py:1498), which fire before `_get_aws_client`, so no AWS interaction is involved. |

I also ran mutation checks to make sure the new tests carry real detection power, since a masked test is what slipped through last round: reverting sync.py to 9fa94cb fails the new dstDiffers test; deleting the prefix guard from `__main__.py` fails the enforcement test; neutering `_sanitize_title_for_prefix` fails both dotdot tests. The suite can no longer stay green if this fix is deleted.

### Heads-up for merging

#1099 and #1100 merged into `release/tool/cli/v1.5.0` today, and this branch now conflicts in sync.py: both this PR and #1099 insert methods right after `_validate_aws_secret_name`. It is a pure insertion collision; the resolution is the union (keep `_sanitize_title_for_prefix` plus `_dst_status`, `_validate_azure_secret_name`, `_validate_gcp_secret_name`). README and sync_test.py auto-merge. I ran the full suite on a trial merge resolved that way: 188 passed, 5 skipped. Also confirmed #1099's `_dst_status` (plain `!=`) is only used by the azure/gcp map paths, which are string-to-string and reject `--record`/`--folder`, so it cannot reintroduce the dict-comparison bug.

### Non-blocking notes

1. `test_sync_map_does_not_require_prefix` is still assertion-free, and the "--map mode is unaffected" promise has no CLI-level case. A fourth invocation in `test_sync_aws_prefix_enforcement` (`-t aws -m key notation` with no `-px`, assert `sync_values` called) would close both; fine as a fast-follow.
2. The destination-missing (`dstDiffers=None`) and plaintext-destination (except) branches of the new dry-run code are verified by repro but have no direct unit test; two small cases would pin them.
3. Pre-existing rather than from this PR: the real-run record-based branch catches only `JSONDecodeError`/`TypeError` (sync.py:1923), so a destination holding valid non-object JSON (a list, number, or quoted string) crashes with `AttributeError` on `.items()`. The new dry-run code handles that same input gracefully. Worth folding into KSM-1172 or a small follow-up.

mgallego-keeper:
Verified the post-approval force-push as well (59c0b0c to cc36b9f):

- The rebase onto 21b22ed preserves the reviewed content exactly: per-file net diffs are identical to 59c0b0c, and the only sync.py difference against my own trial-merge resolution is the placement of `_sanitize_title_for_prefix` relative to the #1099 helpers. The merge-conflict heads-up in my review is already handled by the rebase.
- The new commit cc36b9f (test.cli.yml trigger on `release/tool/cli/**`) closes the CI note from my first review, and it already proved itself here: the synchronize event ran Test-CLI on this PR, all 8 jobs green (3.10 to 3.13, standard and keyring).
- Local suite on cc36b9f: 189 passed, 5 skipped.

Approval stands.

PR #1098, opened 2026-08-07
fix(cli): confine ksm sync AWS destination names and redact dry-run output (KSM-1168)

mgallego-keeper:
## Summary
Hardening for `ksm sync --type aws` when syncing records or folders. Fixes KSM-1168.

## Changes
- **Require `--prefix`** for `--record`/`--folder`/`--folder-recursive`. It is prepended to every AWS secret name derived from a record title and confined against `..` and leading-slash escapes, so a derived name stays within the operator-controlled namespace.
- **Redact dry-run output** in both the plain (`sync_aws_with_client`) and record/folder (`sync_aws_json_with_client`) paths: the output now reports `dstExists` and `dstDiffers` instead of the live destination value (`dstValue`).

## Behavior change
Existing `--record`/`--folder` commands must add `--prefix` (for example `--prefix keeper/`). Noted in the CLI changelog.

## Tests
Adds regression tests for prefix confinement (including traversal rejection) and for dry-run redaction on both AWS code paths. `pytest tests/` passes (176 passed, 5 skipped).

Detail and rationale: KSM-1168.


mgallego-keeper:
Closing as a duplicate of #1097 (Stas), which is the canonical fix for KSM-1168 and lands the same two changes: required --prefix confinement and dry-run dstValue redaction. Confirmed #1097 redacts both the --map path (sync_aws_with_client) and the record/folder path (sync_aws_json_with_client), and its _sanitize_title_for_prefix handles the leading-slash and '..' cases, so no gap remains there. This branch's extra regression tests (record/folder dry-run redaction and prefix traversal rejection) can be folded into #1097 if useful.

PR #1099, opened 2026-08-07
fix(cli): redact ksm sync Azure/GCP dry-run output and validate destination names (KSM-1169)

mgallego-keeper:
## Summary
Applies the KSM-1168 hardening to the Azure and GCP `ksm sync` backends. Fixes KSM-1169 (relates to KSM-1168).

## Changes
- **Redact dry-run output** for Azure (`sync_azure`) and GCP (`sync_gcp`) via a shared `_dst_status()` helper: the output reports `dstExists`/`dstDiffers` instead of the live destination value.
- **Validate destination names** for Azure (Key Vault rules) and GCP (Secret Manager rules) before a write; an invalid name is reported per entry and skipped.

## Notes
Reachable today only through operator-supplied `--map` keys, since `--record`/`--folder` mode is AWS-only; the dry-run value disclosure still reaches the sync log. `--record`/`--folder` should not be extended to Azure or GCP until prefix confinement (KSM-1168) exists for them.

## Tests
Adds unit tests for `_dst_status` redaction and the Azure/GCP name validators. `pytest tests/` passes. End-to-end Azure/GCP sync tests require the optional Azure/GCP SDKs (the repo has no Azure/GCP e2e suite today); the logic changes are covered by the helper and validator unit tests.

Detail and rationale: KSM-1169.


integration/keeper_secrets_manager_cli/keeper_secrets_manager_cli/sync.py
```diff
@@ -749,6 +749,37 @@ def _validate_aws_secret_name(self, name):
...
+            "dstExists": current_value is not None,
+            "dstDiffers": (current_value != src_value) if current_value is not None else None,
+        }
+
+    def _validate_azure_secret_name(self, name):
+        """Return an error message if name is not a valid Azure Key Vault secret name, else None.
+
+        Azure Key Vault permits only alphanumeric characters and dashes, 1 to 127 characters.
+        """
+        if not name or not re.match(r'^[0-9A-Za-z-]{1,127}$', name):
```
stas-schaller, line 775: Azure Key Vault requires secret names to start with a letter. The current regex `^[0-9A-Za-z-]{1,127}$` accepts names like `123-secret`, which pass this check but fail at the API with a generic error.

I recommend:

```python
if not name or not re.match(r'^[A-Za-z][0-9A-Za-z-]{0,126}$', name):
    return "Azure Key Vault secret names must start with a letter and may contain only alphanumeric characters and dashes (1 to 127 characters)"
```

And a test case to cover it:

```python
self.assertIsNotNone(self.sync._validate_azure_secret_name("1invalid"))
```

The same regex also accepts trailing hyphens (`name-`) and consecutive hyphens (`a--b`), which Azure also rejects. Worth adding coverage for those while you're here, though the start-with-letter gap is the most likely real-world hit.

stas-schaller (changes requested):
One regex upgrade and we're good to go!

integration/keeper_secrets_manager_cli/keeper_secrets_manager_cli/sync.py
mgallego-keeper: I checked this against the Azure sources before changing the regex, and the start-with-letter rule turns out to apply to vault names, not secret names:

- Microsoft Learn, [About keys, secrets and certificates](https://learn.microsoft.com/en-us/azure/key-vault/general/about-keys-secrets-certificates): an `object-name` (secrets included) "must be a 1-127 character string, containing only 0-9, a-z, A-Z, and -". The "not consecutive -" clause in that table belongs to `vault-name` (3-24 chars).
- Key Vault data-plane REST spec, SetSecret `secret-name` path parameter: `"pattern": "^[0-9a-zA-Z-]+$"` (azure-rest-api-specs, `specification/keyvault/data-plane/Legacy/stable/7.4/secrets.json`).
- PSRule for Azure, [Azure.KeyVault.SecretName](https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.KeyVault.SecretName/): "Between 1 and 127 characters long. Alphanumerics and hyphens (dash)."

So `123-secret`, `name-` and `a--b` are all accepted by the service, and requiring a leading letter would make sync refuse names Azure allows (the suggested regex would also still accept `name-` and `a--b`). I kept the regex aligned with the spec and instead pinned those edge shapes plus the 127/128 length boundary in `test_validate_azure_secret_name`, and added the 255/256 boundary for GCP, in 2d76dee1, so any future tightening shows up as a deliberate test change. Happy to switch to the stricter pattern if you have a repro of the API rejecting these.

stas-schaller (approved):
👍

PR #1100, opened 2026-08-07
fix(cli): warn when ksm sync resolves --record/--folder by mutable name (KSM-1170)

mgallego-keeper:
## Summary
Hardening for `ksm sync` target resolution. Fixes KSM-1170 (found while auditing KSM-1168).

## Changes
- `_resolve_records` and `_resolve_folders` now print a stderr warning, naming the resolved UID, when a `--record`/`--folder` token resolves by mutable record title or folder name/path rather than by UID.

## Notes
Resolution behavior is otherwise unchanged: ambiguous matches still error (fail closed). This is a visibility and hardening change to steer scheduled syncs toward UIDs.

## Tests
Adds tests that a title/name resolution warns and a UID resolution does not. `pytest tests/` passes.


integration/keeper_secrets_manager_cli/tests/sync_test.py
```diff
@@ -1061,6 +1061,33 @@ def test_sync_values_record_validation_errors_aggregated(self):
...
+        folder.folder_uid = "BBBBBBBBBBBBBBBBBBBBBB"
+        folder.name = "prod"
+        folder.parent_uid = ""
+        self.cli_mock.client.get_folders.return_value = [folder]
+        with patch('keeper_secrets_manager_cli.sync.click.echo') as mock_echo:
+            by_name = self.sync._resolve_folders(["prod"])
+        self.assertEqual(len(by_name), 1)
+        self.assertEqual(by_name[0].folder_uid, "BBBBBBBBBBBBBBBBBBBBBB")
+        warnings = " ".join(str(c) for c in mock_echo.call_args_list)
+        self.assertIn("resolved by name or path", warnings)
```
stas-schaller: This only verifies the warning fires for name/path resolution. It doesn't verify that UID-based resolution is silent, so a regression that made `_resolve_folders` warn on UID lookups would pass this test.

`test_resolve_records_by_title_warns` handles this correctly by resolving both ways in the same mock context and asserting an exact count of 1. The folder test should do the same:

```python
with patch('keeper_secrets_manager_cli.sync.click.echo') as mock_echo:
    by_name = self.sync._resolve_folders(["prod"])
    by_uid = self.sync._resolve_folders(["BBBBBBBBBBBBBBBBBBBBBB"])
self.assertEqual(len(by_name), 1)
self.assertEqual(by_name[0].folder_uid, "BBBBBBBBBBBBBBBBBBBBBB")
self.assertEqual(len(by_uid), 1)
self.assertEqual(by_uid[0].folder_uid, "BBBBBBBBBBBBBBBBBBBBBB")
warnings = " ".join(str(c) for c in mock_echo.call_args_list)
# Exactly one warning, from by-name only.
self.assertEqual(warnings.count("resolved by name or path"), 1)
```

stas-schaller (changes requested):
Minor test coverage gap then we're good

integration/keeper_secrets_manager_cli/tests/sync_test.py
mgallego-keeper: Done in e6b5b3c8: the folder test now resolves by name and by UID in the same mock context and asserts exactly one warning, matching the records test, so a regression that made `_resolve_folders` warn on UID lookups fails the test. It also asserts the warning names the resolved folder UID.

stas-schaller (approved):
😃

PR #1101, opened 2026-08-07
fix(cli): build ksm init k8s manifest with a YAML serializer (KSM-1171)

mgallego-keeper:
## Summary
Hardening for `ksm init k8s` manifest generation. Fixes KSM-1171 (found while auditing KSM-1168).

## Changes
- `Init.get_k8s` now builds the Secret manifest as a dict and serializes it with `yaml.safe_dump` instead of string-formatting `--name`/`--namespace` into a YAML template. Both values are always emitted as encoded scalars, so they cannot inject additional manifest content. `pyyaml` is already a runtime dependency.

## Tests
Adds a regression test that a newline-bearing `--name` is preserved as a single scalar and injects no extra manifest key. Existing `init` tests pass unchanged.


integration/keeper_secrets_manager_cli/tests/init_test.py
```diff
@@ -172,3 +172,53 @@ def test_k8s(self):
...
+                    malicious = "evil\ninjected: pwned"
+                    runner = CliRunner()
+                    result = runner.invoke(cli, [
+                        'init ', 'k8s', token,
+                        '--name', malicious,
+                        '--namespace', 'default'
+                    ], catch_exceptions=False)
+                    self.assertEqual(0, result.exit_code, "k8s init did not succeed")
+
+                    script = yaml.load(StringIO(result.output), yaml.Loader)
```
stas-schaller: `yaml.Loader` is the unsafe PyYAML loader, ie it can execute arbitrary Python via `!!python/object` tags. The output here comes from `yaml.safe_dump` so no exploit is reachable, but the pattern is still wrong (the existing `test_k8s` at line 153 has the same issue). Let's switch to `yaml.safe_load`:

```python
script = yaml.safe_load(StringIO(result.output))
```

integration/keeper_secrets_manager_cli/tests/init_test.py
```diff
@@ -172,3 +172,53 @@ def test_k8s(self):
...
+                mock_init_client.return_value = init_secrets_manager
+                with patch('keeper_secrets_manager_cli.init.Init.init_config') as mock_init_config:
+                    mock_init_config.return_value = init_config
+
+                    token = "US:MY_TOKEN"
+                    malicious = "evil\ninjected: pwned"
+                    runner = CliRunner()
+                    result = runner.invoke(cli, [
+                        'init ', 'k8s', token,
+                        '--name', malicious,
```
stas-schaller, line 215: `--namespace` gets a benign value here, but the PR description says both name and namespace are always emitted as encoded scalars. If the namespace path were ever changed back to string formatting, this test wouldn't catch it. Let's give namespace the same malicious input and assert it round-trips:

```python
malicious_ns = "evil\nns-injected: pwned"
# ... '--namespace', malicious_ns ...
self.assertNotIn("ns-injected", script)
self.assertEqual(malicious_ns, script["metadata"]["namespace"])
```

stas-schaller (changes requested):
Two security minded fixes and we're good

integration/keeper_secrets_manager_cli/tests/init_test.py
mgallego-keeper: Done in 44906401: switched to `yaml.safe_load` here, and fixed the same pattern in `test_k8s` (line 153) as well.

integration/keeper_secrets_manager_cli/tests/init_test.py
mgallego-keeper: Done in 44906401: `--namespace` now gets its own malicious payload with the round-trip assertions you suggested, plus a check that `metadata` contains exactly `name` and `namespace` (catches indented injections that would land inside `metadata` rather than at the top level). Also added `--immutable` coverage while in here, since the manifest builder rewrite changed that path and it had none.

stas-schaller (approved):
Good stuff!

PR #1103, opened 2026-08-12
fix(cli): gate sync cloud-deps error on frozen binary, fix AWS zsh quoting

stas-schaller:
## Summary
CLI: `ksm sync --type aws|azure|gcp` gave pip install advice for a missing cloud dependency even in a frozen (PyInstaller) binary install, where pip cannot fix anything. This became user-reachable in 1.5.0 because KSM-1106 made component-less installer builds a supported GUI configuration.

### Fixed
- All three cloud-dependency `ImportError` handlers (Azure, AWS, GCP) in `sync.py` now detect `sys.frozen` and, when frozen, point the user at re-running the installer and enabling the "Cloud Sync" component instead of printing a pip command. Non-frozen (pip) installs keep the original pip advice, unchanged. (KSM-1182)
- The AWS pip command is now single-quoted (`pip install 'keeper-secrets-manager-cli[aws]'`). The unquoted form is a glob pattern in zsh and fails with "no matches found" when copy-pasted on macOS's default shell. (KSM-1182)

## Testing
```
cd integration/keeper_secrets_manager_cli
python3 -m pytest tests/sync_test.py -x -q
```
6 new tests in `SyncMissingCloudDependencyTest` cover the frozen/non-frozen branch for each of the three providers, plus the AWS quoting fix specifically.

## Breaking Changes
None.

## Related Issues
- Jira: KSM-1182


mgallego-keeper:
👍

PR #1104, opened 2026-08-12
fix(cli): validate RFC 1123 secret name in ksm init k8s

stas-schaller:
## Summary
CLI: `ksm init k8s --apply` passed `--name` to `kubectl` as a positional argument with no validation. A value beginning with `-` was consumed by `kubectl` as one of its own flags rather than as the secret name (CWE-88, argument injection). KSM-1171 fixed the manifest (non-apply) branch by switching to a YAML serializer, but left the `--apply` branch unaddressed.

### Security
- `--name` is now validated against the RFC 1123 subdomain rule Kubernetes applies to Secret names (lowercase alphanumeric characters, `-` or `.`, each `.`-separated label starting and ending alphanumeric, at most 253 characters total) before it reaches either output branch of `get_k8s()`. A validated name can never begin with `-`, closing the injection surface without needing a `--` separator. (KSM-1183)
- Validation now runs in a click `callback=` at argument-parse time, before `Init()` redeems the one-time token over the network. Previously, a rejected `--name` still burned the token, so a typo forced the user to mint a new one.

### Fixed
- The regex originally implemented the RFC 1123 *label* rule (no dots), rejecting legal Kubernetes Secret names like `tls.example.com` or `v1.2.3` that worked before this PR. It now implements the subdomain rule Kubernetes actually applies.
- A trailing newline (`"my-secret\n"`) slipped past the original regex, since Python's `$` matches before a string-final newline. Switched to `fullmatch`.
- Names that are legal Kubernetes Secret names but ambiguous in YAML 1.1 (`y`, `n`, `1e5`) were emitted unquoted in the manifest branch. `kubectl`'s stricter YAML 1.1 parser reads them as bool/number instead of string. They're now always quoted.

A `--` end-of-options separator was considered and rejected: `kubectl` derives NAME via `ArgsLenAtDash()`, which counts only positionals appearing before `--`, so `kubectl create secret generic -- my-secret` reports `exactly one NAME is required, got 0`, and flags placed after `--` are swallowed as positionals instead of being parsed. There is no argument order where `--` both preserves NAME and keeps `--from-literal` parsed as a flag, so the regex validation is the actual fix.

## Testing
```
cd integration/keeper_secrets_manager_cli
python3 -m pytest tests/init_test.py -x -q
```
Covers: a flag-shaped `--name` is rejected under `--apply` without ever calling `kubectl`; a valid name, including dotted names and the 253-character boundary, reaches `kubectl` unmodified; RFC 1123 validation applies to both the `--apply` and manifest branches; an invalid `--name` never redeems the one-time token; and YAML-1.1-ambiguous names are quoted in the manifest output.

## Breaking Changes
None.

## Related Issues
- Jira: KSM-1183


mgallego-keeper (changes requested):
**Verdict: request changes.** The approach is correct and the hard part of the PR is right: I verified that the apply branch passes `name` as a positional in a `subprocess.run` argv list (no shell), that click really does hand a flag-shaped value through to `--name` (the CI-green injection test proves it), and that the PR body's rejection of a `--` separator is technically accurate (kubectl's `NameFromCommandArgs` counts positionals via `ArgsLenAtDash`, so `--` before the name yields "exactly one NAME is required, got 0"). Validation is the right fix. But the validation rule itself has two defects.

### Finding 1 (medium, functional regression): the regex bans dots, which Kubernetes Secret names allow

`K8S_NAME_PATTERN = r'^[a-z0-9]([a-z0-9-]{0,251}[a-z0-9])?$'` implements an RFC 1123 **label** rule stretched to 253 chars. But the comment, error message, README entry, and PR description all claim the RFC 1123 **subdomain** rule, which is what Kubernetes actually applies to Secret names (apimachinery `NameIsDNS1123Subdomain` via `NameIsDNSSubdomain`), and subdomains permit dots. Comparing this regex against the real Kubernetes rule, the only mismatches are dotted names:

| name | this PR | Kubernetes |
|---|---|---|
| `tls.example.com` | rejected | valid |
| `prod.db.credentials` | rejected | valid |
| `v1.2.3` | rejected | valid |

These names worked in both branches before this PR, so this is a regression on legal input, and it contradicts the PR's own claim that only names Kubernetes would reject anyway now fail early. Suggested fix: use Kubernetes' own regex plus a separate length check (k8s checks length separately too, since total length can't be encoded cleanly once dots are involved):

```python
K8S_NAME_PATTERN = re.compile(r'^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$')
# plus: if len(name) > 253: raise ...
```

Allowing dots does not reopen the injection surface: the first character still must be alphanumeric, so a leading `-` remains impossible. The error message ("lowercase alphanumeric characters or '-'") should become "'-' or '.'", the module comment and README entry need the same correction, and a dotted-name positive test would have caught this (`test_k8s_rejects_invalid_names` has no valid-dotted case).

### Finding 2 (low): `$` accepts a trailing newline; use `re.fullmatch`

Python's `$` matches just before a string-final newline, so `K8S_NAME_PATTERN.match("my-secret\n")` succeeds (verified empirically). A user passing `--name $'my-secret\n'` sails past the new validation and then fails later at kubectl or the API with a raw Kubernetes error instead of the friendly CLI message. No security impact (`^` still anchors the first character), but a validation gate should be exact: switch `.match(name)` to `.fullmatch(name)` in `get_k8s()`. The existing newline test only covers a mid-string newline, which is why it passes.

### Minor and optional

* Namespace stays unvalidated. Kubernetes namespaces are RFC 1123 **labels** (63 chars, no dots). The manifest branch is injection-safe via the KSM-1171 YAML serializer, so this is only a friendliness gap, but a parity check would be cheap.
* Boundary test: 254 chars is tested as invalid; adding the 253-char positive case would pin the boundary (253 passes the current regex).
* The pre-existing `--apply` branch issues that the new test's argv assertion makes visible (ignored `--namespace`/`--immutable`, no kubectl error handling) are tracked separately as KSM-1189 rather than growing this PR.

The test work is otherwise good: `_patched_cli` follows the existing mock patterns, and `test_k8s_apply_valid_name` asserts exactly the right property (the name is its own argv element and not flag-shaped).

mgallego-keeper:
Follow-up to my review after a deeper verification pass: one more finding I'd fold into the same change, one adjacent gap the new docstring makes relevant, and three test-hardening nits.

### Finding 3 (medium): a rejected `--name` burns the one-time token

`init_k8s_command` constructs and calls in one expression (`__main__.py:1364`), and `Init.__init__` redeems the token before `get_k8s()` ever runs: `redeem_sm.get_secrets()` binds the device, after which the SDK deletes the client key (reusing the token then fails with "Initialization failed: ... Please generate a new One-Time Token and try again"). Nothing is persisted at that point (`Config()` has no backing file and `set_profile` is in-memory), and the new validation is the first statement of `get_k8s()`. Net effect: `ksm init k8s --name MySecret <token>` spends a single-use token over the network, then fails on a purely local argv check, and the user must mint a fresh token to fix a typo. Before this PR the same command still emitted a usable manifest. This also interacts with Finding 1: currently-legal dotted names hit this path too.

Suggested fix: validate at parse time via a click `callback=` on the `--name` option (the CLI already uses this idiom: `validate_non_empty`, wired via `callback=` on six options in `__main__.py`); the `get_k8s()` check can stay as a cheap defensive guard.

### Adjacent: regex-valid names still break the manifest branch (YAML 1.1)

Names the new pattern accepts, like `y`, `n`, and `1e5` (all valid Kubernetes Secret names), are emitted unquoted by `yaml.safe_dump`, and kubectl's YAML 1.1 parser reads `y`/`n` as booleans and `1e5` as a number. Reproduced with the exact manifest shape `get_k8s()` emits for `--name y` (PyYAML leaves `name: y` unquoted):

```
$ <manifest for --name y> | kubectl apply --dry-run=client -f -
error: unable to decode "STDIN": json: cannot unmarshal bool into Go struct field ObjectMeta.metadata.name of type string
```

The same name under `--apply` works, since argv bypasses YAML. The serializer behavior predates this PR (it came with the KSM-1171 change), but this PR removed the only assertions that pushed `--name` through the serializer, and the updated docstring states `--name` can no longer reach the serializer with problematic content, which these cases contradict. Since the names are legal in Kubernetes, the fix belongs in the serializer (force string/quoted style for the `name` and `namespace` scalars) plus a restored name-through-manifest assertion. `--namespace y` has the same issue.

### Test nits

* `test_k8s_apply_rejects_dash_name` exercises `--apply` without patching `keeper_secrets_manager_cli.init.subprocess.run` (its positive sibling patches it), so the validation under test is the only barrier keeping the negative test from executing real kubectl if the check ever regresses. Hoisting the patch into `_patched_cli` and adding `mock_run.assert_not_called()` closes that.
* `test_k8s_rejects_invalid_names` loops six cases without `self.subTest`, so the first failing name masks the other five.
* `_patched_cli` carries some inert scaffolding: the `KeeperCli.get_client` patch is never hit and only one of the four queued responses is consumed per invocation, so it can be slimmed (and the three older inline copies of the same setup could be converted to it).

PS: now that #1103 has merged, this branch has the add/add changelog conflict in README.md (both PRs anchored their bullet after the KSM-1171 line); a rebase placing the KSM-1183 bullet after the KSM-1182 one resolves it.

stas-schaller:
Thanks for the deep pass, @mgallego-keeper. Addressed all four findings plus the test nits:

- **Finding 1 (dots)**: switched the pattern to the RFC 1123 subdomain rule, using your suggested shape (with `fullmatch`). `tls.example.com`, `v1.2.3`, etc. are valid again.
- **Finding 2 (trailing newline)**: fixed with the same regex change, verified `fullmatch("my-secret\n")` is `False` while `fullmatch("tls.example.com")` is `True`.
- **Finding 3 (token burned on rejection)**: moved validation into a `callback=` on `--name`, following the existing `validate_non_empty` idiom in this file. It now runs at argument-parse time, before `Init()` is ever constructed, so a bad name never touches `Init.__init__`'s token redemption. Added a test that patches `Init.__init__` and asserts it's never called for an invalid name.
- **Adjacent (YAML quoting)**: added a `QuotedStr` wrapper + a `SafeDumper` subclass that quotes only `name`/`namespace`. Confirmed plain `yaml.safe_dump` leaves `y`/`1e5` bare today (the actual bug), and that the fix forces `name: 'y'`/`namespace: '1e5'` in the raw output.
- **Test nits**: hoisted `subprocess.run` into `_patched_cli` with `mock_run.assert_not_called()` on the dash-rejection test, wrapped the six-case loop in `self.subTest`, and slimmed the dead `KeeperCli.get_client`/unused response-queue scaffolding (traced the actual call path: nothing in the `init k8s` command touches those).

Left namespace validation and the `--apply` branch gaps (ignored `--namespace`/`--immutable`, no kubectl error handling) out of this PR per your note that those are tracked separately as KSM-1189.

Full `init_test.py` suite is green (10 passed, 7 subtests), and the broader CLI suite (198 passed) shows no regressions.

mgallego-keeper (approved):
Re-reviewed `fa6a7ab0` in full: all four findings from my review and follow-up comment are properly fixed, and I verified them independently rather than by reading the diff alone (regex equivalence against the apimachinery subdomain rule across a hostile case table including the 253/254 boundary and trailing-newline traps; parse-time rejection firing before `Init()` is ever constructed; the quoted manifest passing the kubectl decode stage that the bare one failed; `QuotingSafeDumper` isolation, so base `SafeDumper` call sites elsewhere are unaffected; and the newline-namespace payload still round-tripping lossless under the new emitter). Test nits all addressed. Approving.

Non-blocking notes for a follow-up commit; none of these need to hold the PR:

1. `_patched_cli` no longer patches `KeeperCli.get_client` (the old inline tests did). On a machine where profile discovery succeeds (a developer keyring profile from `ksm profile init` exists), the cli group callback builds a real client from stored credentials inside these tests, so the suite is machine-dependent: a stored profile that loads but fails client construction turns the exit-code-0 assertions red locally while CI (no profile) stays green. Re-adding that one patch line to `_patched_cli` restores hermeticity, and `test_k8s_invalid_name_does_not_redeem_token` could take it as well.
2. The `QuotedStr` protection is opt-in per value: a future manifest field added without the wrapper reintroduces the YAML 1.1 misparse silently (and PyYAML masks the gap by auto-quoting some ambiguous values while leaving exactly `y`/`n`/`1e5` bare). Since `QuotingSafeDumper` exists solely for this manifest, registering the quoting representer for plain `str` and dropping the wrapper makes quoting the default; that is verified lossless for the current manifest, including the raw-text test assertions. If it stays as is, avoid `default_style`, which would break the `immutable` bool.
3. Naming: nothing in `is_valid_k8s_name`/`K8S_NAME_ERROR` says this is the Secret-name (RFC 1123 subdomain) rule, and namespaces follow the stricter label rule (63 chars, no dots). A `_secret_name` rename or a contrastive comment would keep the KSM-1189 namespace-validation work from reusing it wholesale, where the "253 characters total" message would be wrong.
4. Smaller: `test_k8s` could restore the old appKey equality assertion by having `_patched_cli` also yield its `mock_config`; the three positive `--apply` tests could collapse into one table-driven `subTest` applying the full argv assertions to every name; and `is_valid_k8s_name` is not total (`None` raises TypeError), so an `isinstance(name, str)` prefix would harden the defensive `get_k8s` guard for programmatic callers.

The argv secret exposure the deep pass re-flagged (full config in `--from-literal=` visible via `/proc/<pid>/cmdline`) is already tracked as defect 4 of KSM-1189, where the stdin-pipe fix also covers the deferred `--namespace`/`--immutable` gaps.

PR #1108, opened 2026-08-13
Ruby SDK v17.2.1

stas-schaller:
## Summary

Ruby SDK v17.2.1: a single security fix in \`create_secret\`. This is a patch release and carries nothing else.

### Security Fixes
- **Record key encryption on create** (KSM-1193): \`create_secret\` now correctly encrypts the record key to the application owner's public key before including it in the create payload, matching the behavior of all other SDKs.

### Maintenance
- Version set to 17.2.1 in \`version.rb\`, \`CHANGELOG.md\`, and the README change log.

## Security Impact

\`create_secret\` was omitting the record key wrap step. All records created via the Ruby SDK through v17.2.0 are affected. Upgrade to 17.2.1 to protect newly created records. Existing records are not remediable through the SDK.

## Breaking Changes

\`create_secret\` now raises when the application config carries no \`appOwnerPublicKey\`, where it previously continued silently. This brings Ruby in line with all other SDKs.

## Related Issues

- Closes #1108
- Jira: KSM-1193

### DIFF
--- a/sdk/ruby/CHANGELOG.md
+++ b/sdk/ruby/CHANGELOG.md
@@ -2,8 +2,8 @@
 
 ## [17.2.1]
 
-### Fixed
-- **KSM-1193**: `create_secret` now encrypts the record key to the application owner's public key before sending it, matching the other SDKs. Previously the record key was placed in the create payload unencrypted, while the same request carried the record data encrypted under that key. All earlier published versions are affected: 17.0.3, 17.0.4, 17.1.0 and 17.2.0. A record key cannot be changed after creation, so upgrading protects newly created records only; existing records must be re-created to place them under a new key. `create_secret` now also raises an error when the application configuration has no owner public key, instead of continuing without the encryption step.
+### Security
+- **KSM-1193**: `create_secret` now encrypts the record key to the application owner's public key before sending it, matching the other SDKs. Previously the record key was placed in the create payload without that wrap, while the same request carried the record data encrypted under it. All earlier published versions are affected: 17.0.3, 17.0.4, 17.1.0 and 17.2.0; 17.0.0-17.0.2 were never published to RubyGems. A record key cannot be changed after creation, so upgrading protects newly created records only; records created by an earlier version must be re-created to place them under a new key. `create_secret` now also raises an error when the application configuration has no owner public key, instead of continuing without the encryption step.
 
 ## [17.2.0]
 
@@ -80,7 +80,7 @@
   - Removed emojis from all example files for professional appearance
 - **Dependencies:** Added base32 gem to test dependencies for TOTP support
 
-## [17.1.0] - 2025-01-06
+## [17.1.0]
 
 ### Changed
 - **BREAKING**: Minimum Ruby version increased to 3.1.0 (from 2.6.0)
@@ -99,28 +99,28 @@
 - `create_folder` now properly encrypts folder key with AES-CBC and sets correct parent_uid (nil for root-level folders)
 - Fixed AES-CBC encryption to not double-pad data (OpenSSL handles padding automatically)
 
-## [17.0.4] - 2025-10-20
+## [17.0.4]
 
 ### Changed
 - Maintenance release with internal improvements
 
-## [17.0.3] - 2025-06-25
+## [17.0.3]
 
 ### Changed
 - Cleaned up directory structure
 - Removed development and debug files from distribution
 
-## [17.0.2] - 2025-06-25
+## [17.0.2]
 
 ### Security
 - Updated all examples to use environment variables or placeholders
 
-## [17.0.1] - 2025-06-25
+## [17.0.1]
 
 ### Fixed
 - Added missing files to gem package (folder_manager, notation_enhancements, totp)
 
-## [17.0.0] - 2025-06-25
+## [17.0.0]
 
 ### Added
 - Initial release of Keeper Secrets Manager Ruby SDK
@@ -135,4 +135,5 @@
 - Version 17.0.0 to align with other Keeper SDKs
 - No runtime dependencies (base32 is optional)
 
+[17.2.1]: https://github.com/Keeper-Security/secrets-manager/compare/ruby-sdk-v17.2.0...ruby-sdk-v17.2.1
 [17.2.0]: https://github.com/Keeper-Security/secrets-manager/compare/ruby-sdk-v17.1.0...ruby-sdk-v17.2.0
--- a/sdk/ruby/CHANGELOG.md
+++ b/sdk/ruby/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [17.3.0]
 
 ### Fixed
-- **KSM-1193**: `create_secret` now encrypts the record key to the application owner's public key before sending it, matching the other SDKs. Previously the record key was placed in the create payload unencrypted, while the same request carried the record data encrypted under that key. Records created by versions 17.1.0 and 17.2.0 are affected. A record key cannot be changed after creation, so upgrading protects newly created records only; existing records must be re-created to place them under a new key. `create_secret` now also raises an error when the application configuration has no owner public key, instead of continuing without the encryption step.
+- **KSM-1193**: `create_secret` now encrypts the record key to the application owner's public key before sending it, matching the other SDKs. Previously the record key was placed in the create payload unencrypted, while the same request carried the record data encrypted under that key. All earlier published versions are affected: 17.0.3, 17.0.4, 17.1.0 and 17.2.0. A record key cannot be changed after creation, so upgrading protects newly created records only; existing records must be re-created to place them under a new key. `create_secret` now also raises an error when the application configuration has no owner public key, instead of continuing without the encryption step.
 
 ## [17.2.0]
 
--- a/sdk/ruby/CHANGELOG.md
+++ b/sdk/ruby/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [17.3.0]
+## [17.2.1]
 
 ### Fixed
 - **KSM-1193**: `create_secret` now encrypts the record key to the application owner's public key before sending it, matching the other SDKs. Previously the record key was placed in the create payload unencrypted, while the same request carried the record data encrypted under that key. All earlier published versions are affected: 17.0.3, 17.0.4, 17.1.0 and 17.2.0. A record key cannot be changed after creation, so upgrading protects newly created records only; existing records must be re-created to place them under a new key. `create_secret` now also raises an error when the application configuration has no owner public key, instead of continuing without the encryption step.
--- a/sdk/ruby/CHANGELOG.md
+++ b/sdk/ruby/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [17.3.0]
+
+### Fixed
+- **KSM-1193**: `create_secret` now encrypts the record key to the application owner's public key before sending it, matching the other SDKs. Previously the record key was placed in the create payload unencrypted, while the same request carried the record data encrypted under that key. Records created by versions 17.1.0 and 17.2.0 are affected. A record key cannot be changed after creation, so upgrading protects newly created records only; existing records must be re-created to place them under a new key. `create_secret` now also raises an error when the application configuration has no owner public key, instead of continuing without the encryption step.
+
 ## [17.2.0]
 
 ### Fixed
--- a/sdk/ruby/README.md
+++ b/sdk/ruby/README.md
@@ -5,9 +5,9 @@ For more information see our official documentation page https://docs.keeper.io/
 # Change Log
 
 ## 17.2.1
-- KSM-1193 - Security fix: `create_secret` now encrypts the record key to the application owner's public key before sending it, matching the other SDKs. Earlier versions placed the record key in the create payload unencrypted, while the same request carried the record data encrypted under that key. All previously published versions are affected (17.0.3, 17.0.4, 17.1.0, 17.2.0). A record key cannot be changed after creation, so upgrading protects newly created records only; records created by an earlier version must be re-created to place them under a new key.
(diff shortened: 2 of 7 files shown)


craiglurey (dismissed):

sdk/ruby/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## [17.2.1]
+
+### Fixed
```
mgallego-keeper: Three advisory-surface nits on this entry, all cheap: (1) this file's precedent for security items is a `### Security` section (used further down); under `### Fixed`, people scanning for security notices will miss the one entry they most need to see. (2) The bracketed heading has no matching link definition; the footer defines only `[17.2.0]`, so `[17.2.1]` renders as literal brackets and the compare-link trail dead-ends at this release. Adding `[17.2.1]: https://github.com/Keeper-Security/secrets-manager/compare/ruby-sdk-v17.2.0...ruby-sdk-v17.2.1` works once the tag exists. (3) Neither this entry nor the README one carries a date, and the publish date is the first thing an incident responder reconstructing a timeline needs.

sdk/ruby/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
...
+- **KSM-1193**: `create_secret` now encrypts the record key to the application owner's public key before sending it, matching the other SDKs. Previously the record key was placed in the create payload unencrypted, while the same request carried the record data encrypted under that key. All earlier published versions are affected: 17.0.3, 17.0.4, 17.1.0 and 17.2.0. A record key cannot be changed after creation, so upgrading protects newly created records only; existing records must be re-created to place them under a new key. `create_secret` now also raises an error when the application configuration has no owner public key, instead of continuing without the encryption step.
```
mgallego-keeper: Two precision tweaks on the wording: "All earlier published versions are affected: 17.0.3, 17.0.4, 17.1.0 and 17.2.0" implicitly clears 17.0.0-17.0.2, which this same file documents as releases. They were never published to RubyGems (the API 404s for all three), and one clause saying exactly that protects anyone running a vendored or git-installed 17.0.x from reading this list as clearing their version. Separately, "existing records must be re-created" reads as all records; the README's parallel sentence scopes it correctly to records created by an earlier version, worth mirroring here.

sdk/ruby/README.md
```diff
@@ -4,6 +4,9 @@ For more information see our official documentation page https://docs.keeper.io/
 
 # Change Log
 
+## 17.2.1
+- KSM-1193 - Security fix: `create_secret` now encrypts the record key to the application owner's public key before sending it, matching the other SDKs. Earlier versions placed the record key in the create payload unencrypted, while the same request carried the record data encrypted under that key. All previously published versions are affected (17.0.3, 17.0.4, 17.1.0, 17.2.0). A record key cannot be changed after creation, so upgrading protects newly created records only; records created by an earlier version must be re-created to place them under a new key.
```
mgallego-keeper: Two things worth adding here, since the README ships in the gem: (1) the one user-visible behavior change, `create_secret` now raising when the configuration has no `appOwnerPublicKey`, is in the CHANGELOG entry but not here. (2) Neither doc names the remedy for a user who hits that error: `appOwnerPublicKey` is only persisted during binding (core.rb:951 and 985), so an already-bound config can never acquire it, and the actual fix is re-binding with a fresh one-time token (or exporting `KSM_APPOWNERPUBLICKEY` when using the read-only EnvironmentStorage). The error message itself matches Python byte-for-byte, so rewording it is a cross-SDK conversation, but the README can state the remedy today.

sdk/ruby/lib/keeper_secrets_manager/core.rb
```diff
@@ -1245,11 +1245,20 @@ def prepare_get_payload(query_options = nil)
 
       # Prepare create payload
       def prepare_create_payload(record_uid:, record_key:, folder_uid:, folder_key:, data:, subfolder_uid: nil)
+        owner_public_key = @config.get_string(ConfigKeys::KEY_OWNER_PUBLIC_KEY)
+        unless owner_public_key && !owner_public_key.empty?
```
mgallego-keeper, line 1249: Non-blocking, and fine to defer to KSM-1226: a present-but-malformed value escapes this guard with raw low-level errors instead of this friendly one. Three shapes, all reproduced: whitespace or truncated base64 raises bare `ArgumentError` from `Utils.url_safe_str_to_bytes` (which, unlike `base64_to_bytes`, has no rescue); 65 well-formed-base64 random bytes pass decode and die in `Crypto.encrypt_ec` with `OpenSSL::PKey::EC::Point::Error`; and a JSON number value raises `NoMethodError` inside the guard itself, since `FileStorage#get_string` does no `to_s`. Wrapping the decode and the wrap in a rescue that re-raises this same `Error` (or the currently unused `ConfigurationError`) closes all three.

sdk/ruby/spec/keeper_secrets_manager/unit/create_secret_record_key_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+      :prepare_create_payload,
+      record_uid: 'test_record_uid',
+      record_key: record_key,
+      folder_uid: 'test_folder_uid',
+      folder_key: folder_key,
+      data: encrypted_data
+    )
+  end
+
+  describe '#prepare_create_payload record key handling' do
```
mgallego-keeper, line 39: The control assertions in these specs are exactly right. Two cheap strengthenings while the file is fresh, both spec-only and zero risk to the release artifact: (1) assert the serialized wire field, e.g. `JSON.parse(payload.to_json)['recordKey']`, not only the DTO accessor, so a `BasePayload#to_h` / `snake_to_camel` regression cannot send a create request with no `recordKey` while all three examples stay green; (2) add an empty-string owner key example: a JSON `null` becomes `''` through `InMemoryStorage`, and the guard here already handles it, but only the deleted-key branch is pinned. Related nit in the sibling file: `core_spec.rb:58`'s `bound_config` still stores `SecureRandom.random_bytes(65)` under `KEY_OWNER_PUBLIC_KEY`, the exact trap the new comment at `core_spec.rb:15` warns about; the same one-line fix applies. The longer tail of test-robustness items is listed in KSM-1226.

mgallego-keeper (dismissed):
Approving. I verified the fix independently end to end; details below. The asks in the inline comments are non-blocking, but the docs ones only count if they land before the gem publish dispatch, since README, CHANGELOG, and gemspec metadata are baked into the 17.2.1 package.

**Verification performed**

- Full suite locally on Ruby 3.3.8: 783 examples, 0 failures, 13 pre-existing pending; the `KEEPER_MOCK_MODE` offline integration test is green.
- All three new specs fail on the unfixed code for the advertised reasons (raw key on the wire, EC point parse failure on decrypt, no raise). `core_spec.rb` alone passes 77/77 in both fixed and unfixed states, which confirms the description's explanation of why the old suite never caught this.
- Cross-language wire check: bytes from Ruby `Crypto.encrypt_ec` decrypt with the Python SDK's `decrypt_ec` recipe; identical ECIES construction (ephemeral P-256, ECDH x-coordinate, SHA-256 KDF, AES-256-GCM), 125 bytes with leading `0x04` for a 32-byte key, matching the live QA observation in the description.
- Owner-key decode verified against all four stored formats (url-safe or standard alphabet, padded or unpadded), including keys whose base64 contains `+` or `/`, so Commander-written configs, Web Vault configs, and the SDK's own binding format all work.
- All six other SDKs hard-fail without `appOwnerPublicKey` as claimed; the new message is byte-identical to Python's current one.
- `client_version` derives from `VERSION`, so fixed clients stamp `mb17.2.1`, cleanly separable from `mr17.0.0` / `mr17.91.0` / `mb17.2.0` for server-side forensics.
- The branch carries exactly these six files over master (ahead 3, behind 0), and the affected-version list matches RubyGems (17.0.0-17.0.2 were never published).
- Payload sweep found no sibling issue: every other key the SDK transmits is wrapped (folderKey, file keys, shared-folder traversal), and `UpdatePayload` has no key field, confirming the cannot-rotate statement.

**One ask that cannot be an inline comment (file not in the diff):** `keeper_secrets_manager.gemspec:18` sets `changelog_uri` to a repo-root `CHANGELOG.md` that does not exist, so the Changelog link on the 17.2.1 rubygems.org page will 404 at exactly the moment users come looking for this advisory. One-line change to `.../blob/master/sdk/ruby/CHANGELOG.md`; worth including in this release since the metadata ships in the gem.

**PR description nit:** "Breaking Changes: None for any configuration that can create records today" contradicts the sentence that follows it; a configuration without the owner key can create records today (that is the vulnerability) and will raise after upgrading. "Any correctly provisioned configuration" is probably the intended meaning.

**Optional, while both docs files are open:** the neighboring dates look wrong: 17.1.0 is dated 2025-01-06 in both files while the gem published on 2025-11-06 (looks like a 01/11 transposition), and the README's 17.2.0 entry says 2025-11-14 while the gem published on 2026-07-20. If 17.2.1 gets a date, best not to propagate the pattern.

Pre-existing items found during the review are filed separately so nothing extra lands on this security patch: KSM-1223 (undefined `KEY_ENCRYPTED_APP_KEY` raises NameError on the no-appKey path), KSM-1224 (`get_file_data` output incompatible with `download_file`), KSM-1225 (redact key material from debug logging), KSM-1226 (owner-key handling unification and payload-builder hardening), KSM-1227 (remove the silent AES-GCM to CBC fallbacks), KSM-1228 (0600 on `ksm_cache.bin`). The application-log aspect of the exposure is noted on KSM-1193.

sdk/ruby/CHANGELOG.md
stas-schaller: Done — moved under `### Security` to match the 17.0.0/17.0.2 precedent in this same file, and added the missing `[17.2.1]` compare link in the footer.

sdk/ruby/CHANGELOG.md
stas-schaller: Done — added the "17.0.0-17.0.2 were never published to RubyGems" clause, and scoped "existing records" to "records created by an earlier version" to match the README wording.

sdk/ruby/README.md
stas-schaller: Done — added a sentence covering the new raise on missing `appOwnerPublicKey` and the remedy (re-bind with a fresh one-time token, or set `KSM_APPOWNERPUBLICKEY` when using the read-only `EnvironmentStorage`).

sdk/ruby/lib/keeper_secrets_manager/core.rb
stas-schaller: Agreed this is non-blocking — leaving as-is here. Tracking with the rest of the owner-key hardening pass you scoped (malformed-value guard, unification) rather than expanding this security patch.

sdk/ruby/spec/keeper_secrets_manager/unit/create_secret_record_key_spec.rb
stas-schaller: Agreed on both — not touching test assertions on this security patch. Tracking the serialized-wire assertion, the empty-string case, and the `core_spec.rb:58` random-bytes owner key together for the follow-up hardening pass.

stas-schaller:
Thank ya, here are my additions:

- `keeper_secrets_manager.gemspec:18` — `changelog_uri` now points at `sdk/ruby/CHANGELOG.md` (the blocking one ahead of gem publish)
- CHANGELOG: KSM-1193 entry moved under `### Security`, added the 17.0.0-17.0.2-never-published clause, scoped "existing records" to "records created by an earlier version," added the missing `[17.2.1]` compare link
- README: documented the `create_secret` raise on missing `appOwnerPublicKey` and the remedy (re-bind, or `KSM_APPOWNERPUBLICKEY` for `EnvironmentStorage`)
- PR description: "None for any configuration that can create records today" → "None for any correctly provisioned configuration"
- Dates: went further than the transposition fix and dropped dates from every version header in both files (`CHANGELOG.md` and the README change log) — publish date isn't known when the branch is prepared, so we don't carry it on the header at all going forward

mgallego-keeper (dismissed):
Re-reviewed at 85252966; everything verified on the new head.

- CHANGELOG: entry now under `### Security` (matching this file's own 17.0.2 precedent), `[17.2.1]` compare link defined, and both wording fixes are in (never-published clause, scoped re-creation sentence).
- README: documents the new raise and the actual remedy (re-bind with a fresh one-time token, or `KSM_APPOWNERPUBLICKEY` on the read-only `EnvironmentStorage`).
- Gemspec: `changelog_uri` fixed. I built the gem from this head and the packaged metadata carries the corrected link (`keeper_secrets_manager-17.2.1.gem` -> `blob/master/sdk/ruby/CHANGELOG.md`).
- Dates: removing them from all version headers instead of adding one is a sensible resolution; the provably wrong dates are gone and rubygems.org remains the single source of truth for publish timestamps.
- Deferrals (guard error contract, spec strengthenings) match what KSM-1226 already tracks; agreed they do not belong on this security patch.
- Re-verified locally on 85252966: 783 examples, 0 failures, 13 pre-existing pending; offline mock test green; `gem build` clean at 17.2.1.

Resolving the five threads. Still approved; good to publish and merge.

mgallego-keeper (approved):

mgallego-keeper (approved):
Re-approved at dd78c5ad. The branch was force-pushed over my previous approval with one amendment: two lines of advisory wording in CHANGELOG.md and README.md ("placed the record key in the create payload unencrypted" becomes "without that wrap"), which is the more precise phrasing given the encrypted transport envelope. Verified the full tree delta from the previously approved 85252966 is exactly those two lines and nothing else, and that the published gem (17.2.1, sha256 ca73f928d6d51e421394ecfb1ca03181e5a3a4c8823157a906494b0b0206a7a7) is byte-identical to this head: lib/ and the gemspec match the reviewed code, the docs match the amendment. Cumulative diff vs master remains the same seven files. Still approved.

PR #1109, opened 2026-08-17
fix(java): validate server key_id range and eliminate config write race window

stas-schaller:
## Summary

Java SDK: validates and bounds server-supplied key ID rotation in `postQuery`, and eliminates the config/cache file read window on write.

### Fixed

- **Server key_id validation and rotation cap** (KSM-1248): the SDK accepted any `key_id` from a server error response and wrote it directly to storage without checking whether the key existed, and retried with no bound. An unrecognized key ID corrupted the stored configuration and made every subsequent call fail. A valid key ID that the server kept re-suggesting caused an infinite loop of full encrypt/sign/POST cycles. The SDK now (1) rejects any key ID not in the embedded public key table, throwing `SecretsManagerException` naming the suggested ID and the supported range, and (2) caps key rotation retries at `MAX_KEY_ROTATION_RETRIES` (3), throwing a typed terminal error naming the key ID that was not accepted.
- **Config and cache file write race window** (KSM-1262): on POSIX systems, config and cache files are now written via a temp-file swap with 0600 permissions set before data is written, closing the window where other local users could read the file during a write. Two behavior changes: (1) a symlinked config path is replaced by a regular file on the first write; (2) a config file in a directory without write permission will fail at temp-file creation — move the config to a writable directory or use `InMemoryStorage` with an injected config string instead.

## Testing

```bash
cd sdk/java/core && ./gradlew test
```

`testServerKeyIdOutOfRangeIsRejected` confirms that a server-supplied `key_id` of 999 throws `SecretsManagerException` naming the suggested ID and the supported range, and that `KEY_SERVER_PUBLIC_KEY_ID` is not written to storage.
`testCustomServerKeyGuardTakesPriorityOverKeyTable` confirms that when a custom server key is configured, a server key-rotation hint throws the custom-key error before the key-table guard — pinning the branch boundary.
`testServerKeyRotationRetryCap` confirms that a server that keeps returning an in-table key ID causes exactly `MAX_KEY_ROTATION_RETRIES + 1` calls and then throws a terminal `SecretsManagerException` naming the retry limit.
`testConfigFileAtomicWriteIsOwnerOnly` confirms that the config file is written with `rw-------` POSIX permissions.

## Security Impact

KSM-1248 prevents a hostile server response from either corrupting the stored key ID or pinning the client in an unbounded retry loop.

KSM-1262 eliminates the window where other local users can read the config or cache file on POSIX systems. The config file holds the app key, which decrypts all record field values.

## Breaking Changes

None.

## Related Issues

- Jira: KSM-1248, KSM-1262

### DIFF
--- a/sdk/java/core/README.md
+++ b/sdk/java/core/README.md
@@ -7,6 +7,8 @@ For more information see our official documentation page https://docs.keeper.io/
 ## 17.4.0
 - KSM-1081 - Fixed `getFolders()` crashing when any folder in the response has a corrupted or missing key. The SDK now skips undecryptable folders and returns the remaining folders normally.
 - KSM-1086 - Fixed `deleteFolder()` to return `SecretsManagerDeleteFolderResponse` (typed per-folder status), matching `deleteSecret()`. The SDK now logs per-item server failures to stderr and includes them in the return value so callers can detect partial failures.
+- KSM-1248 - Server-supplied key IDs are validated against the embedded public key table before being stored. An unrecognized key ID throws `SecretsManagerException` and leaves storage unchanged. Key rotation retries are now capped at `MAX_KEY_ROTATION_RETRIES` (3); exhausted retries throw a typed error naming the last suggested key ID.
+- KSM-1262 - On POSIX systems, config and cache files are now written via a temp-file swap with 0600 permissions set before data is written, closing the window where other local users could read the file during a write. Two behavior changes from the new approach: (1) a symlinked config path is replaced by a regular file on the first write; (2) a config file in a directory without write permission (for example, a read-only container volume mount) will fail at temp-file creation — move the config to a writable directory or use `InMemoryStorage` with an injected config string instead.
 
 ## 17.3.0
 **Breaking Changes**
--- a/sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/LocalConfigStorage.kt
+++ b/sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/LocalConfigStorage.kt
@@ -6,33 +6,42 @@ import kotlinx.serialization.decodeFromString
 import kotlinx.serialization.encodeToString
 import kotlinx.serialization.json.Json
 import java.io.*
+import java.nio.file.AtomicMoveNotSupportedException
 import java.nio.file.Files
-import java.nio.file.attribute.PosixFilePermission
+import java.nio.file.StandardCopyOption
 import java.nio.file.attribute.PosixFilePermissions
 import java.util.*
 import kotlin.collections.HashMap
 
 fun saveCachedValue(data: ByteArray) {
-    val file = File("cache.dat")
-    FileOutputStream(file).use { fos -> fos.write(data) } // .use{} closes on exception even if the operation throws
-
-    // Set file permissions to 0600 (owner read/write only)
+    val targetPath = File("cache.dat").absoluteFile.toPath()
+    val tmpPath = Files.createTempFile(targetPath.parent, "ksm_", ".tmp")
     try {
-        val perms = PosixFilePermissions.fromString("rw-------")
-        Files.setPosixFilePermissions(file.toPath(), perms)
-    } catch (e: UnsupportedOperationException) {
-        // Windows or file system doesn't support POSIX permissions
-        // File.setReadable/setWritable provides basic protection
-        file.setReadable(false, false)  // Remove all read permissions
-        file.setWritable(false, false)  // Remove all write permissions
-        file.setReadable(true, true)     // Owner read only
-        file.setWritable(true, true)     // Owner write only
+        try {
+            Files.setPosixFilePermissions(tmpPath, PosixFilePermissions.fromString("rw-------"))
+        } catch (_: UnsupportedOperationException) {
+            tmpPath.toFile().let { f ->
+                f.setReadable(false, false)
+                f.setWritable(false, false)
+                f.setReadable(true, true)
+                f.setWritable(true, true)
+            }
+        }
+        Files.write(tmpPath, data)
+        try {
+            Files.move(tmpPath, targetPath, StandardCopyOption.ATOMIC_MOVE)
+        } catch (_: AtomicMoveNotSupportedException) {
+            Files.move(tmpPath, targetPath, StandardCopyOption.REPLACE_EXISTING)
+        }
+    } catch (e: Exception) {
+        try { Files.deleteIfExists(tmpPath) } catch (_: Exception) { }
+        throw e
     }
 }
 
 fun getCachedValue(): ByteArray {
     try {
-        return FileInputStream("cache.dat").use { it.readBytes() } // .use{} closes on exception even if the operation throws
+        return FileInputStream("cache.dat").use { it.readBytes() } // .use{} closes on exception
     } catch (e: Exception) {
         throw SecretsManagerException("Cached value does not exist")
     }
@@ -115,7 +124,7 @@ class LocalConfigStorage(configName: String? = null) : KeyValueStorage {
 
     private val file = configName?.let { File(it) }
     private var storage: InMemoryStorage = if (file != null && file.exists()) {
-        val content = BufferedReader(FileReader(file)).use { it.readText() } // previously never closed on exception; .use{} guarantees it
+        val content = file.readText(Charsets.UTF_8)
         InMemoryStorage(content)
     } else {
         InMemoryStorage()
@@ -125,29 +134,38 @@ class LocalConfigStorage(configName: String? = null) : KeyValueStorage {
 
     private fun saveToFile() {
         if (file == null) return
-        val config = LocalConfig()
-        config.hostname = storage.getString(KEY_HOSTNAME)
-        config.clientId = storage.getString(KEY_CLIENT_ID)
-        config.privateKey = storage.getString(KEY_PRIVATE_KEY)
-        config.clientKey = storage.getString(KEY_CLIENT_KEY)
-        config.appKey = storage.getString(KEY_APP_KEY)
-        config.appOwnerPublicKey = storage.getString(KEY_OWNER_PUBLIC_KEY)
-        config.serverPublicKeyId = storage.getString(KEY_SERVER_PUBLIC_KEY_ID)
-        config.serverPublicKey = storage.getString(KEY_SERVER_PUBLIC_KEY)
-        val json = prettyJson.encodeToString(config)
-        BufferedWriter(FileWriter(file)).use { it.write(json) } // .use{} closes on exception even if the operation throws
-
-        // Set file permissions to 0600 (owner read/write only)
+        val targetPath = file.absoluteFile.toPath()
+        val tmpPath = Files.createTempFile(targetPath.parent, "ksm_", ".tmp")
         try {
-            val perms = PosixFilePermissions.fromString("rw-------")
-            Files.setPosixFilePermissions(file.toPath(), perms)
-        } catch (e: UnsupportedOperationException) {
-            // Windows or file system doesn't support POSIX permissions
-            // File.setReadable/setWritable provides basic protection
-            file.setReadable(false, false)  // Remove all read permissions
-            file.setWritable(false, false)  // Remove all write permissions
-            file.setReadable(true, true)     // Owner read only
-            file.setWritable(true, true)     // Owner write only
+            try {
+                Files.setPosixFilePermissions(tmpPath, PosixFilePermissions.fromString("rw-------"))
+            } catch (_: UnsupportedOperationException) {
+                tmpPath.toFile().let { f ->
+                    f.setReadable(false, false)
+                    f.setWritable(false, false)
+                    f.setReadable(true, true)
+                    f.setWritable(true, true)
+                }
+            }
+            val config = LocalConfig(
+                hostname = storage.getString(KEY_HOSTNAME),
+                clientId = storage.getString(KEY_CLIENT_ID),
+                privateKey = storage.getString(KEY_PRIVATE_KEY),
+                clientKey = storage.getString(KEY_CLIENT_KEY),
+                appKey = storage.getString(KEY_APP_KEY),
+                appOwnerPublicKey = storage.getString(KEY_OWNER_PUBLIC_KEY),
+                serverPublicKeyId = storage.getString(KEY_SERVER_PUBLIC_KEY_ID),
+                serverPublicKey = storage.getString(KEY_SERVER_PUBLIC_KEY)
+            )
+            Files.write(tmpPath, prettyJson.encodeToString(config).toByteArray())
+            try {
+                Files.move(tmpPath, targetPath, StandardCopyOption.ATOMIC_MOVE)
+            } catch (_: AtomicMoveNotSupportedException) {
+                Files.move(tmpPath, targetPath, StandardCopyOption.REPLACE_EXISTING)
+            }
+        } catch (e: Exception) {
+            try { Files.deleteIfExists(tmpPath) } catch (_: Exception) { }
+            throw e
         }
     }
 
--- a/sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
+++ b/sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
@@ -36,6 +36,7 @@ const val KEEPER_CLIENT_VERSION = "mj17.3.0"
 const val MAX_THROTTLE_RETRIES = 5
 const val BASE_THROTTLE_DELAY_SEC = 11 // 1s safety margin over the backend's 10s memcached TTL
 const val MAX_THROTTLE_DELAY_SEC = 176 // caps a backend-supplied retry_after from forcing an excessive wait
+const val MAX_KEY_ROTATION_RETRIES = 3 // parity with JS SDK open PR #1078
 
 const val KEY_HOSTNAME = "hostname" // base url for the Secrets Manager service
 const val KEY_SERVER_PUBLIC_KEY_ID = "serverPublicKeyId"
@@ -1806,6 +1807,7 @@ private inline fun <reified T> postQuery(
     val url = "https://${hostName}/api/rest/sm/v1/${path}"
     val throttleSleep = options.throttleSleepMillis ?: { ms -> Thread.sleep(ms) }
     var throttleAttempt = 0
+    var keyRotationAttempt = 0
     while (true) {
         val transmissionKey = generateTransmissionKey(options.storage)
         val encryptedPayload = encryptAndSignPayload(options.storage, transmissionKey, payload)
@@ -1846,6 +1848,18 @@ private inline fun <reified T> postQuery(
                     val currentKeyId = options.storage.getString(KEY_SERVER_PUBLIC_KEY_ID)
                     throw SecretsManagerException("Server rejected the custom server public key (id $currentKeyId). The server suggested key id ${error.key_id}. Please update your IL5 KSM configuration.")
                 }
+                if (!keeperPublicKeys.containsKey(error.key_id)) {
+                    val ids = keeperPublicKeys.keys.sorted()
+                    throw SecretsManagerException(
+                        "Server suggested unsupported key id ${error.key_id}; this SDK version supports key ids ${ids.first()}-${ids.last()}"
+                    )
+                }
+                if (keyRotationAttempt >= MAX_KEY_ROTATION_RETRIES) {
+                    throw SecretsManagerException(
+                        "Server key rotation exhausted $MAX_KEY_ROTATION_RETRIES retries; key id ${transmissionKey.publicKeyId} was not accepted"
+                    )
+                }
+                keyRotationAttempt++
                 options.storage.saveString(KEY_SERVER_PUBLIC_KEY_ID, error.key_id.toString())
                 continue
             }
--- a/sdk/java/core/src/test/kotlin/com/keepersecurity/secretsManager/core/SecretsManagerTest.kt
+++ b/sdk/java/core/src/test/kotlin/com/keepersecurity/secretsManager/core/SecretsManagerTest.kt
@@ -181,6 +181,94 @@ internal class SecretsManagerTest {
     }
 
     @Test
+    fun testServerKeyIdOutOfRangeIsRejected() {
+        // Server-supplied key_id outside the known range must be rejected before storage is written.
+        val storage = InMemoryStorage()
+        initializeStorage(storage, "fake.keepersecurity.com:FAKE_CLIENT_KEY")
+        TestStubs.transmissionKeyStub = { ByteArray(32) }
+        val testPostFunction: (String, TransmissionKey, EncryptedPayload) -> KeeperHttpResponse = { _, _, _ ->
+            KeeperHttpResponse(400, """{"error":"key","key_id":999}""".toByteArray())
+        }
+        val options = SecretsManagerOptions(storage, testPostFunction)
+        val ex = assertFailsWith<SecretsManagerException> {
+            getSecrets(options)
+        }
+        assertTrue(
+            ex.message?.contains("unsupported key id 999") == true,
+            "Exception must come from the key-table guard and name the rejected id. Got: ${ex.message}"
+        )
+        assertNull(
+            storage.getString(KEY_SERVER_PUBLIC_KEY_ID),
+            "Storage must not be poisoned with key_id=999. Got: ${storage.getString(KEY_SERVER_PUBLIC_KEY_ID)}"
+        )
+    }
+
+    @Test
+    fun testCustomServerKeyGuardTakesPriorityOverKeyTable() {
+        // When a custom server key is configured, a server key-rotation hint must reach
+        // the custom-key branch and throw the actionable error before the key-table guard.
+        val storage = InMemoryStorage()
+        initializeStorage(storage, "fake.keepersecurity.com:FAKE_CLIENT_KEY")
+        // Use a real EC public key so webSafe64ToBytes succeeds in generateTransmissionKey.
+        // The value here is keeper public key #7 (from the embedded table).
+        storage.saveString(KEY_SERVER_PUBLIC_KEY, "BK9w6TZFxE6nFNbMfIpULCup2a8xc6w2tUTABjxny7yFmxW0dAEojwC6j6zb5nTlmb1dAx8nwo3qF7RPYGmloRM")
+        storage.saveString(KEY_SERVER_PUBLIC_KEY_ID, "20")
+        TestStubs.transmissionKeyStub = { ByteArray(32) }
+        val testPostFunction: (String, TransmissionKey, EncryptedPayload) -> KeeperHttpResponse = { _, _, _ ->
+            KeeperHttpResponse(400, """{"error":"key","key_id":999}""".toByteArray())
+        }
+        val options = SecretsManagerOptions(storage, testPostFunction)
+        val ex = assertFailsWith<SecretsManagerException> {
+            getSecrets(options)
+        }
+        assertTrue(
+            ex.message?.contains("custom server public key") == true,
+            "Custom-key path must fire before the key-table guard. Got: ${ex.message}"
+        )
+    }
+
+    @Test
+    fun testServerKeyRotationRetryCap() {
+        // When the server keeps returning an in-table key_id, the loop must stop after MAX_KEY_ROTATION_RETRIES.
+        val storage = InMemoryStorage()
+        initializeStorage(storage, "fake.keepersecurity.com:FAKE_CLIENT_KEY")
+        TestStubs.transmissionKeyStub = { ByteArray(32) }
+        val callCount = intArrayOf(0)
+        val testPostFunction: (String, TransmissionKey, EncryptedPayload) -> KeeperHttpResponse = { _, _, _ ->
+            callCount[0]++
+            KeeperHttpResponse(400, """{"error":"key","key_id":10}""".toByteArray())
+        }
+        val options = SecretsManagerOptions(storage, testPostFunction)
+        val ex = assertFailsWith<SecretsManagerException> {
+            getSecrets(options)
+        }
+        assertTrue(
+            ex.message?.contains("key rotation exhausted $MAX_KEY_ROTATION_RETRIES retries") == true,
+            "Exception must name the retry limit. Got: ${ex.message}"
+        )
+        assertEquals(
+            MAX_KEY_ROTATION_RETRIES + 1, callCount[0],
+            "Must make exactly MAX_KEY_ROTATION_RETRIES+1 calls (initial + cap retries)"
+        )
+    }
+
+    @Test
+    fun testConfigFileAtomicWriteIsOwnerOnly() {
+        // KSM-1262: the config file must be written via atomic rename and be readable only by the owner.
+        val configFile = File.createTempFile("ksm-test-", ".json").also { it.delete() }
+        try {
+            val storage = LocalConfigStorage(configFile.absolutePath)
+            initializeStorage(storage, "fake.keepersecurity.com:FAKE_CLIENT_KEY")
+            assertTrue(configFile.exists(), "Config file must exist after initializeStorage")
+            val perms = java.nio.file.Files.getPosixFilePermissions(configFile.toPath())
+            val expected = java.nio.file.attribute.PosixFilePermissions.fromString("rw-------")
+            assertEquals(expected, perms, "Config file must be owner-read/write only. Got: $perms")
+        } finally {
+            configFile.delete()
+        }
+    }
+
+    @Test
     fun testRecordCreateEmptyCustomSerialized() {
         // RecordCreate with no custom fields must include "custom": [] in JSON payload
         val recordData = KeeperRecordData(
@@ -195,7 +283,7 @@ internal class SecretsManagerTest {
 
     @Test
     fun testKeeperFileDataMissingLastModified() {
-        // lastModified entirely absent; must deserialize without throwing
+        // lastModified is absent. The SDK must deserialize without throwing.
         val json = """{"title":"test.txt","name":"test.txt","type":"text/plain","size":1024}"""
         val result = Json.decodeFromString<KeeperFileData>(json)
         assertEquals(0L, result.lastModified)
@@ -212,7 +300,7 @@ internal class SecretsManagerTest {
 
     @Test
     fun testKeeperFileDataFractionalLastModified() {
-        // Regression guard: fractional lastModified (iOS client format)
+        // Regression guard for fractional lastModified (iOS client format)
         val json = """{"title":"test.txt","name":"test.txt","size":1024,"lastModified":1760646182.790214}"""
         val result = Json.decodeFromString<KeeperFileData>(json)
         assertEquals(1760646182L, result.lastModified)
@@ -220,7 +308,7 @@ internal class SecretsManagerTest {
 
     @Test
     fun testKeeperFileNullUrl() {
-        // KeeperFile.url must be nullable; server may omit url for files without a download link
+        // KeeperFile.url must be nullable. The server may omit url for files without a download link.
         val fileData = KeeperFileData("test.txt", "test.txt", "text/plain", 1024)
         val file = KeeperFile(ByteArray(32), "uid123", fileData, null, null)
         assertNull(file.url)
@@ -228,7 +316,7 @@ internal class SecretsManagerTest {
 
     @Test
     fun testBase64EmptyStringThrowsTypedException() {
-        // Empty string must throw a typed Keeper exception, not an NPE from inside java.util.Base64
+        // Empty string must throw a typed Keeper exception, not an NPE from inside java.util.Base64.
         val base64Ex = assertFailsWith<Exception> { base64ToBytes("") }
         assertTrue(base64Ex.message?.isNotEmpty() == true)
         val webSafe64Ex = assertFailsWith<Exception> { webSafe64ToBytes("") }


sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
```diff
@@ -1846,6 +1846,9 @@ private inline fun <reified T> postQuery(
                     val currentKeyId = options.storage.getString(KEY_SERVER_PUBLIC_KEY_ID)
                     throw SecretsManagerException("Server rejected the custom server public key (id $currentKeyId). The server suggested key id ${error.key_id}. Please update your IL5 KSM configuration.")
                 }
+                if (!keeperPublicKeys.containsKey(error.key_id)) {
```
mgallego-keeper, line 1851: **[blocking] This guard closes the config-poisoning half of KSM-1248 and leaves the unbounded loop, which is the half in the ticket title.**

The guard only fires for an id *outside* the bundled table. When the server answers with an id that is inside it (7 through 18), nothing has changed: the id is stored, `continue` runs, and `while (true)` goes around again with no counter, no backoff and no log line. A backend or intermediary that keeps answering `{"error":"key","key_id":10}` pins the client in a hot loop.

Probed at this head with a stub that always returns key id 10, bailing out at 50 to stop the test:

```
PROBE1 in-table key_id=10: calls=50 terminal=BailOut: bail after 50 calls
PROBE1 stored key id after loop: 10
```

Fifty full `generateTransmissionKey` + `encryptAndSignPayload` + POST cycles in well under a second, and it only stopped because I made it stop. Against a live backend that is a self-inflicted request flood as well as a hung client.

KSM-1248 asks for three things and this PR does the second:

1. bound rotation retries (JS uses 3), mirroring the in-file throttle pattern, ending in a typed terminal error naming the key id actually rejected
2. validate `error.key_id` membership before persisting it
3. consider failing fast when the server re-suggests the id that was just used

The pattern to copy is nine lines up in this same function: `throttleAttempt >= MAX_THROTTLE_RETRIES` at line 1826. The JS twin (#1078, KSM-1128) caps at 3 and throws a typed error, so capping here also keeps the two SDKs telling users the same story.

Roughly:

```kotlin
var keyRotationAttempt = 0
// ...
if (!keeperPublicKeys.containsKey(error.key_id)) {
    throw SecretsManagerException("Server returned an unrecognized key id ${error.key_id}")
}
if (keyRotationAttempt >= MAX_KEY_ROTATION_RETRIES) {
    throw SecretsManagerException(
        "Server keeps rejecting the server public key; exhausted $MAX_KEY_ROTATION_RETRIES rotation retries (last suggested key id ${error.key_id})"
    )
}
keyRotationAttempt++
options.storage.saveString(KEY_SERVER_PUBLIC_KEY_ID, error.key_id.toString())
continue
```

Item 3 is close to free once the counter exists: if `error.key_id.toString()` equals the id already in storage, the server is not making progress and one retry is enough to know it.

Two knock-on effects if this ships as-is. The `README.md` entry claims only the validation, so the release note would understate the fix, and KSM-1248 would be closed with the defect named in its own title still live.

Worth saying plainly: the guard itself is correct and I am not asking you to change it. This is about the loop it sits inside.

sdk/java/core/README.md
```diff
@@ -4,6 +4,10 @@ For more information see our official documentation page https://docs.keeper.io/
 
 # Change Log
 
+## 17.4.0
```
mgallego-keeper, line 7: **[blocking] This creates a second `## 17.4.0` heading and the PR is currently unmergeable because of it.**

The base branch already has a `## 17.4.0` section, added by the STE pass (`16dc9957`) earlier today, holding KSM-1081 and KSM-1086. This adds another one directly above it.

Confirmed rather than assumed:

```
$ git merge-tree --write-tree origin/release/sdk/java/core/v17.4.0 origin/pr/1109
Auto-merging sdk/java/core/README.md
CONFLICT (content): Merge conflict in sdk/java/core/README.md
```

That is the only conflicting file, so a rebase plus moving these two bullets into the existing `## 17.4.0` section clears it completely.

Same failure mode I flagged on #1078 today, and the cause is the same in both: the branch predates the release branch's own changelog section. Worth a glance at any other open branch cut before today.

Two notes on the wording once it moves:

- The KSM-1248 line describes only the validation. If the retry cap lands too (see `SecretsManager.kt:1849`), this needs to say so, since "can loop forever" is what users would actually notice.
- "eliminating the window where file contents were readable by other local users" is accurate and I verified it, but "atomic" here means atomic against concurrent readers, not durable across a crash. There is no `fsync` before the rename, so a power loss can still leave a short file. Not worth fixing, worth not implying.

sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
```diff
@@ -493,8 +494,9 @@ data class KeeperRecordLink(
     }
     
     /**
-     * Decrypt the link data using the provided record key.
-     *
+     * Decrypt the link data using the provided record key
+     * This method attempts decryption only if the data appears to be encrypted
+     *
```
mgallego-keeper: **[blocking] This hunk, and about a dozen like it, revert comment work that is already merged.**

This commit's parent is `8c028be6`, which *is* #1080 ("humanize comments in core SDK on the release branch"), so these are genuine reversions rather than base drift. The KDoc here goes from the merged one-line form back to the older "This method attempts decryption only if..." phrasing.

Measured across the diff:

```
SecretsManager.kt            code +5/-2      comment/blank +28/-28   new-trailing-ws: 6
Notation.kt                  code +0/-0      comment/blank +0/-1
RecordData.kt                code +1/-1
IL5EdgeCaseTest.kt           code +0/-0      comment/blank +1/-1
KeeperRecordLinkTest.kt      code +0/-0      comment/blank +3/-3
RecordDataPamFieldsTest.kt   code +0/-0      comment/blank +1/-1
TOTAL: code=117  comment/blank=87
```

So 87 of 204 changed lines are comment churn unrelated to either ticket. Sharper still, comparing each file against its pre-#1080 state:

```
RecordDataPamFieldsTest.kt: differing lines vs PRE-1080=0 ; vs base=2
KeeperRecordLinkTest.kt:    differing lines vs PRE-1080=2 ; vs base=6
```

`RecordDataPamFieldsTest.kt` is byte-identical to the file as it stood before #1080. That reads like a working copy that predates #1080 rather than an intentional revert, which is the easy explanation and the easy fix.

Specifically undone in this commit:

- KDoc on `hasEncryptedData`, `decryptData`, `getLinkData` and `getSettingsForPath` back to the longer pre-#1080 prose
- KDoc deleted outright from `parseJsonToMap` and `isPrintableText`
- `parseThrottle` and `throttleJitter` comments replaced with vaguer versions that drop the "why" (`throttleJitter` no longer explains why the range is one-sided)
- trailing whitespace back on six lines here, plus one whitespace-only line
- Jira and PR references back into KDoc that #1080 had stripped: "(KSM-992, PR #1036)" in `KeeperRecordLinkTest.kt:13` and "added in KSM-738 (VAUL-7662)" in `RecordDataPamFieldsTest.kt:10`

A rebase onto current `release/sdk/java/core/v17.4.0`, keeping only the `LocalConfigStorage.kt`, `SecretsManager.kt:1849`, `SecretsManagerTest.kt` and `README.md` hunks, fixes this and the README conflict together. It also shrinks the release diff to something a reader can check: about 80 lines of real change instead of 204.

sdk/java/core/src/test/kotlin/com/keepersecurity/secretsManager/core/SecretsManagerTest.kt
```diff
@@ -180,6 +180,29 @@ internal class SecretsManagerTest {
...
+        TestStubs.transmissionKeyStub = { ByteArray(32) }
+        val testPostFunction: (String, TransmissionKey, EncryptedPayload) -> KeeperHttpResponse = { _, _, _ ->
+            KeeperHttpResponse(400, """{"error":"key","key_id":999}""".toByteArray())
+        }
+        val options = SecretsManagerOptions(storage, testPostFunction)
+        val ex = assertFailsWith<SecretsManagerException> {
+            getSecrets(options)
+        }
+        assertTrue(
+            ex.message?.contains("999") == true,
```
mgallego-keeper: The test is the right test to add, but this assertion cannot tell which branch threw.

The pinned-custom-key branch twelve lines above the new guard puts the suggested id in its message too, so `contains("999")` matches both:

```
PROBE3 custom-key branch msg=Server rejected the custom server public key (id 20). The server suggested key id 999. Please update your IL5 KSM configuration.
PROBE3 contains 999? true
```

With this storage the custom-key branch is not configured, so today it does exercise the new guard. It just would not notice if that stopped being true. Asserting on the distinguishing phrase pins it:

```kotlin
assertTrue(
    ex.message?.contains("unrecognized key id 999") == true,
    "Exception must come from the key-table guard and name the rejected id. Got: ${ex.message}"
)
```

`assertNull` on the stored id is the right check and it does discriminate: `initializeStorage` only writes `KEY_SERVER_PUBLIC_KEY_ID` on the 4-segment extended token path (line 818), so with this 2-segment token it starts null and would be non-null if the guard were removed. Verified.

Two coverage gaps while you are here. Nothing covers the in-table path, which is where the unbounded loop lives, so a retry cap would land untested; a stub that always returns key id 10 with a call counter pins it in a few lines. And nothing covers KSM-1262 at all: `PROBE4` above is four lines and asserts the thing the ticket is about, so it would be a shame to ship the fix without it.

Also, `TestStubs.transmissionKeyStub` is set here and never reset. That is the existing convention in this file (line 172 does the same) so I am not asking you to change it in this PR, but the file has no `@AfterTest` anywhere and the stub is process-global, so it leaks into any later test that does not set it.

sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/LocalConfigStorage.kt
```diff
@@ -125,29 +135,38 @@ class LocalConfigStorage(configName: String? = null) : KeyValueStorage {
...
-        config.appKey = storage.getString(KEY_APP_KEY)
-        config.appOwnerPublicKey = storage.getString(KEY_OWNER_PUBLIC_KEY)
-        config.serverPublicKeyId = storage.getString(KEY_SERVER_PUBLIC_KEY_ID)
-        config.serverPublicKey = storage.getString(KEY_SERVER_PUBLIC_KEY)
-        val json = prettyJson.encodeToString(config)
-        BufferedWriter(FileWriter(file)).use { it.write(json) } // .use{} closes on exception even if the operation throws
-
-        // Set file permissions to 0600 (owner read/write only)
+        val targetPath = file.absoluteFile.toPath()
+        val tmpPath = Files.createTempFile(targetPath.parent, "ksm_", ".tmp")
```
mgallego-keeper, line 138: Two behavior changes worth a decision, both from moving to create-then-rename. Neither is wrong, both are currently undocumented.

**A symlinked config file is now replaced instead of written through.** The old `FileWriter(file)` followed the link and updated the target; the rename replaces the link itself with a regular file:

```
PROBE5 target still a symlink after write? false
PROBE5 original target content unchanged? true (content={})
```

Anyone who symlinks a config into a managed or shared location silently stops having their writes land where they expect, and the old target keeps stale content. Security-wise the new behavior is the better default, since following a symlink on write is a classic escalation vector, so my suggestion is to keep it and say so in the changelog rather than to restore the old behavior.

**A config file in a non-writable directory can no longer be updated.** `Files.createTempFile(targetPath.parent, ...)` needs write permission on the *directory*; `FileWriter` on an existing file needed it only on the *file*. A read-only mount holding a writable config, which does happen in containers, worked before and now fails. If that matters, falling back to an in-place write when `createTempFile` throws `AccessDeniedException` would preserve it, at the cost of the window this PR is closing. I would rather see it documented than papered over.

Minor, same area: `saveToFile()` runs from `saveString`, `saveBytes` *and* `delete`, so `initializeStorage` alone now does several create-write-rename cycles instead of several truncating writes. Not a concern for correctness or for this file's size, but it does mean the leftover-temp exposure below is per key write, not per session.

sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/LocalConfigStorage.kt
```diff
@@ -6,33 +6,43 @@ import kotlinx.serialization.decodeFromString
...
+            }
+        }
+        Files.write(tmpPath, data)
+        try {
+            Files.move(tmpPath, targetPath, StandardCopyOption.ATOMIC_MOVE)
+        } catch (_: AtomicMoveNotSupportedException) {
+            Files.move(tmpPath, targetPath, StandardCopyOption.REPLACE_EXISTING)
+        }
+    } catch (e: Exception) {
+        Files.deleteIfExists(tmpPath)
```
mgallego-keeper: `Files.deleteIfExists` throws `IOException`, and here that would replace the exception you are trying to propagate. If the write failed because the disk is full, the caller sees the cleanup failure instead of the real cause:

```kotlin
} catch (e: Exception) {
    try { Files.deleteIfExists(tmpPath) } catch (_: Exception) { }
    throw e
}
```

Small, but this is the path that only runs when something has already gone wrong, so it is the one place the original message matters most.

Separately, and not fixable in this `catch`: if the JVM is killed between `createTempFile` and the `move`, a `ksm_*.tmp` file stays behind next to the config, containing a full copy of the app key. On success there is no litter, I checked:

```
PROBE4 litter=[]
PROBE6 litter=[]
```

So this is only the crash path. Worth deciding rather than leaving to chance, given the contents: either `tmpPath.toFile().deleteOnExit()` right after creation, which covers orderly shutdown, or a sweep of stale `ksm_*.tmp` in the target directory on load. The temp files are 0600, so this is untidiness plus a stale secret on disk, not a fresh exposure.

mgallego-keeper (changes requested):
## First review of `dfd1840d`: the config-write half is right, the key-rotation half stops short of the defect

Built and ran the suite at this head: `./gradlew test` is green, 63 tests, 0 failures, 0 skipped. I also wrote throwaway probes so the claims below come from running the code rather than reading the diff, and I have pasted their output rather than describing it.

### What is right, and better than asked for

**KSM-1262 is fixed, and verified.** A config file sitting at 0644 comes back 0600 after a single write, with no temp file left behind:

```
PROBE4 perms before=rw-r--r-- after=rw------- litter=[]
PROBE6 cache.dat perms=rw------- litter=[]
```

That is a stronger fix than the ticket asked for. KSM-1262 suggested re-calling `setPosixFilePermissions` after every write, which would only shorten the window. You narrow the temp file to 0600 *before* any data reaches it and then swap it in, so there is no window at all. Details worth crediting: the temp file goes in the target's own directory so the rename stays on one filesystem; `.absoluteFile` before `.parent` avoids the null-parent trap; the `AtomicMoveNotSupportedException` fallback is sensible belt and braces. Coverage is complete too, `grep -rn 'FileOutputStream|FileWriter|Files.write|Files.newOutputStream' core/src/main storage` returns exactly the two sites this PR rewrites, so there is no third write path left behind.

**The out-of-range guard does what it claims.** Terminal on the first response, storage untouched:

```
PROBE2 out-of-table key_id=999: calls=1 msg=Server returned an unrecognized key id 999
PROBE2 stored key id: null
```

Worth noting the improvement is bigger than "clearer error": before this, an out-of-table id was persisted and every later call failed at transmission-key generation until someone hand-edited the config. Now nothing is written.

### Three blocking items

**1. KSM-1248's headline defect is still live.** The ticket is titled "Server key rotation retries with no bound in postQuery: can loop forever, and the suggested key_id is persisted unvalidated". This PR fixes the second clause and leaves the first. When the suggested id is *inside* the bundled table, `while (true)` still spins with no cap and no backoff:

```
PROBE1 in-table key_id=10: calls=50 terminal=BailOut: bail after 50 calls
PROBE1 stored key id after loop: 10
```

Fifty full encrypt, sign and POST cycles, and it only stopped because my stub threw to break out. Detail in the inline comment at `SecretsManager.kt:1849`.

**2. The changelog conflicts with the base branch.** `sdk/java/core/README.md` is the one conflicting file and the PR is currently unmergeable. Inline at `README.md:7`.

**3. The commit reverts merged comment work from #1080.** This commit's parent *is* #1080 (`8c028be6`), so these are real reversions rather than base drift. 87 of the 204 changed lines are comment and blank-line churn unrelated to either ticket, six lines put trailing whitespace back into `SecretsManager.kt`, and `RecordDataPamFieldsTest.kt` comes back byte-identical to its pre-#1080 state. Inline at `SecretsManager.kt:499`.

Items 2 and 3 both point the same way: this branch wants a rebase onto current `release/sdk/java/core/v17.4.0`, keeping only the KSM-1248 and KSM-1262 hunks. That resolves the conflict and drops the churn in one go.

The rest are smaller and non-blocking, inline: one weak assertion in the new test, a symlink behavior change, a read-only-parent regression, and an exception that can be masked in the cleanup path.

None of this touches the shape of the fix, which is sound. It is the scope that needs closing: item 1 is the reason the ticket exists.

mgallego-keeper (changes requested):
**Re-review at `dfd1840d`. No code has been pushed since my last review. The change since then is a PR description edit, and the description now claims work that is not in the tree.**

From the API rather than from reading:

```
head_sha        dfd1840df  (2026-08-17 19:17 UTC)   unchanged
force-push      2026-08-17 19:18 UTC                the only push, 2h before my review
my review       2026-08-17 21:11 UTC                CHANGES_REQUESTED
body edit #3    2026-08-18 15:35 UTC                stas-schaller
timeline after my review: nothing
```

`userContentEdits` shows three body revisions. The third one added the rotation cap, the two new test names, and the symlink / non-writable-directory caveats to the prose. The tree has none of it:

```
MAX_KEY_ROTATION_RETRIES              0 hits
keyRotationAttempt                    0 hits
testServerKeyRotationRetryCap         MISSING
testConfigFileAtomicWriteIsOwnerOnly  MISSING
```

I built this head (Zulu 8.0.492) and ran `./gradlew test`: **63 tests, 0 failures**. Green, but green over the old code. Of the four tests the description names, two ran and two do not exist.

### Description against the tree

| Description says | Tree |
| --- | --- |
| "caps key rotation retries at `MAX_KEY_ROTATION_RETRIES` (3)" | Not present. `while (true)` at `SecretsManager.kt:1809` still bounds only `throttleAttempt`. |
| "This matches the cap in the JS SDK" | No JS release branch has a rotation cap. It exists only in open, unmerged #1078. JS `MAX_THROTTLE_RETRIES` is 5, same as Java. |
| "`testServerKeyRotationRetryCap` confirms ... exactly `MAX_KEY_ROTATION_RETRIES + 1` calls" | Test does not exist. |
| "`testConfigFileAtomicWriteIsOwnerOnly` confirms `rw-------`" | Test does not exist. KSM-1262 ships with zero coverage. |
| symlink and non-writable-directory notes | Correct, and good to see, but they went into the PR body only. `README.md` is what ships and says neither. |
| KSM-1248 line claims validation and cap | `README.md:7` describes validation only, so the changelog in this PR contradicts its own description. |
| "`testIL5ConfigFieldOverridesEmbeddedTable` confirms ... not affected by the server-response validation" | Pre-existing test (the previous body said "existing"). Its stub returns `{"error":"generic"}`, so it never reaches the guard. |

KSM-1248's title is "Server key rotation retries **with no bound** in postQuery: can loop forever, and the suggested key_id is persisted unvalidated". The ticket lists two impacts; this PR fixes the second.

### All six earlier comments are still live

Re-verified at this head rather than assumed. Details are in the individual threads. Summary:

1. Unbounded rotation loop (`SecretsManager.kt:1849`), blocking, and it now also floods the config file with one temp-create/write/rename per iteration.
2. Duplicate `## 17.4.0` heading, still the only conflicting file.
3. Comment reversions of merged #1080: `RecordDataPamFieldsTest.kt` is still byte-identical to its pre-#1080 state, plus 6 new trailing-whitespace lines.
4. `contains("999")` still cannot tell which branch threw.
5. `LocalConfigStorage` symlink and non-writable-directory behavior changes, still undocumented in the changelog.
6. `Files.deleteIfExists` in the catch can replace the exception being propagated, in both functions.

### New in this pass

**A. Nothing builds this PR.** `test.java.yml` triggers on `pull_request: branches: [ master ]`. This PR's base is `release/sdk/java/core/v17.4.0`, so the Java 8/11/17/21 matrix never ran on it. The only checks on `dfd1840d` are two Socket Security runs. The matrix runs later, on the release PR #1110. That is how a Testing section naming two nonexistent tests goes unchallenged. The repo already has the pattern: `test.javascript.storage.*.yml` list their `release/storage/javascript/...` branches under `pull_request.branches`. Adding `release/sdk/java/core/**` to `test.java.yml` closes it. Filed separately.

**B. The rejection message should name the supported range.** The guard converts a legitimate future key id into a permanent hard failure: if the backend rotates to key 19, every call throws "Server returned an unrecognized key id 19" with nothing indicating that an SDK upgrade is the remedy. JS #1078 words this well. See the `SecretsManager.kt:1849` thread for a patch that lands this and the cap together.

**C. KSM-1262's ticket text is wrong in two ways**, and I am correcting it so QA does not test the wrong thing. It names `SecretsManager.kt`, which has zero file-write sites (both writes are in `LocalConfigStorage.kt`, which is what this PR correctly changed), and its premise "not on subsequent writes" does not match the old code, which chmod'd on every write. The real exposure is the create-then-chmod window, which is what this PR fixes and what the PR body describes accurately.

**D. A second unbounded loop, same defect class, out of scope here.** `getSharedFolderKey` (`SecretsManager.kt:1464`) walks server-supplied folder parents in a `while (true)` with no visited set, so two folders naming each other as parent hang the thread, and the caller's `try/catch` cannot help because a spin is not an exception. Filed separately, no change wanted in this PR.

**E. Windows is outside the security claim, and may regress slightly.** Reasoned, not tested, so flagging rather than asserting: the POSIX branch is skipped there, and `File.setReadable(false, false)` does not remove read access on Windows, so "eliminates the window where other local users can read the config" is a POSIX-only statement. Separately, replacing the target file object discards any explicit ACL an administrator had applied to the config path, since the replacement carries the temp file's inherited ACL. Both belong in the changelog wording.

### Verified positives, so please do not change these

- Legitimate rotation still converges past the new guard: transmission key ids used per call were `[7, 11]`, stored key id 11.
- The guard also catches a string-typed `key_id`: the lenient parser coerces `"999"` and the guard then rejects it.
- `{"error":"key"}` with no `key_id` is already terminal, because `KeeperError.key_id` is a non-nullable `Int`, so deserialization fails and the raw body is thrown. Java needs no analogue of #1078's `typeof` check.
- Both files land at `rw-------`, with no temp litter on success, and the cache round-trips.
- The 3-line guard itself is correct and well placed: one line earlier and IL5 deployments, which pin key id 20, would lose their actionable message.

### Where this leaves the PR

The guard is right and I am not asking for changes to it. The problem is that the description now overstates the change in a way that is dangerous if the PR is merged on the strength of the prose: a reviewer or release engineer reading the body would reasonably believe KSM-1248 is fully fixed and that four tests back it. Either revert the body to describe the validation only, or let the code catch up to the body. The second is better: the cap is small, the shape is already written in #1078, and KSM-1248 cannot honestly close without it.

Blocking set unchanged: the rotation cap, the `README.md` rebase into the existing `## 17.4.0` section, dropping the #1080 comment reversions, and a discriminating assertion plus any KSM-1262 coverage at all.

sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
mgallego-keeper: Still live at `dfd1840d`. The description now says this is capped; the code is unchanged, so I re-probed it rather than assume.

Same stub as before (always answers `{"error":"key","key_id":10}`), plus an inode read of the config file on each call:

```
PROBE1 calls=50 terminal='BAILOUT after 50 calls'
PROBE1 stored key id after loop = 10
PROBE1 distinct config-file inodes observed = 50
```

New detail worth having: with `LocalConfigStorage`, each iteration calls `saveString`, which calls `saveToFile`, which is now a temp-file create plus write plus rename. Fifty iterations produced fifty distinct config inodes. So KSM-1262's change turns this loop into a disk-write flood as well as a request flood, each iteration laying down a fresh temp copy of the client key and private key next to the config. The two fixes interact, and only in the direction of making the missing cap matter more.

One correction to the description while we are here: "This matches the cap in the JS SDK" is aspirational. No JavaScript release branch has a rotation cap; it exists only in #1078, which is still open and unmerged. Java and JS both sit at `MAX_THROTTLE_RETRIES = 5` for throttling, and neither caps rotation today.

This patch lands the cap and fixes the message problem at the same time. The message matters because the guard turns a legitimate future key id into a permanent hard failure: if the backend rotates to key 19, every call throws "Server returned an unrecognized key id 19" with nothing telling the operator that an SDK upgrade is the remedy. #1078 already words this well.

```kotlin
const val MAX_KEY_ROTATION_RETRIES = 3   // parity with #1078
// ...
var keyRotationAttempt = 0
// ...
if (!keeperPublicKeys.containsKey(error.key_id)) {
    val ids = keeperPublicKeys.keys.sorted()
    throw SecretsManagerException(
        "Server suggested unsupported key id ${error.key_id}; this SDK version supports key ids ${ids.first()}-${ids.last()}"
    )
}
if (keyRotationAttempt >= MAX_KEY_ROTATION_RETRIES) {
    throw SecretsManagerException(
        "Server key rotation exhausted $MAX_KEY_ROTATION_RETRIES retries; key id ${transmissionKey.publicKeyId} was not accepted"
    )
}
keyRotationAttempt++
options.storage.saveString(KEY_SERVER_PUBLIC_KEY_ID, error.key_id.toString())
continue
```

Two things I verified so you do not have to change them: the guard is correctly placed after the custom-key branch (one line earlier and IL5, which pins key id 20, loses its actionable message), and legitimate rotation still converges past it. Transmission key ids used per call were `[7, 11]` with a stub that rejects 7 once and suggests 11.

sdk/java/core/README.md
mgallego-keeper: Still live and still the only conflicting file, re-checked at this head:

```
$ git merge-tree --write-tree origin/release/sdk/java/core/v17.4.0 pr/1109
100644 4174b4e2... 1  sdk/java/core/README.md
100644 444f97ea... 2  sdk/java/core/README.md
100644 bafa2660... 3  sdk/java/core/README.md
```

The base still has exactly one `## 17.4.0` section (KSM-1081, KSM-1086) from `16dc9957`, and this branch adds a second one above it. A rebase plus moving these two bullets into the existing section clears the conflict completely.

There is now a second reason to touch this file. The PR description was updated to claim the retry cap, the symlink replacement, and the non-writable-directory failure, but the changelog here still describes only the validation. Since the changelog is what ships and the description is not, the two should not disagree. Whatever the code ends up doing, this entry needs to say it, including the symlink and directory-permission notes that currently live only in the PR body.

Same wording point as before, unchanged: "atomic" here means atomic against concurrent readers, not durable across a crash, since there is no `fsync` before the rename. Worth not implying.

sdk/java/core/src/test/kotlin/com/keepersecurity/secretsManager/core/SecretsManagerTest.kt
mgallego-keeper: Still live, and the description now makes this thread more urgent rather than less.

The assertion still cannot tell which branch threw. Re-probed at this head:

```
PROBE3 custom-key branch msg='Server rejected the custom server public key (id 20). The server suggested key id 999. Please update your IL5 KSM configuration.'
PROBE3 contains("999")=true   contains("unrecognized key id 999")=false
```

On the coverage gaps I raised: the description now names two tests that do not exist in the tree. I built this head with Zulu 8.0.492 and ran the suite:

```
63 tests, 0 failures
testServerKeyIdOutOfRangeIsRejected      RAN
testIL5ConfigFieldOverridesEmbeddedTable RAN   (pre-existing; the earlier body correctly said "existing")
testServerKeyRotationRetryCap            MISSING
testConfigFileAtomicWriteIsOwnerOnly     MISSING
```

So the in-table path is still uncovered, and KSM-1262 still has no test at all. Worth knowing that the suite passing is not evidence here: `test.java.yml` only triggers on `pull_request: branches: [ master ]`, and this PR's base is the release branch, so the Java 8/11/17/21 matrix never ran on it. The only checks on `dfd1840d` are two Socket Security runs. I have filed that trigger gap separately.

One more note on `testIL5ConfigFieldOverridesEmbeddedTable`, since the description now leans on it: its stub returns `{"error":"generic"}`, so it never reaches the key branch and therefore does not confirm anything about the server-response validation. The claim is directionally right but that test is not the evidence for it. A four-line test with an `IL5:...:20:<key>` storage and a `{"error":"key","key_id":999}` response is, and it would pin the branch boundary the assertion above cannot see.

sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/LocalConfigStorage.kt
mgallego-keeper: Both behavior changes are still undocumented, and the PR body now mentions them but the changelog does not. Re-probed at this head so the decisions rest on measurements:

```
PROBE6 config perms = rw-------
PROBE6 leftover tmp = []
PROBE6 inode changed on each save: 130376507 -> 130376508 (changed=true)
PROBE6 symlink survived write? false      original target still '{}': true
PROBE6 read-only dir -> java.nio.file.AccessDeniedException: .../probe6_ro_.../ksm_7987515504787726593.tmp
PROBE7 cache.dat perms = rw-------        roundtrip ok = true
```

Worth noting why the fix genuinely works, because it is not the explicit chmod doing it: when `Files.createTempFile` gets no `FileAttribute` arguments on a POSIX filesystem, the JDK supplies `rw-------` at creation time, atomically. The file is never group or other readable, and the `setPosixFilePermissions` two lines later is belt and braces. That is the substance of KSM-1262 and it holds.

Three additions to the earlier decisions, all in the same hunk:

The read-only-directory failure surfaces as a raw `AccessDeniedException` naming a `ksm_*.tmp` path the caller never asked to write. Whatever you decide about supporting that layout, the diagnostic should name the config file.

Windows is outside the claim the PR body makes. Reasoned rather than tested, so treat it as a flag: the POSIX branch is skipped there, and `File.setReadable(false, false)` does not remove read access on Windows, so "eliminates the window where other local users can read the config" is a POSIX-only statement. Separately, replacing the target file object discards any explicit ACL an administrator had applied to the config path, since the replacement carries the temp file's inherited ACL. Both belong in the changelog scoping.

Small asymmetry introduced here: the write is now UTF-8 (`String.toByteArray()` at line 161) while the read is still the platform default (`FileReader` at line 128). Bounded in practice, since every value is base64 except `hostname`, but the two halves no longer agree by construction on a JDK 8 JVM whose default charset is not UTF-8. Moving the read to UTF-8 makes it symmetric for free.

Unrelated one-liner while you are in the imports: `PosixFilePermission` (singular) at line 12 is unused. It was unused before this PR too, so entirely your call.

sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/LocalConfigStorage.kt
mgallego-keeper: Still live, and it applies in two places rather than one. After the change there are two identical catch blocks, `LocalConfigStorage.kt:38` in `saveCachedValue` and `LocalConfigStorage.kt:168` in `saveToFile`, each calling `Files.deleteIfExists(tmpPath)` unguarded. Both can replace the exception you are trying to propagate, so both need the wrap:

```kotlin
} catch (e: Exception) {
    try { Files.deleteIfExists(tmpPath) } catch (_: Exception) { }
    throw e
}
```

The leftover-temp-file question is unchanged, and I re-confirmed there is no litter on the success path (`PROBE6 leftover tmp = []`, `PROBE7 cwd leftover tmp = []`), so it remains a crash-path decision only: `deleteOnExit()` right after creation, or a sweep of stale `ksm_*.tmp` on load.

One thing that does change the weight of it: the unbounded rotation loop now drives one temp-create/write/rename per iteration (fifty distinct config inodes in fifty iterations), so on the crash path the exposure is per key write inside that loop, not once per session.

sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
mgallego-keeper: Still present at this head. Re-measured with a cleaner comparison than last time: changed lines against the pre-#1080 file versus against the base.

```
                             vs PRE-1080   vs BASE
RecordDataPamFieldsTest.kt      0+0          1+1     byte-identical to pre-#1080
KeeperRecordLinkTest.kt         1+1          3+3
SecretsManager.kt              19+27        33+30    closer to pre-#1080 than to base
Notation.kt                     0+1          0+1
RecordData.kt                   3+3          1+1
```

`SecretsManager.kt` being closer to the pre-#1080 file than to the file it is based on is the clearest signal that this is a stale working copy rather than an intentional revert, which is also the easy fix. Plus 6 new trailing-whitespace lines, unchanged from my earlier count.

`RecordData.kt` and `Notation.kt` are trailing-comment edits on code lines, no logic in either, so my earlier "code +1/-1" line for `RecordData.kt` was a classification artifact of counting a line that carries both. Nothing functional hides in those two files.

A rebase onto current `release/sdk/java/core/v17.4.0`, keeping only the `LocalConfigStorage.kt`, `SecretsManager.kt:1849`, `SecretsManagerTest.kt` and `README.md` hunks, still fixes this and the README conflict together.

mgallego-keeper (approved):
## Approved at `123da7e9`. All three blocking items are closed, verified by running the code.

```
head          123da7e9  (2026-08-18 16:59 UTC)   force-push, rebased onto 16dc9957
parent        16dc9957  == current base tip       no drift
mergeable     MERGEABLE / CLEAN                   was CONFLICTING
build         ./gradlew clean build test          BUILD SUCCESSFUL, 0 Kotlin warnings
tests         66 passed, 0 failures, 0 skipped    was 63
```

Built with Zulu 8.0.492 and re-probed each claim rather than reading the diff.

### The three blockers

**1. The rotation loop is capped.** Probed against a real `LocalConfigStorage` so the disk side is included:

```
calls=4  keyIdsUsed=[7, 10, 10, 10]
terminal=SecretsManagerException: Server key rotation exhausted 3 retries; key id 10 was not accepted
distinct config inodes = 4   litter=[]   perms=rw-------
```

Four POSTs where it was 50 and still climbing, and four config inodes where it was fifty, so the disk-write flood goes away with the same fix. Legitimate rotation still converges (`keyIdsUsed=[7, 11]`), including one that lands on the last allowed retry (`calls=4 keyIdsUsed=[7, 8, 9, 10]` then success), so the cap does not truncate a real multi-hop rotation.

**2. The README conflict is gone and the entry is accurate.** Both bullets sit inside the existing `## 17.4.0` section and describe the cap, the symlink replacement and the non-writable-directory failure with a remedy. Dropping the word "atomic" also settles my durability-versus-visibility wording point.

**3. The #1080 reversions are gone.** The rebase dropped all of it. `RecordDataPamFieldsTest.kt`, `KeeperRecordLinkTest.kt`, `Notation.kt`, `RecordData.kt` and `IL5EdgeCaseTest.kt` are untouched, and `git diff --check` reports zero new trailing whitespace. The diff is 4 files and about 164 lines of real change.

### The non-blocking threads, also closed

The assertion now discriminates, checked rather than assumed:

```
custom-key msg='Server rejected the custom server public key (id 20)...'
contains('unsupported key id 999')=false   contains('custom server public key')=true
```

Both `Files.deleteIfExists` calls are wrapped so the real cause survives, the read is `readText(Charsets.UTF_8)` against a `toByteArray()` write so the two halves agree by construction, the unused `PosixFilePermission` import is gone, and all four tests named in the description exist and ran.

The parity claim is now simply true rather than aspirational: #1078 merged today at 16:03 UTC, 56 minutes before this push, with `MAX_KEY_ROTATION_RETRIES = 3`, matching guard order and a near-identical terminal message.

### Two things worth recording

Java correctly does not need the explicit `key_id` type guard that JS carries. `data class KeeperError(val key_id: Int, ...)` makes kotlinx.serialization reject a missing, string, float or Int-overflowing `key_id` at parse time, so all of those fall through to the generic throw in a single call. I probed all nine boundary bodies and none of them loops or poisons storage. No parity work is needed here.

The 0600 protection is load-bearing at creation, not at the chmod. `Files.createTempFile` with no `FileAttribute` argument returns `rw-------` on JDK 8 even under `umask 000`, which I ran to confirm. The file is owner-only from the instant it exists, and the explicit `setPosixFilePermissions` two lines later is belt and braces. That is why this closes the window rather than shortening it.

### Five follow-ups, none blocking this merge

1. **Nothing still builds this PR.** `test.java.yml` triggers on `pull_request: branches: [ master ]`, so the Java 8/11/17/21 matrix has never run on this branch. The only checks on `123da7e9` are two Socket Security runs, and my local build remains the only evidence the suite passes. Filed separately, and worth closing before the release PR.
2. **Two ticket refs went back into code comments**, at `SecretsManager.kt:39` and `SecretsManagerTest.kt:257`. The commit message says it removes ticket refs from comments, and the base has exactly one left across all of `sdk/java/core/src` after #1080. `// parity with JS SDK open PR #1078` is also already stale. "cap of 3, matching the JavaScript SDK" says the same thing and stays true.
3. **`testConfigFileAtomicWriteIsOwnerOnly` breaks the suite on Windows.** It calls `Files.getPosixFilePermissions` unguarded and is the only POSIX-dependent test in the file. CI is ubuntu-only so CI will not catch it, but `./gradlew test` on Windows throws `UnsupportedOperationException`. One `assumeTrue` on the file store supporting `posix` fixes it.
4. **The read-only-directory diagnostic** still surfaces raw: `AccessDeniedException: .../ksm_7649927828172540471.tmp`, naming a path the caller never asked to write. The changelog now documents the failure mode with a remedy, which was the substance of the ask, so this is message quality only.
5. **Crash-path temp litter** is unchanged, with no `deleteOnExit` and no stale-`ksm_*.tmp` sweep. Success paths leave nothing behind, and the cap bounds the exposure to four temp copies instead of unbounded, so this stays low priority.

Two smaller notes. Four unrelated comment rewrites still ride along in the test file, forward churn this time rather than reverts. And the body says "Breaking Changes: None" while the same body describes "Two behavior changes"; the changelog gets this right, so only the body line disagrees.

Out of scope, for whoever owns it: `GcpKeyValueStorage.java` writes its KMS-encrypted config with default permissions at three sites. It is the only other Java module that writes a config file.

sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
mgallego-keeper: Fixed, and verified rather than assumed. The cap is in at `SecretsManager.kt:1851` and the loop terminates:

```
calls=4  keyIdsUsed=[7, 10, 10, 10]
terminal=SecretsManagerException: Server key rotation exhausted 3 retries; key id 10 was not accepted
distinct config inodes = 4   litter=[]
```

Four POSTs instead of 50 and still climbing, and four config inodes instead of fifty, so the disk-write flood I flagged as a knock-on effect goes with it. Legitimate rotation still converges past both guards (`[7, 11]`), and a rotation that lands on the last allowed retry still succeeds (`[7, 8, 9, 10]` then 200), so the cap does not truncate a real multi-hop rotation. The message names the supported range, and the range is honest: `keeperPublicKeys` is keyed 7 through 18 with no gaps.

The parity note in my earlier reply no longer applies. #1078 merged today at 16:03 UTC, 56 minutes before this push, with the same cap of 3, the same guard order and a near-identical terminal message. The two SDKs now tell users the same story.

One cosmetic leftover, not worth another round: `// parity with JS SDK open PR #1078` at line 39 is already stale, since #1078 is merged. "cap of 3, matching the JavaScript SDK" says the same thing and stays true. Closing this thread.

sdk/java/core/README.md
mgallego-keeper: Fixed. Both bullets are inside the existing `## 17.4.0` section now, and the merge is clean:

```
mergeable = MERGEABLE / CLEAN     (was CONFLICTING)
parent    = 16dc9957 == base tip  (no drift)
```

The entry also matches the tree: it names the cap, the symlink replacement and the non-writable-directory failure, and it gives a remedy for the last one. Dropping the word "atomic" settles my wording point too, since nothing now implies crash durability. Closing.

sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/SecretsManager.kt
mgallego-keeper: Fixed by the rebase onto `16dc9957`. `RecordDataPamFieldsTest.kt`, `KeeperRecordLinkTest.kt`, `Notation.kt`, `RecordData.kt` and `IL5EdgeCaseTest.kt` are no longer touched at all, and `git diff --check` reports zero new trailing whitespace. The diff is 4 files and about 164 lines, all of it KSM-1248 or KSM-1262. Closing.

sdk/java/core/src/test/kotlin/com/keepersecurity/secretsManager/core/SecretsManagerTest.kt
mgallego-keeper: The assertion is fixed and it now genuinely discriminates. Probed both branches at this head:

```
custom-key msg='Server rejected the custom server public key (id 20). The server suggested key id 999...'
contains('unsupported key id 999') = false
contains('custom server public key') = true
```

So the two branches can no longer be confused, and `testCustomServerKeyGuardTakesPriorityOverKeyTable` pins the boundary directly rather than relying on which storage happens to be configured. Both coverage gaps are closed too: `testServerKeyRotationRetryCap` covers the in-table path, and `testConfigFileAtomicWriteIsOwnerOnly` gives KSM-1262 its first test. 63 tests to 66, all green on Zulu 8.0.492.

The thread is resolved and I am not reopening it. One small thing does remain in the new test, non-blocking and also recorded in my approval.

**`testConfigFileAtomicWriteIsOwnerOnly` breaks the suite on Windows.** Line 263 calls `Files.getPosixFilePermissions` unguarded, and this is the only POSIX-dependent test in the file. CI will not catch it, because `test.java.yml` runs `ubuntu-latest` only, but a developer running `./gradlew test` on Windows gets `UnsupportedOperationException` rather than a skip. One line fixes it:

```kotlin
org.junit.jupiter.api.Assumptions.assumeTrue(
    configFile.toPath().fileSystem.supportedFileAttributeViews().contains("posix"),
    "POSIX permissions not supported on this file system"
)
```

Same reasoning as the changelog scoping: the guarantee is POSIX-only, so the test that asserts it should say so rather than fail.

Unrelated to this PR and still just an observation: `TestStubs.transmissionKeyStub` is process-global and never reset, and the file still has no `@AfterTest`. Four tests set it now instead of two, so it leaks a little further than it did.

sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/LocalConfigStorage.kt
mgallego-keeper: Both behavior changes are documented now, in the changelog rather than only the PR body, which was the ask. Re-probed at this head:

```
perms before=rw-r--r-- after=rw------- litter=[]
symlink survived write? false   original target still '{}': true
read-only dir -> java.nio.file.AccessDeniedException: .../ksm_7649927828172540471.tmp
cache.dat perms=rw------- roundtrip=hello cwdLitter=[]
UTF-8 round trip of a non-ASCII hostname: ok=true
```

The charset asymmetry is fixed: the read is `readText(Charsets.UTF_8)` at line 127 and the write is `toByteArray()`, so the two halves agree by construction rather than by luck of the JVM default. The unused `PosixFilePermission` import is gone.

Worth recording why the fix works, since it is not the explicit chmod doing the work: `Files.createTempFile` with no `FileAttribute` argument returns `rw-------` on JDK 8 even under `umask 000`. I ran that to check rather than trusting the docs. The file is owner-only from the instant it exists, so there is no window at all, and `setPosixFilePermissions` two lines later is belt and braces.

One item is unaddressed. It is a message-quality call rather than a behavior call, so it does not reopen this thread.

**The read-only-directory failure still surfaces raw**, naming a `ksm_*.tmp` path the caller never asked to write. Now that the changelog tells operators to move the config to a writable directory, the exception is the place they will actually read first, and it currently points at a temp file that no longer exists. Wrapping `createTempFile` so the message names the config file and the directory would close the loop:

```kotlin
val tmpPath = try {
    Files.createTempFile(targetPath.parent, "ksm_", ".tmp")
} catch (e: IOException) {
    throw SecretsManagerException(
        "Cannot write config $targetPath: directory ${targetPath.parent} is not writable. " +
        "Move the config to a writable directory or use InMemoryStorage with an injected config string.", e
    )
}
```

Windows scoping is handled by the changelog now saying "On POSIX systems". The remaining unstated part is that replacing the target file discards any explicit ACL an administrator applied to the config path, since the replacement carries the temp file's inherited ACL. Your call whether that is worth a clause.

sdk/java/core/src/main/kotlin/com/keepersecurity/secretsManager/core/LocalConfigStorage.kt
mgallego-keeper: The exception-masking half is fixed in both places, `saveCachedValue` at line 36 and `saveToFile` at line 168. The real cause now survives a failed cleanup.

The crash-path decision is unchanged, and does not reopen this thread: no `deleteOnExit`, no sweep of stale `ksm_*.tmp` on load. Success paths leave nothing behind, re-confirmed at this head:

```
config litter=[]   cache.dat cwdLitter=[]   perms=rw------- on both
```

One thing that got better on its own: with the rotation cap in, a server that keeps re-suggesting a key drives 4 temp-create/write/rename cycles instead of an unbounded number, so the worst-case crash-path exposure is now four stale copies of the client key and private key rather than however long the loop ran. That drops this from "worth deciding before release" to a tidy-up whenever you are next in the file. `tmpPath.toFile().deleteOnExit()` right after creation is the one-liner; a sweep on load is the thorough version.

PR #1117, opened 2026-08-19
fix(java): restore Java API compatibility and unblock the 17.4.0 release

mgallego-keeper:
Review findings from a pass over the v17.4.0 release branch (#1110), covering #1116 (KSM-1207) and #1115 (KSM-1176). Five fixes, one commit each; every commit builds and tests green on its own.

Two of these are silent public API breaks that CI cannot see: `test.java.yml` runs `gradle build test` with `working-directory: ./sdk/java/core`, so it compiles core against itself and never compiles the Java consumers in `integration/` or `examples/`. Kotlin callers are unaffected either way, which is why the four green JDK matrix jobs on #1110 pass.

## 1. `postFunction` lost its four-argument Java overload (KSM-1207)

Kotlin default arguments do not exist in bytecode, so adding two defaulted parameters replaced the published static method rather than extending it.

| | `SecretsManager.postFunction` |
|---|---|
| 17.3.0 | `(String, TransmissionKey, EncryptedPayload, boolean)` |
| 17.4.0 before this PR | `(String, TransmissionKey, EncryptedPayload, boolean, int, int)` only |

Compiling the existing call site against 17.4.0:

```
error: method postFunction in class SecretsManager cannot be applied to given types;
  required: String,TransmissionKey,EncryptedPayload,boolean,int,int
  found:    String,TransmissionKey,EncryptedPayload,boolean
```

The runtime half is worse. Bytecode compiled against 17.3.0, run against 17.4.0 as a drop-in jar upgrade:

```
java.lang.NoSuchMethodError: 'KeeperHttpResponse SecretsManager.postFunction(
    String, TransmissionKey, EncryptedPayload, boolean)'
```

That compiles clean and fails at the first vault call.

Callers of the four-argument form:
- `integration/servicenow-external-credential-resolver/src/main/java/com/keepersecurity/secretsManager/CredentialResolver.java:333` (pinned `core:17.3.0`)
- `examples/java/hello-secret-custom-caching/app/src/main/java/org/example/App.java:24` (pinned `core:17.1.2`)

The pins defer this rather than prevent it. External customers following the documented custom-caching pattern have no pin.

Fix: `@JvmOverloads`, which restores the four- and five-argument forms alongside the new one.

## 2. `KeeperRecord`'s constructor went from nine parameters to ten (KSM-1176)

`isEditable` was inserted between `revision` and `files`. `KeeperRecord` carries no `@JvmOverloads`, so Java sees only the full-arity constructor. Verbatim from this repo's own `integration/servicenow-external-credential-resolver/src/test/java/com/keepersecurity/secretsManager/KeeperCredentialMapperTest.java:39`:

```java
new KeeperRecord(new byte[32], uid, null, null, null, data, 0L, null, null);
```

compiles against 17.3.0 and fails against 17.4.0 with `constructor KeeperRecord cannot be applied to given types`.

`@JvmOverloads` alone does not fix this one: with `isEditable` mid-list the generated nine-argument overload drops `links` and ends in `boolean`, so the call still fails. The field has to move to the end **and** get `@JvmOverloads`. `javap` then shows the original nine-argument constructor restored exactly.

`KeeperRecord` is not `@Serializable`, so reordering does not affect JSON decoding. The internal construction site now uses named arguments so a future field cannot silently shift positions again.

## 3. `uploadFile` ignored the timeouts the caller configured (KSM-1207)

`uploadFile(options, ownerRecord, file)` already holds the options and passes them to `postQuery` for the `add_file` request, but the private helper doing the actual upload hardcoded the module defaults. Setting `readTimeoutMillis = 120_000` was honoured for the metadata request and silently ignored for the upload in the same call.

`downloadFile` keeps the module constants, since its public entry points take a `KeeperFile` and no options.

## 4. The timeout test did not test the timeout (KSM-1207)

`testSecretsManagerOptionsAcceptsCustomTimeouts` asserted that `SecretsManagerOptions` echoes its own constructor arguments. That holds for any Kotlin data class, and it passes unchanged if both `connection.readTimeout` assignments are deleted.

`TimeoutTest` points `postFunction` at a `ServerSocket` that accepts the connection and then goes silent, so the client blocks reading the ServerHello. Measured:

| | `readTimeout_boundsAStalledServer` |
|---|---|
| with the fix | passes in 1.1s (`SocketTimeoutException: Read timed out`) |
| with `connection.readTimeout` removed | fails at 20.0s, `call did not return within 20s; the timeout was never applied` |

The call runs under a watchdog on a daemon thread, so a regression fails the build rather than hanging the Gradle test JVM. The documented 5000/30000 defaults are asserted separately.

No behavioural test for connect timeout: it needs a blackholed address, which is not reproducible across CI environments.

## 5. The version was never bumped

`build.gradle.kts` and `KEEPER_CLIENT_VERSION` still said 17.3.0 while the branch and changelog say 17.4.0. `publish.maven.java.core.yml` takes no version input and reads the coordinate straight from `build.gradle.kts`, so publishing would have attempted to republish 17.3.0 to Maven Central and reported `mj17.3.0` in the client version header.

## Verification

- 70 tests, 0 failures. Each of the five commits builds and tests green on its own.
- `javap` confirms the four-argument `postFunction` and the nine-argument `KeeperRecord` constructor are both back.
- Both real Java call sites compile against the fixed build.
- 17.3.0-compiled bytecode links against the fixed 17.4.0 classes (no `NoSuchMethodError`).

## Not addressed here

Left out to keep this reviewable; happy to open tickets:

- A custom `queryFunction` still bypasses both new timeout options, including core's own `cachingPostFunction`. Worth a KDoc note at minimum.
- `SecretsManagerOptions` does not validate the timeouts. `0` is accepted and means infinite in `URLConnection`, so a caller can silently reopen the hang KSM-1207 closed; negatives throw `IllegalArgumentException` from inside the first request instead of at construction.
- `SocketTimeoutException` escapes raw, where every other failure in `postQuery` is a `SecretsManagerException` or `KeeperThrottleException`.
- `SecretsManagerOptions` still uses `5_000` / `30_000` literals while `DEFAULT_CONNECT_TIMEOUT_MS` / `DEFAULT_READ_TIMEOUT_MS` hold the same values.
- Two unbounded `HttpURLConnection` sites remain outside core, in `sdk/java/storage/.../gcp/KMSUtils.java:341` and `:413`.
- Worth considering for CI: a job that compiles `integration/` and `examples/java/**` against the local core, and/or the `binary-compatibility-validator` plugin. Either would have caught items 1 and 2 on the PRs that introduced them.

### DIFF
--- a/sdk/java/core/src/test/kotlin/com/keepersecurity/secretsManager/core/TimeoutTest.kt
+++ b/sdk/java/core/src/test/kotlin/com/keepersecurity/secretsManager/core/TimeoutTest.kt
@@ -1,0 +1,89 @@
+package com.keepersecurity.secretsManager.core
+
+import kotlinx.serialization.ExperimentalSerializationApi
+import java.net.ServerSocket
+import java.net.SocketTimeoutException
+import java.util.concurrent.ExecutionException
+import java.util.concurrent.Executors
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.TimeoutException
+import kotlin.test.*
+
+// Connect/read timeouts on the built-in HTTP transport. The behavioural test points
+// postFunction at a socket that accepts the TCP connection and then goes silent, so the client
+// blocks reading the ServerHello, which is exactly the stall readTimeout has to bound. The call
+// runs under a watchdog on a daemon thread: without the timeout the read never returns, and a
+// plain assertion would hang the Gradle test JVM instead of failing it.
+@ExperimentalSerializationApi
+internal class TimeoutTest {
+
+    private val stubTransmissionKey = TransmissionKey(7, ByteArray(32), ByteArray(32))
+    private val stubPayload = EncryptedPayload(ByteArray(8), ByteArray(8))
+
+    // Generous relative to the 1s timeout under test: this is the "it hung" tripwire, not a
+    // latency assertion, so a loaded CI runner must not trip it.
+    private val watchdogSeconds = 20L
+    private val probeReadTimeoutMillis = 1_000
+
+    @Test
+    fun timeoutDefaults_matchDocumentedValues() {
+        val options = SecretsManagerOptions(InMemoryStorage())
+        assertEquals(5_000, options.connectTimeoutMillis, "documented default connect timeout")
+        assertEquals(30_000, options.readTimeoutMillis, "documented default read timeout")
+    }
+
+    @Test
+    fun timeoutOptions_acceptCustomValues() {
+        val options = SecretsManagerOptions(
+            InMemoryStorage(),
+            connectTimeoutMillis = 2_000,
+            readTimeoutMillis = 10_000
+        )
+        assertEquals(2_000, options.connectTimeoutMillis)
+        assertEquals(10_000, options.readTimeoutMillis)
+    }
+
+    @Test
+    fun readTimeout_boundsAStalledServer() {
+        ServerSocket(0).use { server ->
+            val accepted = Thread { runCatching { server.accept() } }
+            accepted.isDaemon = true
+            accepted.start()
+
+            val url = "https://127.0.0.1:${server.localPort}/"
+            val elapsedMillis = withWatchdog {
+                val start = System.nanoTime()
+                assertFailsWith<SocketTimeoutException> {
+                    postFunction(
+                        url,
+                        stubTransmissionKey,
+                        stubPayload,
+                        true,
+                        readTimeoutMillis = probeReadTimeoutMillis
+                    )
+                }
+                (System.nanoTime() - start) / 1_000_000
+            }
+            assertTrue(
+                elapsedMillis >= probeReadTimeoutMillis / 2,
+                "returned in ${elapsedMillis}ms, too fast to have been the read timeout"
+            )
+        }
+    }
+
+    // Runs [block] on a daemon thread, failing (rather than blocking) if it never returns.
+    private fun <T> withWatchdog(block: () -> T): T {
+        val executor = Executors.newSingleThreadExecutor { runnable ->
+            Thread(runnable, "timeout-test").apply { isDaemon = true }
+        }
+        try {
+            return executor.submit(block).get(watchdogSeconds, TimeUnit.SECONDS)
+        } catch (_: TimeoutException) {
+            fail("call did not return within ${watchdogSeconds}s; the timeout was never applied")
+        } catch (e: ExecutionException) {
+            throw e.cause ?: e
+        } finally {
+            executor.shutdownNow()
+        }
+    }
+}
--- a/sdk/java/core/src/test/kotlin/com/keepersecurity/secretsManager/core/TimeoutTest.kt
+++ b/sdk/java/core/src/test/kotlin/com/keepersecurity/secretsManager/core/TimeoutTest.kt
@@ -2,11 +2,13 @@ package com.keepersecurity.secretsManager.core
 
 import kotlinx.serialization.ExperimentalSerializationApi
 import java.net.ServerSocket
+import java.net.Socket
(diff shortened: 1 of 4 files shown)


sdk/java/core/src/test/kotlin/com/keepersecurity/secretsManager/core/TimeoutTest.kt
```diff
@@ -0,0 +1,89 @@
...
+            readTimeoutMillis = 10_000
+        )
+        assertEquals(2_000, options.connectTimeoutMillis)
+        assertEquals(10_000, options.readTimeoutMillis)
+    }
+
+    @Test
+    fun readTimeout_boundsAStalledServer() {
+        ServerSocket(0).use { server ->
+            val accepted = Thread { runCatching { server.accept() } }
```
stas-schaller: The `Socket` returned by `server.accept()` is discarded here: `runCatching` wraps it in a `Result<Socket>` that the thread lambda throws away, making the socket immediately eligible for GC. If the finalizer runs before the 1-second read timeout fires, the server side closes, the client gets an unexpected EOF during the TLS handshake, and the assertion fails with `SSLHandshakeException` rather than `SocketTimeoutException`.

Capture it so it stays reachable through the `use` block:

```kotlin
var acceptedSocket: java.net.Socket? = null
val accepted = Thread { runCatching { acceptedSocket = server.accept() } }
```

Then after `withWatchdog { ... }` returns, call `acceptedSocket?.close()`. That makes the race structurally impossible rather than just statistically unlikely.

stas-schaller (changes requested):
One request before this merges: anchor the accepted socket so a GC cycle during the 1-second timeout window can't close the connection early and produce `SSLHandshakeException` instead of `SocketTimeoutException`. Everything else here is looking good!

sdk/java/core/src/test/kotlin/com/keepersecurity/secretsManager/core/TimeoutTest.kt
mgallego-keeper: Fixed in 0bb2c33d.

Confirmed, and it is not marginal. Under forced GC the discarded socket never once produced `SocketTimeoutException`:

| accepted socket | JDK 8 | JDK 21 |
|---|---|---|
| discarded (before) | `SSLException` after 30-45ms | `SocketException` after 55-85ms |
| anchored (after) | `SocketTimeoutException` after ~1.04s | `SocketTimeoutException` after ~1.07s |

Running the same negative control through Gradle, with a GC-pressure thread inside the test and the anchor removed, fails in 0.15s on `Expected an exception of class java.net.SocketTimeoutException`. With the anchor and that same pressure thread it passes 3/3 on both JDK 8 and JDK 21.

One deviation from your snippet: I held it in an `AtomicReference<Socket?>` rather than a captured `var`. The anchoring works either way, since the `Ref.ObjectRef` a captured var compiles to is reachable from the test frame regardless of visibility. But the acceptor thread writes it and the test thread reads it to call `close()`, and `Ref.ObjectRef.element` is not volatile, so the reader could see a stale null and silently skip the close. `AtomicReference` plus `acceptor.join(1_000)` before the close makes that handoff ordered rather than probable, which seemed in the spirit of "structurally impossible rather than statistically unlikely".

Also moved the assertions into a `try` and the close into a `finally`, so the socket is released even when the test fails. It was leaking on every run before.

Full suite green on JDK 8 and JDK 21: 70 tests, 0 failures, no warnings.

stas-schaller (approved):
Fix is correct and better than the suggestion. AtomicReference + acceptor.join(1_000) makes the handoff ordered rather than just probably-visible. The empirical table confirms the race was real and is closed. Good to merge.

PR #1118, opened 2026-08-19
KSM-1270 Guard getSharedFolderKey against parent-cycle infinite loop

stas-schaller:
## Summary

- Replaces the unbounded `while (true)` loop in `getSharedFolderKey` with a `HashSet<String>` visited-UID guard. When a folder UID is seen a second time the function returns `null`, which triggers the existing per-folder `try/catch` in `getFolders` to log and skip the affected folder rather than hang the thread indefinitely.
- Adds `testGetFoldersSkipsFolderParentCycle` to verify the call completes and returns an empty list when the server supplies two folders whose `parent` fields form a cycle.

## Test plan

- [ ] `./gradlew test` passes (all existing tests green + new cycle test)
- [ ] New test `testGetFoldersSkipsFolderParentCycle` completes without hanging and asserts an empty list

### DIFF
--- a/sdk/java/core/src/test/kotlin/com/keepersecurity/secretsManager/core/SecretsManagerTest.kt
+++ b/sdk/java/core/src/test/kotlin/com/keepersecurity/secretsManager/core/SecretsManagerTest.kt
@@ -427,7 +427,7 @@ internal class SecretsManagerTest {
         assertEquals("Good Folder", folders[0].name)
     }
 
-    @Test
+    @Test(timeout = 5_000)
     fun testGetFoldersSkipsFolderParentCycle() {
         val transmissionKey = ByteArray(32) { it.toByte() }
         TestStubs.transmissionKeyStub = { transmissionKey }
--- a/sdk/java/core/src/test/kotlin/com/keepersecurity/secretsManager/core/SecretsManagerTest.kt
+++ b/sdk/java/core/src/test/kotlin/com/keepersecurity/secretsManager/core/SecretsManagerTest.kt
@@ -427,6 +427,30 @@ internal class SecretsManagerTest {
         assertEquals("Good Folder", folders[0].name)
     }
 
+    @Test
+    fun testGetFoldersSkipsFolderParentCycle() {
+        val transmissionKey = ByteArray(32) { it.toByte() }
+        TestStubs.transmissionKeyStub = { transmissionKey }
+
+        val appKey = getRandomBytes(32)
+        val dummyFolderKey = bytesToBase64(getRandomBytes(60))
+
+        // Two folders whose parent references form a cycle: neither has a null parent,
+        // so getSharedFolderKey can never reach a root. Both must be skipped.
+        val responseJson = """{"encryptedAppKey":null,"folders":[{"folderUid":"folder-a","folderKey":"$dummyFolderKey","data":null,"parent":"folder-b","records":null},{"folderUid":"folder-b","folderKey":"$dummyFolderKey","data":null,"parent":"folder-a","records":null}],"records":null}"""
+        val encryptedResponse = encrypt(stringToBytes(responseJson), transmissionKey)
+
+        val storage = InMemoryStorage()
+        initializeStorage(storage, "US:FAKE_CLIENT_KEY")
+        storage.saveBytes(KEY_APP_KEY, appKey)
+
+        val options = SecretsManagerOptions(storage, queryFunction = { _, _, _ -> KeeperHttpResponse(200, encryptedResponse) })
+        val folders = getFolders(options)
+
+        assertEquals(0, folders.size, "both cyclic folders must be skipped; the call must complete without hanging")
+    }
+
+
 //    @Test // uncomment to debug the integration test
     fun integrationTest() {
         val trustAllPostFunction: (
--- a/sdk/java/core/README.md
+++ b/sdk/java/core/README.md
@@ -17,6 +17,8 @@ For more information see our official documentation page https://docs.keeper.io/
 - KSM-1248 - Server-supplied key IDs are validated against the embedded public key table before being stored. An unrecognized key ID throws `SecretsManagerException` and leaves storage unchanged. Key rotation retries are now capped at `MAX_KEY_ROTATION_RETRIES` (3); exhausted retries throw a typed error naming the last suggested key ID.
 - KSM-1269 - Fixed the Java CI workflow not running on pull requests targeting release branches.
 The test matrix now triggers on both `master` and `release/sdk/java/core/**` targets.
+- KSM-1270 - Fixed `getSharedFolderKey` looping indefinitely when server folder data contains a parent cycle.
+The function now tracks visited folder UIDs and returns `null` on re-visit; `getFolders` skips the affected folders and continues normally.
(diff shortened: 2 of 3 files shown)


sdk/java/core/src/test/kotlin/com/keepersecurity/secretsManager/core/SecretsManagerTest.kt
```diff
@@ -420,6 +420,29 @@ internal class SecretsManagerTest {
         assertEquals("Good Folder", folders[0].name)
     }
 
+    @Test
```
mgallego-keeper: Could this test get an explicit timeout before merge?

This is the one change I would ask for, and it is not hypothetical. I reverted the `SecretsManager.kt` hunk and re-ran this test: it ran past 150 seconds with no result and I had to kill the Gradle workers manually.

That is the failure mode this test will have in CI if the guard ever regresses. It will not go red, it will hang the `test` task until the job-level timeout kills the whole build, and the test report will show no failure at all, just a dead job. For a regression test whose entire subject is a liveness bug, that is the wrong signal, and it is the one case where an unbounded test is actively worse than no test.

The module uses `kotlin-test-junit`, so this is JUnit 4 and JUnit 5's `@Timeout` is not available. But `kotlin.test.Test` is a typealias for `org.junit.Test`, so the parameter passes straight through with no new dependency and no import change:

```kotlin
@Test(timeout = 5_000)
fun testGetFoldersSkipsFolderParentCycle() {
```

The test currently completes in 3 ms, so 5 s is generous headroom on a slow CI runner while still failing fast on a real regression.

mgallego-keeper (commented):
The guard is correct and I verified it end to end rather than by inspection.

**What I ran**, on Zulu JDK 8 (aarch64), against `489fc20`:

- `./gradlew test` on the branch as-is: 69 tests, 0 failures, 0 errors. `testGetFoldersSkipsFolderParentCycle` passes in 3 ms.
- Then I reverted *only* the `SecretsManager.kt` hunk, keeping the new test, and re-ran it. It ran past 150 seconds with no result and I had to kill the Gradle workers. So the test genuinely exercises the bug, and the guard genuinely fixes it.

**On the fix itself:** termination is bounded at `responseFolders.size + 1` iterations, since every pass either inserts a new UID or exits. The acyclic path is unchanged: same `find`, same two return sites, so well-formed folder data behaves exactly as before. Self-parent (`a -> a`) and longer rings both resolve to `null` correctly.

One request inline, plus two minor notes below. Nothing blocking beyond the inline one.

---

**Minor 1, error message granularity.** All three `null` returns from `getSharedFolderKey` (parent UID absent from the response, parent present but not yet decrypted, cycle detected) collapse into the caller's single `"Folder data inconsistent - unable to locate shared folder"`. A cycle is a server-data defect worth naming on its own, and it is the entire subject of this ticket, so an operator hitting it in the field currently cannot tell it apart from the other two causes. Worth a distinct message.

**Minor 2, README formatting.** The KSM-1270 entry wraps onto an unindented second line while every other entry in the 17.4.0 section is a single line. Markdown lazy continuation still renders it as one bullet, so this is cosmetic only, but it reads as a stray line in the raw file. Suggest folding it into one line.

---

**Out of scope for this PR, flagged for follow-up:** the same unbounded walk is live in four sibling SDKs, which I checked on `master` at `8a93aa7e`:

| SDK | Location | Loop |
| --- | --- | --- |
| .NET | `sdk/dotNet/SecretsManager/SecretsManagerClient.cs:1179` | `while (true)`, no guard |
| Python | `sdk/python/core/keeper_secrets_manager_core/core.py:854` | `while True`, no guard |
| Ruby | `sdk/ruby/lib/keeper_secrets_manager/core.rb:1105` | `while parent_uid`; in a cycle `parent_uid` is never nil, so it never exits |
| Rust | `sdk/rust/src/core/core.rs:1750` | `loop { }`, no guard |

JavaScript core has no equivalent function, and Go is in the separate `secrets-manager-go` repo. Tickets are being filed per SDK; no change requested here.

Also worth knowing while you are in this function, and again not a request against this PR: `getSharedFolderKey` resolves the root's key out of `folders`, the accumulator of folders decrypted *so far*. If the server ever returns a child ahead of its shared-folder root, the root is not in `folders` yet, `find` returns null, and the child is skipped with that same generic message. That ordering dependency predates this change.

mgallego-keeper (approved):
Approving. All three points are addressed, and I re-ran each rather than reading it.

## 1. Test timeout, the blocking ask

`@Test(timeout = 5_000)` compiles as expected: `kotlin.test.Test` is a typealias for `org.junit.Test`, so the parameter passes straight through with no new dependency.

I repeated the same experiment as the first round, reverting only the `SecretsManager.kt` guard and leaving the test in place:

| | First round | Now |
| --- | --- | --- |
| Regression outcome | ran past 150 s, killed manually | **failed in 6 s** |
| Failure reported | none, just a dead job | `TestTimedOutException: test timed out after 5000 milliseconds` |

That is exactly the behavior I was after. A future regression now goes red instead of stalling CI.

## 2. Distinct cycle error

This came out better than what I asked for. Captured stderr from the test run:

```
Folder folder-a skipped due to error: Folder data inconsistent - parent cycle detected at folder UID folder-b
Folder folder-b skipped due to error: Folder data inconsistent - parent cycle detected at folder UID folder-a
```

The interpolated `$currentParent` is the UID where the ring *closes*, not the folder being processed, because the loop exits on the first `add` returning false. So each line names two different UIDs: the folder skipped, and the node that closed the cycle. Read together the two lines let an operator reconstruct the whole ring from logs alone. I had only asked for a distinct message.

I checked the two things that could have gone wrong in turning a `return null` into a `throw`:

* `SecretsManagerException` extends `Exception` (`SecretsManagerExceptions.kt:5`), so the caller's `catch (e: Exception)` still catches it and skip-and-continue is preserved. The empty-list assertion and the stderr above both confirm it.
* There is exactly one call site (`SecretsManager.kt:1461`), inside that try/catch, and the function is `private`. Nothing else can observe the throw.

## 3. README

Folded to one line, and the wording moved from "returns `null` on re-visit" to "exits on re-visit", which the switch to a throw made necessary. Easy detail to miss when the mechanism changes underneath the prose.

## Full suite

69 tests, 0 failures, 0 errors on Zulu JDK 8 (aarch64) at `eea9968e`. The cycle test runs in 4 ms, so the 5 s budget has plenty of headroom on a slow runner.

## One optional nicety, explicitly not worth another round

The signature still reads `ByteArray?`, so a future maintainer adding a second call site could reasonably assume "returns null on failure" and skip the try/catch. A one-line KDoc noting that it throws on a cycle would close that off. Fine to leave as is, or to fold into a later change; it should not hold up this merge.

## Follow-up tickets, no action needed here

The four sibling legs are filed and linked to KSM-1270: KSM-1280 (.NET), KSM-1281 (Python), KSM-1282 (Ruby), KSM-1283 (Rust). Worth knowing before anyone ports this patch across: only Java has the KSM-1081 per-folder try/catch, so in .NET, Python and Rust the cycle guard alone would swap a hang for an aborted call that loses the well-formed folders too. Each ticket says so up front. Ruby is the only one that needs the guard alone.

sdk/java/core/src/test/kotlin/com/keepersecurity/secretsManager/core/SecretsManagerTest.kt
mgallego-keeper: Confirmed fixed, thanks.

I re-ran the same experiment with the guard reverted: the test now fails in 6 s with `TestTimedOutException: test timed out after 5000 milliseconds`, instead of running past 150 s and having to be killed. That is the whole point of the ask, so this is settled.

The 5 s budget is comfortable too, since the test itself completes in 4 ms.

PR #1119, opened 2026-08-19
KSM-1269 Run Java test matrix on release-branch PRs

stas-schaller:
## Summary

- Adds `release/sdk/java/core/**` to the `pull_request` and `push` triggers in `test.java.yml`. Previously, the Java CI matrix only fired on PRs targeting `master`; fix PRs developed against a release branch received no CI signal.
- Adding `push:` as well covers direct commits to the release branch.

## Test plan

- [ ] This PR itself triggers the Java 8/11/17/21 matrix (visible in the PR's checks tab — the fix proves itself)

### DIFF
--- a/.github/workflows/test.java.yml
+++ b/.github/workflows/test.java.yml
@@ -2,7 +2,16 @@ name: Test-Java
 
 on:
   pull_request:
-    branches: [ master ]
+    branches:
+      - master
+      - 'release/sdk/java/core/**'
+    paths:
+      - 'sdk/java/core/**'
+      - '.github/workflows/test.java.yml'
+  push:
+    branches:
+      - master
+      - 'release/sdk/java/core/**'
     paths:
       - 'sdk/java/core/**'
       - '.github/workflows/test.java.yml'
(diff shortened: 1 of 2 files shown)


mgallego-keeper (approved):
Approving. The trigger change is correct and self-verifying: all four matrix jobs (Java 8/11/17/21) ran and passed on this PR against base `release/sdk/java/core/v17.4.0`, which is exactly the case that previously got no signal.

Verified:

- `release/sdk/java/core/**` matches all 17 existing Java core release branches, including the non-versioned `graph-links-update-v17.0.1`.
- Structure matches the existing `test.ruby.yml` precedent exactly (`master` + release glob on both `pull_request` and `push`).
- The `paths` filter still covers both changed files.
- Security posture unchanged: `contents: read`, `persist-credentials: false`, actions pinned to SHAs, no secrets, no `pull_request_target`.

Two non-blocking notes.

**1. The README changelog entry.** Every other 17.4.0 entry describes a change visible to someone consuming the published Maven artifact (`generatePassword`, `isEditable`, HTTP timeouts, `getFolders()`, `deleteFolder()`, key ID validation, file permissions). A GitHub Actions trigger fix has no effect on the shipped JAR, so it reads as noise in the SDK's release notes. I would drop the README hunk; nothing parses that changelog (the Maven publish workflow does not read it, and no script in the repo references it), so removing it is risk-free. If it stays: append it after KSM-1262 rather than inserting above it, since the list is otherwise append-ordered, and keep it on one line like every other bullet.

**2. `push` on the release glob will double runs.** PR #1110 is open with head `release/sdk/java/core/v17.4.0` and base `master`. Once this merges, every push to the release branch fires both `push` (branch matches the release glob) and `pull_request: synchronize` on #1110 (base matches `master`), so 8 matrix jobs per push instead of 4. Draft status does not suppress it, since drafts still emit `synchronize`.

Not worth blocking on for a 2-minute job, and `test.ruby.yml` already carries the identical duplication, so this is consistent with the house pattern. One thing worth knowing if it ever does get addressed: a `concurrency:` block will not dedupe it. The standard group key `${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}` puts the push run in `Test-Java-refs/heads/release/...` and the PR run in `Test-Java-1110`, so both still run. The only real options are dropping the release glob from `push:` (the release PR to master is open for most of the branch's life, so `pull_request` already covers it) or accepting the duplication.

Follow-up tickets for the remaining workflows with the same gap are being filed against the audit item in KSM-1269.

PR #1122, opened 2026-08-19
test(java): close the untested paths in the 17.4.0 security fixes, and gate stderr diagnostics

mgallego-keeper:
Remaining findings from the review pass over the v17.4.0 release branch, after #1117, #1119 and #1120 landed. Four tests and one behaviour fix, one commit each; every commit builds and tests green on its own.

The common thread is that three of the security fixes in this release are guarded by tests that pass against the vulnerable code, so a revert would not be caught. Each test added here was verified to fail against the pre-fix implementation, and the specific failure message is quoted below.

## 1. The KSM-1203 shuffle had no coverage

Every case in `testGeneratePassword` requests a single character category, so all 32 characters come from one charset and the final shuffle is unobservable. Those assertions pass whether the shuffle uses `SecureRandom`, uses `Random.Default`, or is deleted outright.

`randomSample()` emits characters grouped by category, so a mixed request is the only configuration where the shuffle is visible: without one, position 0 is always lowercase. The new test asserts the category of the first character varies across 200 passwords, and that each category keeps exactly the requested count.

| | `testGeneratePasswordMixedCategoriesAreShuffled` |
|---|---|
| with the fix | passes |
| with `Collections.shuffle` removed | fails, `position 0 only ever held category [0]` |

Scope is stated in the test comment rather than implied: it does not distinguish a CSPRNG from a weak PRNG, since both yield a uniform permutation. It passes against the old `Random.Default` shuffle. That the shuffle draws from `SecureRandom` is enforced by review, not by this test.

## 2. The KSM-1262 permission test passes against the vulnerable code

`testConfigFileAtomicWriteIsOwnerOnly` asserted only that the finished config file is `0600`. The implementation it replaced also ended at `0600`, by calling chmod after writing the private key through an `0644` handle, so the test cannot detect the fix being reverted.

What KSM-1262 actually changed is that secrets are never written to the visible path. A rename installs a different inode, so `BasicFileAttributes.fileKey()` changes across a write while an in-place rewrite keeps the same value. That makes the mechanism deterministic to assert, with no race to lose. Also asserts permissions survive the swap and that no `ksm_*.tmp` is left behind.

| | `testConfigFileAtomicWriteIsOwnerOnly` |
|---|---|
| with the fix | passes |
| with in-place write plus chmod-after | fails, `Config was rewritten in place rather than swapped in via rename ... Actual: (dev=100000f,ino=130795626)` |

## 3. #1120's cause retention had no test

#1120 changed both write paths to report what the file system returned and to keep the original `IOException` as the cause. There was no test, so a later refactor can drop the cause silently.

| | `testConfigWriteFailureReportsTheUnderlyingCause` |
|---|---|
| against #1120 as merged | passes |
| with `, e` removed from the handler | fails, `The original IOException must be retained as the cause` |

Asserts the cause survives, that a missing directory is not reported as a permissions problem, and that the message names the directory it could not use.

## 4. Three stderr writes ignored `loggingEnabled`

The diagnostics added for KSM-1081 and KSM-1086 wrote to stderr unconditionally, while the throttle and key-rotation diagnostics in the same file are gated on `options.loggingEnabled`. A consumer that had turned logging off still got output from `deleteSecret`, `deleteFolder` and the `getFolders` skip path.

The folder-skip message also diverged from the four equivalent handlers in `fetchAndDecryptSecrets`, which all include `e.javaClass.simpleName`. One of them logs the identical `Folder <uid> skipped due to error:` prefix, so the file emitted near-identical lines in two formats. The class name is not decoration: the causes this path exists for include a null `data` field, whose `KotlinNullPointerException` carries no message, so the line read `skipped due to error: null`.

| | pre-fix output |
|---|---|
| `testFolderSkipDiagnosticRespectsLoggingEnabled` | fails, `loggingEnabled = false must not write to stderr. Got: Folder bad-uid skipped due to error: Input data too short to contai...` |
| `testFolderSkipDiagnosticNamesTheExceptionType` | fails, `The exception class must appear so a null message is still diagnosable` |

Both drive `getFolders` through an injected `queryFunction` with one undecryptable folder and capture stderr.

## 5. Changelog

Keeps the Breaking Changes block from #1120 verbatim, including its `copy()` and synthetic-constructor arity analysis, and adds:

- The `deleteFolder` bullet notes that no working code can be affected: the old return type required a `records` field while the backend sends `folders`, so every previous call threw `MissingFieldException` rather than returning a value.
- Ascending ticket order restored (1081, 1086, 1176, 1203, 1207, 1248, 1262).
- KSM-1203 states that existing passwords need no rotation. Character selection always drew from `SecureRandom`, so only the arrangement was affected: for `generatePassword(32, 8, 8, 8, 8)` roughly 137 of the 193 bits were never at risk, and the default `generatePassword()` was unaffected because all of its characters come from a single set. Without this the entry reads as though every password the SDK has generated is suspect.
- KSM-1207 records that the configured values apply to API requests and the file upload transport, that `downloadFile()` and `downloadThumbnail()` use the built-in defaults because their entry points take only a `KeeperFile`, and that a custom `queryFunction` owns its own timeouts.
- KSM-1176 notes `isEditable` is the last constructor parameter and `KeeperRecord` carries `@JvmOverloads`.
- KSM-1262 records that the config file is now read as UTF-8 explicitly, where it previously used the JVM default charset while always being written as UTF-8.
- KSM-1081 and KSM-1086 note the per-item diagnostics are gated on `loggingEnabled`.

## Verification

- 74 tests, 0 failures, 0 skipped. Each of the five commits builds and tests green on its own.
- Every test added here verified to fail against the relevant pre-fix implementation, with the message quoted above.
- `javap` confirms the API restorations from #1117 are intact on top of these commits: all three `postFunction` arities (4, 5, 6 args), 8 `KeeperRecord` constructors, and both `SecretsManagerException` constructors.

## Not addressed here

- `Files.move` throws `AccessDeniedException`, not `AtomicMoveNotSupportedException`, when another process holds the target open, so the existing fallback in both write paths does not catch the realistic Windows failure (a second SDK instance, or an on-access virus scanner). A short retry around the swap would cover it. Left out deliberately: `test.java.yml` runs `ubuntu-latest` only, so the retry path cannot be exercised in CI, and this PR argues against changes whose tests cannot fail.
- No `fsync` before the rename. The swap is atomic with respect to visibility but not durability; a crash can leave a truncated config on some filesystems.
- An orphan `ksm_*.tmp` is left in the config directory if the JVM is killed between `createTempFile` and the move.
- `TestStubs` lives in the main source set, so `generateTransmissionKey` consults a test hook on every call in the shipped artifact. `internal` keeps Kotlin consumers out, but a test seam that replaces the transmission key with a constant is worth moving behind a test source set for an audit release.
- `TestStubs.transmissionKeyStub` is set by eight tests and reset by none; there is no `@AfterEach`.
- Same defect as KSM-1203 is still live in the Python SDK: `sdk/python/core/keeper_secrets_manager_core/utils.py:586` is `random.shuffle` (Mersenne Twister) against `secrets.choice` selection. Confirmed the JS, .NET, Ruby and Rust SDKs all already use a CSPRNG for the shuffle.
- `sdk/javascript/packages/core/src/keeper.ts:1298` declares `deleteFolder` as returning `SecretsManagerDeleteResponse` and casts with `as`, which does not validate at runtime, so it hands back an object whose `.records` is `undefined`. Same wire-shape mismatch KSM-1086 fixed in Java; Python reads `.get('folders')` and confirms the shape.
- `binary-compatibility-validator` is still not configured. #1119 fixed the CI trigger so the JDK 8/11/17/21 matrix now runs on release-branch PRs, including this one, which removes the larger half of the gap that let the two API breaks #1117 fixed reach a fully green release branch. An ABI check would close the rest: the matrix compiles core against itself and still does not compile the Java consumers in `integration/` or `examples/`.


stas-schaller (approved):
👍👍

PR #1123, opened 2026-08-19
fix(java): proxy credential validation and timeout reporting follow-ups (KSM-531, KSM-1207)

mgallego-keeper:
## Summary
Four follow-ups from the code review of #1110, across the KSM-531 proxy path and the KSM-1207 timeout work. Two are behavior fixes, two close gaps that let a shipped change go unreported or drift.

## 1. Incomplete proxy credentials are rejected (KSM-531, behavior fix)

`parseProxy` accepted `http://user:@host` as a valid credential and registered an empty-password `Authenticator`. The guard asked whether exactly one half of the userinfo was missing:

```kotlin
(username == null) != (password == null)
```

which structurally cannot catch a half that is present but empty. An empty password now nulls out during parsing exactly as an empty username already did, so both halves share one notion of absence and the guard reduces to "either half missing".

| Input | Before | After |
|---|---|---|
| `http://user:pass@host` | accepted | accepted |
| `http://user@host` | rejected | rejected |
| `http://:secret@host` | rejected | rejected |
| `http://user:@host` | **accepted, empty password** | rejected |
| ambient `http://user:@host` | `hasCredentials = true` | `hasCredentials = false` |

Why it matters: a templated proxy URL whose password variable goes unset (`http://$USER:$PASS@host`) renders as the fourth form. Before this change it reached the proxy and returned a 407, which sent the operator down the `jdk.http.auth.tunneling.disabledSchemes` remediation path for what is actually a configuration typo. That remediation message is long and specific, and pointing it at the wrong problem is worse than no message.

Because `hasCredentials` derives from both fields, the parse fix also corrects the `Authenticator` registration, the ambient-vs-explicit branch in the 407 message, and the `checkedResponseCode` fast path, with no call site changes.

## 2. Process-wide scope of the tunneling flag is documented (KSM-531, docs)

`enableBasicProxyAuthOverTunnel()` clears `jdk.http.auth.tunneling.disabledSchemes`, a JVM system property. Clearing it re-enables Basic auth over CONNECT for every `java.net.HttpURLConnection` in the process, not only the SDK's. The 17.4.0 changelog said the SDK "does clear it automatically" without saying whose connections that affects.

The SDK's own write is already narrow: it happens only when credentials are supplied in an explicit `proxyUrl`, and a value the host application already set is left untouched. That scoping is worth stating rather than leaving readers to infer, because an application sharing a JVM with other HTTP clients should treat re-enabling a CVE-2016-5597 mitigation bypass as a deliberate decision rather than one inherited from a dependency.

Documented in two places, since the code and the changelog have different readers: the KDoc on `enableBasicProxyAuthOverTunnel()`, and a new changelog bullet, "Clearing that flag is process-wide, not SDK-scoped". No behavior change in this half.

## 3. Timeouts now appear in `SecretsManagerOptions.toString()` (KSM-1207)

`toString()` is hand-written rather than data-class generated, so it has to be edited whenever a field is added. It was not: `connectTimeoutMillis` and `readTimeoutMillis` were missing from the rendered output, which is exactly the output someone reads when diagnosing a timeout. Both now appear.

## 4. Timeout defaults share one constant (KSM-1207)

The defaults were declared twice: as bare literals in `SecretsManagerOptions`, and again as `DEFAULT_CONNECT_TIMEOUT_MS` / `DEFAULT_READ_TIMEOUT_MS` for the overloads that take a `KeeperFile` without options. `TimeoutTest` asserted only the data-class side, so the pair could drift silently and a caller of `downloadFile(file)` would get a different timeout than a caller of `downloadFile(options, file)` with no test objecting. The constants are now the single source, moved up beside the other connection tunables.

This widens the reach of the existing test rather than needing a new one: with the literals gone, `timeoutDefaults_matchDocumentedValues` fails if the constant changes. Verified by drifting `DEFAULT_CONNECT_TIMEOUT_MS` to `6_000`, confirming the test fails, and reverting.

## Tests
- `partialUserinfoExplicitProxyThrows` now covers all three rejected credential forms rather than only `http://user@host`.
- New `emptyPasswordAmbientProxyIsNotTreatedAsCredentials` pins the ambient path, which must not throw but also must not register a credential.
- New `toString_reportsTimeoutsAndRedactsProxyUrl`. Nothing tested `toString()` before, so the `proxyUrl` redaction shipped in this release was unguarded; this covers it alongside the two timeout fields. It uses sentinel credentials deliberately: the rendered `storage` field carries the package name, so asserting on a realistic password like `secret` matches `com.keepersecurity.secretsManager` and passes for the wrong reason.
- Full suite: 102 tests, 0 failures, verified locally on a real JDK 8 (Zulu 8.0.492). No new compiler warnings.

## Not included
The remaining review findings are left for a later pass, since none affect shipped behavior: the inert `@JvmOverloads` on the 4-argument `postFunction` (a live compiler warning), the unconditional `cachingPostFunction` stderr warning already tracked as KSM-1298, `ProxyTest` leaving the tunneling property and a registered credential set for later tests in the same JVM, the doubled Java CI matrix now that both `push` and `pull_request` fire on release branches, and the `android-example` README still pinning `17.1.2`.

## Related Issues
- Follow-up to #1110 (Release Java SDK v17.4.0)
- KSM-531, KSM-1207


stas-schaller (approved):
👍👍👍

PR #1131, opened 2026-08-26
JavaScript SDK: fix config file permissions not re-applied on every write (KSM-1263)

stas-schaller:
## Summary
JavaScript SDK: fixes config and cache file permissions not being re-applied on every write.

### Fixed
- `fs.openSync`'s mode argument only takes effect when the file is created, so a config or cache file that already existed with looser permissions kept them. Permissions are now explicitly reset to 0600 after every write. (KSM-1263)

## Testing
cd sdk/javascript/packages/core && npm test -- test/localConfigStorage.test.ts

## Breaking Changes
None.

## Related Issues
- Jira: KSM-1263


mgallego-keeper (commented):
## Review summary

Small, clean, correctly targeted fix. `fs.openSync`'s mode argument is indeed only honored at file creation (confirmed independently), so re-asserting 0600 via `chmodSync` after every write is the right approach, and the new test correctly exercises it (pre-create the file at 0644, write through `localConfigStorage`, assert the mode comes back to 0600). Two notes below, neither blocking.

## Notes

### CHANGELOG entry describes a fix broader than this diff delivers

The new CHANGELOG line says "Fixed config and cache file permissions not being re-applied on every write," but this PR's diff only adds `chmodSecure` to the config file's `saveStorage`. The old `cachingPostFunction` still writes `cache.dat` via a plain `fs.openSync('cache.dat', 'w', 0o600)` with no re-assertion at this point in the stack; that only gets fixed once KSM-1265 lands. If this PR merges on its own before the follow-on PR, the changelog would be claiming a cache-file fix that isn't actually in the code yet. Worth tightening the wording here to "config file" only, or holding the broader claim for the PR that actually touches the cache path.

### CI will not run for this PR

`.github/workflows/test.js.yml` triggers only on `pull_request: branches: [master]`. This PR targets `release/sdk/javascript/core/v17.6.0`, so that workflow will not fire (confirmed via the PR's checks: only Socket Security ran). The new test does pass when run manually (`npm test -- test/localConfigStorage.test.ts`), so there's no evidence of an actual failure, just no automated proof for any PR in this stack. For reference, `test.javascript.storage.*-kms.yml` and `test.ruby.yml` already include their respective `release/**` branches in their triggers; `test.js.yml` is missing the equivalent `release/sdk/javascript/core/**` pattern. Flagging it here since this PR is the one anchored directly to the release branch; the same gap will carry through the rest of this stack until the workflow trigger is fixed.

### Portability note

The fix and its test rely on POSIX permission bits. On Windows, `fs.chmodSync` only toggles the read-only attribute and does not provide the same owner-only guarantee, so both the protection and the test's assertion are effectively POSIX-only. Not a defect in this PR, just worth being aware of if Windows is a supported target for this package.

PR #1135, opened 2026-08-26
JavaScript SDK: fix Node platform hash() ignoring tag parameter (KSM-1254)

stas-schaller:
## Summary
JavaScript SDK: fixes the Node platform's `hash()` ignoring its `tag` parameter and always hashing with a hardcoded string.

### Fixed
- Node platform's `hash()` hardcoded `'KEEPER_SECRETS_MANAGER_CLIENT_ID'` as the HMAC input instead of using its `tag` parameter, unlike the browser implementation and the `Platform` contract. It now hashes with the caller-supplied tag. (KSM-1254)

## Testing
cd sdk/javascript/packages/core && npm test -- test/nodePlatform.test.ts

## Breaking Changes
None.

## Related Issues
- Jira: KSM-1254


mgallego-keeper (approved):
## Summary

Clean, minimal, correct fix. `nodePlatform.ts`'s `hash()` (line 169) now uses its `tag` parameter instead of a hardcoded string, matching the browser implementation and the `Platform` contract. Ran the suite against this exact branch: 7 suites, 64 tests, all passing.

## Correctness

Confirmed non-breaking today: both real production call sites (`getClientId` and the client-key-hash path in `keeper.ts`) pass the same `CLIENT_ID_HASH_TAG` constant that was previously hardcoded inside the buggy implementation, so output is identical before and after this fix for the SDK's current usage.

## One coverage nuance (non-blocking)

The two new unit tests in `test/nodePlatform.test.ts` correctly exercise `hash(data, tag)` directly with two distinct tags, which proves the fix works. But since both real call sites in `keeper.ts` always pass the same tag, no integration-level test can distinguish the old tag-ignoring bug from this fix, only the new isolated unit test does. Worth keeping in mind if a second call site with a different tag is ever introduced (the entire reason `hash()` takes a `tag` parameter): a future regression there would only be caught by this direct unit test, nothing higher up the stack.

## Process note

This PR targets `release/sdk/javascript/core/v17.6.0` rather than `master`, so `test.js.yml` (the only workflow that runs the JS SDK test suite) does not trigger on it. The only passing checks are Socket Security scans. Ran the suite locally against this branch to confirm health: 7 suites, 64 tests, all green.

Approving.

mgallego-keeper (approved):

PR #1141, opened 2026-08-28
JavaScript SDK: fix browser IndexedDB storage hanging forever on a storage failure (KSM-1332)

mgallego-keeper:
## Summary
JavaScript SDK: every IndexedDB call in the browser config storage wired only `onsuccess`, so any storage failure left the caller's promise pending forever with no error, no rejection and no timeout.

### Fixed
- `src/browser/localConfigStorage.ts` wired only `onsuccess` (plus `onupgradeneeded` on the two `indexedDB.open` calls) across all eight promise wrappers in `localConfigStorage` and `secureStorage`. A failing `IDBRequest` fires `onerror` and never `onsuccess`, so a failure left the promise pending forever. Four of the eight even destructured `reject` from the Promise executor and never called it. All eight now reject with a typed `KeeperError`. (KSM-1332)

Three failure paths were unhandled, and all three are covered:

| Path | Fires | Previously |
| --- | --- | --- |
| Request failure (quota, corrupt DB, blocked site data, private browsing) | `onerror` | promise never settles |
| Blocked version change, another live connection holds the database | `onblocked` | promise never settles |
| Missing object store | `transaction()` throws synchronously inside `onsuccess` | promise never settles |

The third one is easy to miss: a throw inside an event handler is not caught by the enclosing `Promise` executor, so it leaks as an uncaught exception and the promise stays pending.

`secureStorage` was the worst case. It awaits a read of `__secureKey__` before it returns the storage object, so a failure hung the constructor itself and never handed the caller anything to catch.

The cause is duck-typed on `name` and `message` rather than tested with `instanceof Error`. A `DOMException` only satisfies `instanceof Error` in the realm it was constructed in, so a cross-realm failure (a worker, an iframe) would otherwise report `unknown error` and lose the diagnostic entirely.

## Testing
```
```
9 suites, 76 tests, all passing. `tsc --noEmit` clean.

Adds `test/browserConfigStorage.test.ts`: 10 tests covering both exported storages, driven by a minimal IndexedDB double, with **no new dependency**. Each failure test races a 500ms timer, so a regression fails fast with `promise never settled` rather than stalling the suite until Jest's default timeout.

Verified against the unfixed file, which is what makes the suite meaningful here:

```
✕ rejects when opening the database fails            (506 ms)
✕ rejects when the open is blocked by another conn.  (502 ms)
✕ rejects when the object store is missing           (503 ms)
✕ rejects when a read fails                          (501 ms)
✕ rejects when a write fails                         (503 ms)
✕ rejects when a delete fails                        (503 ms)
✓ still reads back what it wrote when healthy        ( 13 ms)
✕ secureStorage: rejects when opening fails          (504 ms)
✕ secureStorage: rejects when a read fails           (503 ms)
✓ secureStorage: still round-trips through encryption ( 18 ms)
```

All eight failure tests fail on the old code, each at roughly 500ms, which is the hang being caught by the timer. Both happy-path tests pass on either side, which is the evidence that the healthy path is unchanged.

## Security Impact
No change to what is stored, encrypted, or transmitted. `secureStorage`'s AES-GCM wrapping key handling is untouched. This is availability only: a browser client that previously hung indefinitely on a storage failure now fails with an error the caller can catch, log, or retry. Same defect class as KSM-1209 (no request timeout on network I/O), on the storage path rather than the network path.

## Breaking Changes
None for a working IndexedDB. Callers that previously hung now receive a rejected promise, which is the intended fix; any caller that awaited these methods without a `catch` will surface an error where it previously waited forever.

## Merge notes
Merge-tested against the other open branches targeting this release branch:

- `src/browser/localConfigStorage.ts` **auto-merges cleanly** with #1139, which edits `createCachingFunction` further down the same file.
- The `CHANGELOG.md` conflict with #1131 and #1139, and the `test/nodePlatform.test.ts` add/add conflict with #1139, are **pre-existing**. Both already conflict with the release-branch tip at `1ffba2b5` without this commit. This PR adds no new conflict.

## Known adjacent issue, not fixed here
Each `getObjectStore` call opens a new `IDBDatabase` connection and never calls `close()`. Those leaked connections are exactly what causes `onblocked` for a later version change. Fixing the lifecycle changes resolve timing and is out of scope for this ticket; worth a follow-up.

## Related Issues
- Jira: KSM-1332
- Related: KSM-1209 (same hang class, network path), KSM-1265, KSM-1266

### DIFF
--- a/sdk/javascript/packages/core/CHANGELOG.md
+++ b/sdk/javascript/packages/core/CHANGELOG.md
@@ -8,6 +8,7 @@
 - KSM-1035 - Fixed throttle retry jitter being two-sided, which could reduce a retry delay below the computed floor. Jitter is now one-sided (0 to +25%). The SDK also caps a server-supplied `retry_after` at 176s to prevent an arbitrarily long wait.
 - KSM-1128 - Bounded the server key-rotation retry in `postQuery`. When the server sends `{"error":"key"}`, the code retries at most 3 times before throwing a typed `KeeperError`, instead of retrying forever. Before storing a suggested `key_id`, the code validates its shape (positive integer) and its membership in the bundled key table (keys 7-18). An unsupported key id can no longer corrupt the configuration. The pinned custom-key path does not change.
 - KSM-1254 - Fixed the Node platform's `hash()` ignoring its `tag` parameter and always hashing with a hardcoded string; it now hashes with the caller-supplied tag, matching the browser implementation and the `Platform` contract. No behavior change for existing callers (the SDK's only caller already passed that same string).
+- KSM-1332 - Fixed the browser IndexedDB storage hanging forever on a storage failure. `localConfigStorage` and `secureStorage` wired only `onsuccess`, so a failed IndexedDB open, read, write or delete left the promise pending with no error, no rejection and no timeout. All eight wrappers now reject with a typed `KeeperError`, and the blocked-upgrade and missing-object-store paths reject too instead of hanging.
 - Maintenance: Updated `minimatch`, `@babel/core`, and `handlebars` dev dependencies.
 
 ## 17.5.0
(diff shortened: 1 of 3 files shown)


stas-schaller (approved):
good stuff

PR #1144, opened 2026-08-31
JavaScript SDK: run throttle detection on the full response body, not the truncated slice (KSM-1251)

stas-schaller:
## Summary
Fixes throttle (HTTP 403) and server key-rotation detection running on an error body truncated to 1000 bytes, which silently disabled the KSM-880 throttle retry and key-rotation for any response padded past that size.

### Fixed
- `postQuery` now runs throttle and key-rotation detection on the full decoded response body (capped at 64KB via `MAX_ERROR_BODY_DECODE_BYTES`, generous headroom over any real throttle/key-rotation body) instead of a 1000-byte slice. Previously, an oversized body was cut mid-document, `JSON.parse` threw on the fragment, and the response fell through to a generic error with zero retries or key rotation. (KSM-1251)
- The message in the error the SDK throws when retries/rotation are exhausted, or the response isn't recognized, is truncated by byte count on the raw response bytes before decoding — not by character count on the already-decoded string, which could cut a multi-byte UTF-16 surrogate pair at the 1000-character boundary for non-ASCII bodies.
- The key-rotation check reads `result_code ?? error`, matching `parseThrottle` and the equivalent Python/Ruby/Rust handlers, instead of only `error`.
- A non-200 response with a genuinely empty body is treated identically on Node and browser: Node's `fetchData` leaves `data: null` for an empty body, but the browser platform always constructs a `Uint8Array`, so an empty body there is still a truthy, zero-length object. The check now verifies `response.data && response.data.length > 0` on both platforms, falling through to the `"unknown ksm error, code X"` message instead of a blank one.

## Testing
```
npm test
```
Key scenarios covered by new tests, all confirmed failing on the pre-fix code before the fix landed:
- A throttle (403) or key-rotation (400) body padded past the 1000-byte truncation slice is correctly detected and acted on, not treated as unrecognized.
- A body over the 64KB decode cap fails to parse and falls through to the catch-all error, rather than being misread as a valid throttle/rotation signal.
- Key rotation triggers correctly via `result_code` alone, and `result_code` takes precedence over a stale `error: 'key'` field when both are present.
- The catch-all throw truncates its message to exactly 1000 bytes on the raw response bytes, not the decoded string.
- A zero-length response body on a non-200 response falls through to the generic `"unknown ksm error"` message on both platforms, not a blank error.

Full suite 122/122 passing, `tsc --noEmit` clean.

## Breaking Changes
None.

## Related Issues
- Jira: KSM-1251
- This branch also carries KSM-1252 (PR #1145, "strengthen throttle e2e assertions"), merged in as a stacked commit ahead of this fix. The two tickets need to land on the release branch as separate commits, not squashed into one, when this merges.


mgallego-keeper (commented):
This is a thorough pass focused on regressions and test coverage for the KSM-1251 fix.

**Summary:** the throttle-detection fix itself is correct, but the identical truncation-before-parse defect this PR fixes is still live in the sibling key-rotation branch, with no test coverage there either. Two other behavior changes are worth a second look before this ships.

**Key-rotation branch still parses the truncated string** (`keeper.ts:825`)
The key-rotation check three lines below the throttle fix still runs `JSON.parse` on the truncated `errorMessage` instead of the new `fullErrorMessage`. A key-rotation response body over 1000 characters will fail to parse and fall through to a generic error instead of rotating the key. Suggest pointing line 825 at `fullErrorMessage`, matching the throttle branch above it.

**No test for the key-rotation truncation path** (`test/keeper.test.ts`)
The throttle fix got a dedicated oversized-body regression test in `throttle.test.ts`; the key-rotation path did not. A parallel test with an oversized `error: key` body would catch the issue above.

**Full response body is now unconditionally decoded** (`keeper.ts:806`)
Previously the body was sliced to at most 1000 bytes before decoding. Now the full body is decoded first, then truncated. On a pathologically large error body this can throw `RangeError: Invalid string length` or cause a memory and CPU spike, a failure mode the prior code structurally didn't have. Worth confirming this tradeoff is intentional, this path runs on every failing request.

**Truncation changed from byte-slice to UTF-16 char-slice** (`keeper.ts:807`)
`response.data.slice(0, 1000)` (bytes) became `fullErrorMessage.slice(0, 1000)` (UTF-16 code units). For non-ASCII error bodies this changes what "1000" bounds, and can cut a surrogate pair at the boundary. The new padded-body test only uses ASCII padding, so it can't catch this class of issue.

**Smaller items:**
- The key-rotation check (`errorObj?.error === 'key'`, line 826) reads only `error`, never `result_code`, unlike `parseThrottle` just above it (`result_code ?? error`) and unlike the equivalent Python, Ruby, and Rust handlers. Possibly worth aligning while this block is already open.
- The comment above line 805 (and the matching CHANGELOG entry) states truncation "is applied only to what actually gets thrown below," which isn't accurate since it also drives the key-rotation parse decision.
- The new test hand-rolls a response literal instead of extending the existing `throttle403` helper to accept extra fields.

mgallego-keeper (commented):
A few things worth addressing before merge, plus some non-blocking cleanup below.

At `sdk/javascript/packages/core/test/keeper.test.ts:15`, `keyErrorResponse()` always sends `error: 'key'` and never sets `result_code`, so the new `result_code ?? error` fallback in the key-rotation branch (`keeper.ts:850`) has zero coverage either way. Could you add a case that sends `{result_code: 'key'}` with no `error` field (confirms the new path works), and maybe also one with `result_code` present-but-not-`'key'` alongside `error: 'key'` (documents the current behavior either way, whichever you decide is correct)?

At `sdk/javascript/packages/core/src/keeper.ts:827`, the new `MAX_ERROR_BODY_DECODE_BYTES` (64KB) cap has no test anywhere near its boundary; the largest body in the suite is ~1100 bytes. A one-line test with a body just over 64KB (asserting it still throws sensibly rather than hanging/OOMing) would cover this.

Similarly, at `sdk/javascript/packages/core/src/keeper.ts:874`, no test asserts the content of the final catch-all throw for a large unrecognized body, so we're not actually verifying the 1000-byte truncation happens here (vs., say, accidentally throwing the full decoded body).

The changelog entry at `sdk/javascript/packages/core/CHANGELOG.md:14` is also worth fixing, it'll confuse whoever debugs this next: it says the 1000-byte truncation "applies only to the message... when retries/rotation are exhausted," but the exhaustion throws (`KeeperThrottleError`, `KeeperError` at lines 839/865) are fully synthetic strings with no body content at all. The truncation only ever feeds the separate "response isn't recognized" throw at line 874.

---

**Non-blocking / cleanup**

- At `keeper.ts:22`, the comment above `MAX_KEY_ROTATION_RETRIES` still describes the branch as `error === 'key'` only, now stale since the condition includes `result_code`.
- At `keeper.ts:849` (compare `keeper.ts:101`), the `result_code ?? error` fallback is now hand-duplicated in two places (`parseThrottle` and this branch); might be worth a shared helper so a future envelope change only needs one edit.
- At `keeper.ts:831`, `errorMessage` is computed unconditionally on every non-200 response but only read on the final throw; wasted work on the (common) retry/rotation paths that `continue`.
- At `keeper.ts:827`, `Uint8Array.prototype.slice` clamps out-of-range ends rather than throwing, so the ternary could just be an unconditional `slice(0, MAX_ERROR_BODY_DECODE_BYTES)`.
- Both `test/keeper.test.ts:15` and `test/throttle.test.ts:20` independently added a near-identical `...extra` spread for padding a body; the spread-last ordering also means a colliding key (e.g. `key_id`) would silently override the named field with no type error. Low risk today since only `{padding}` is ever passed, but worth a shared fixture helper if a third case shows up.
- At `keeper.ts:31`, there are two independent byte caps (1000, 65536) with no comment tying them together; a one-liner on why they differ by 65x would help the next reader.

mgallego-keeper (commented):
Re-reviewed after the latest push. The result_code, test-gap, and changelog items from the previous round all look resolved now, thanks for adding coverage for key rotation via result_code, the 64KB boundary, and the catch-all message.

A handful of smaller things worth a look before this merges:

**1. `keeper.ts` line 828, browser platform empty-body handling**
On browser, a non-200 response with a genuinely empty body ends up as a truthy zero-length `Uint8Array` (`browserPlatform.ts` always does `data: new Uint8Array(body)`, even for a 0-byte `ArrayBuffer`), while `nodePlatform.ts` leaves `data: null` when no bytes arrive. That means `if (response.data)` takes the has-data branch on browser but not on Node for the same empty-body response, and on browser it ends up throwing `Error('')` instead of the "unknown ksm error, code X" fallback. Confirmed by mocking a 500 with an empty body. Might be worth checking `response.data?.length` (or aligning the two platforms so an empty body is represented the same way in both).

**2. `keeper.ts` line 35, MAX_ERROR_BODY_DECODE_BYTES comment**
The comment is correct that this bounds the decode step, but since the full response body is already buffered into memory unconditionally on both platforms before this cap ever runs, a reader could come away thinking this caps the cost of a large body more broadly than it actually does. Might be worth a one-line caveat.

**3. `keeper.test.ts` lines 373 and 447**
These two key-rotation tests don't have the `calls > 50` runaway-loop guard that the other loop-driving tests in this file already use (lines 325, 345). Cheap insurance against a hang in CI if a future change breaks loop termination.

**4. `platform.ts` line 58**
`KeeperHttpResponse.data` is typed as non-nullable `Uint8Array`, but `nodePlatform.ts` can set it to `null` at runtime for an empty body. This is the same underlying gap as point 1. Typing it `Uint8Array | null` would let the compiler catch this instead of relying on every caller remembering.

**5. `throttle.test.ts`**
The new 64KB decode-cap boundary test only covers the key-rotation (400) path. The throttle (403) path runs through the same cap in `postQuery` but doesn't have an equivalent test.

Nothing blocking, but 1 and 4 look like a real (if narrow) behavior gap on the browser platform specifically, worth fixing given the SDK targets both platforms.

mgallego-keeper (commented):
Re-reviewed after the latest push (commit 28bb6972).

Four of the five items from the last review are fixed now, thank you for the quick turnaround.

- The browser and Node empty-body mismatch is fixed. The `.length` check works correctly.
- The `MAX_ERROR_BODY_DECODE_BYTES` comment is accurate now.
- Both key-rotation tests have the `calls > 50` safety guard now.
- The throttle path has a 64KB boundary test now, matching the key-rotation path.

One new item from this fix itself:

**The empty-body fix has no regression test.** I confirmed this by reverting the `.length` check and rerunning the full suite. All 121 tests still passed. A future change could reintroduce this exact bug with CI staying green. A short test mocking a zero-length body on a non-200 response would close this gap.

Two items from the prior review are still open, no change needed yet, just a reminder:

- `platform.ts` line 58: `KeeperHttpResponse.data` is still typed as non-nullable `Uint8Array`, but `nodePlatform.ts` can still set it to `null` at runtime.
- `keeper.ts` line 838: the unconditional `.slice(0, MAX_ERROR_BODY_DECODE_BYTES)` still copies the array on every non-200 response on browser, even for a small body.

Small note, not blocking: the shared padding-body test helper deferred in the last commit message ("wait for a third occurrence") could go in now. The count is at four occurrences across `keeper.test.ts` and `throttle.test.ts`.

mgallego-keeper (commented):
Round 4, on commit `68f066ba`.

Comment, not a change request. Round 3 was also a comment, and nothing in this PR's own diff is wrong. The one commit since round 3 closes the round-3 ask, and I verified that by mutation rather than by reading. Every remaining item is a one-line test addition, a wording fix, or a gap that predates this PR.

## Where this stands

Your read that this is "no major changes, should be simple" holds, and the evidence supports it. The whole change since round 3 is 15 added lines in test/keeper.test.ts, with no source change at all. Of the 26 findings I posted across rounds 1 to 4, 22 are now fixed. The 4 still open are all low or nit, and 3 of the 4 are pre-existing on the base branch rather than caused by this PR. The PR is a strict improvement over f7c33bc1 on the key-rotation path, which I confirmed end to end: a padded key-rotation body now rotates where the base branch threw. I recommend merging on your schedule and tracking the separate-ticket items outside this PR.

## The new test

The single new test closes the round-3 headline finding, and it is load-bearing. I verified this by mutation rather than by reading. Baseline at head 68f066ba is 14 suites and 122 tests passing. I reverted the guard at src/keeper.ts:830 from `if (response.data && response.data.length > 0)` back to the base branch's `if (response.data)`, rebuilt with rollup, and ran the whole suite: 1 suite failed, 1 test failed of 122, and the failure was exactly "a zero-length response body on a non-200 response falls through to the generic error, not a blank one", with Received message: "". So the round-3 ask ("a short test mocking a zero-length body on a non-200 response") is answered exactly as written, and the fix is now genuinely pinned. Test count moved 121 to 122, matching the one added test. The one real limit: the test pins only the browser shape. It injects `new Uint8Array(0)`, which covers the `.length > 0` half this PR adds. It does not cover Node's `data: null` shape, which the older truthiness half protects. I confirmed the consequence: reducing the guard to `if (response.data.length > 0)` leaves 122 of 122 passing and `npx tsc --noEmit` at exit 0, and a probe then shows a 400 with `data: null` throwing TypeError: Cannot read properties of null (reading 'length') instead of "unknown ksm error, code 400". That gap is pre-existing, not created here: base f7c33bc1 line 816 already read `if (response.data)`, and `grep -c "data: null"` over the base test file returns 0. So the honest verdict is: the new test is correct, minimal, effective, and answers the request; one adjacent shape stays unasserted, and closing it costs one line (NB2).

## Checked against the backend contract

Two things I checked against the backend source, both of which confirm your design.

First, a non-200 error body is plaintext UTF-8 JSON, not AES-GCM ciphertext. A key rejection means the server never recovered the transmission key, so it cannot encrypt the reply. Running `platform.bytesToString` and `JSON.parse` on the raw bytes is therefore correct, and the SDK calls `decryptWithKey` only on the 200 branch. Six SDK implementations all parse the error body without decrypting.

Second, `MAX_ERROR_BODY_DECODE_BYTES` at 65536 has roughly 1000 times the headroom it needs. The real key-rotation body is about 26 bytes. The cap is generous rather than tight, which is the right choice here.

## Merge order

This PR is the lowest-risk of the five to land, and its position in the merge order does not matter. I verified that in a real trial merge. It conflicts with PR 1132, PR 1133 and PR 1157 in `sdk/javascript/packages/core/CHANGELOG.md` only, and it merges cleanly against PR 1136 in every file. `src/keeper.ts` and `test/keeper.test.ts` auto-merge at every step, and the auto-merge result is correct: it keeps this PR's `resultCode(errorObj) === 'key'` alongside PR 1157's own change to the same branch.

Each CHANGELOG conflict is a two-entry keep-both resolution, so please do not move your entry to avoid it. The class fix is repo-level and genuinely missing: there is no `.gitattributes` at the repo root or under `sdk/javascript`, so a union merge attribute on that CHANGELOG would remove this conflict class for every PR.

## What is posted where

8 non-blocking items are inline on the lines they concern. 1 does not anchor to a line this PR changed, so it is below.

<details>
<summary>Non-blocking item with no anchor in this diff</summary>

**NB4 (low): Key-rotation throws now carry up to about 65500 characters of server-controlled text, 65x more than the base branch**
`sdk/javascript/packages/core/src/keeper.ts` line 866

Moving JSON.parse to the 64KB slice also raised the size ceiling on two body-derived messages.

Measured on this head: a 400 body of {"error":"key","key_id":"Z" repeated 60000 times} is 60027 bytes, and the thrown KeeperError message is 60053 characters. The same body against the merge base 9a4ebfe3 throws a plain Error of exactly 1000 characters, because the 1000-byte slice cut the JSON and the parse failed.

So the throws at src/keeper.ts:863 and src/keeper.ts:866 now carry up to about 65500 characters of server-chosen text. Any caller that logs error.message writes that per failed request.

This needs a hostile or compromised backend, so the cost is log volume, not confidentiality. Severity is low.

Suggested fix: bound the value before either message, for example `const shownKeyId = JSON.stringify(suggestedKeyId).slice(0, 200)`, then use shownKeyId at both sites.

Note src/keeper.ts:863 interpolates the value bare today, so a newline in key_id starts a new log line. That bare interpolation is pre-existing and unchanged from base, but the same fix closes it.

One more clause: the comment at src/keeper.ts:834 and the CHANGELOG both list which throws carry body content. Neither names these two.

</details>

<details>
<summary>Previously posted findings and their status at this head (4 items)</summary>

- **STILL_OPEN, unchanged, and pre-existing on the base branch (git show ...** platform.ts line 58 types KeeperHttpResponse.data as non-nullable Uint8Array, while src/node/nodePlatform.ts line 206 sets it to null for an empty body.
  Recommendation: File a ticket, do not re-flag on this PR. I measured the fix cost: widening the field to `Uint8Array | null` produces exactly 4 tsc errors, and postQuery produces none, because its `response.data &&` guards narrow correctly. The 4 sites are src/keeper.ts:1499 ...
- **STILL_OPEN in the code, but the round-4 wording was wrong and should be corrected rather ...** keeper.ts line 838 slices unconditionally, copying the array on every non-200 response even for a small body.
  Recommendation: Close it as won't-fix and correct the record. Two facts undercut it. First, the copy is bounded: slice clamps the end index, so it copies min(body length, 65536) bytes, and a small body copies only its own size. Second, the "browser only" framing does not ...
- **STILL_OPEN, no helper added. The real count is 5, not 4, and it has not grown since round ...** The shared padding-body test helper the author deferred until a third occurrence could go in now; the count is at four occurrences.
  Recommendation: Mention it once more inside NB10 rather than as its own item, and correct the count to 5 so the author is not confused by two different numbers. If the author declines again, close it. The 5 sites are test/keeper.test.ts:422, test/keeper.test.ts:439, ...
- **STILL_OPEN, unchanged. The override is now load-bearing at test/keeper.test.ts:387, which ...** Both test/keeper.test.ts:15 and test/throttle.test.ts:20 added a near-identical `...extra` spread, and spread-last lets a colliding key silently override a named field.
  Recommendation: Re-flag once, as NB10, because there is genuinely new detail worth giving the author: the throttle403 helper is the sharper case, and a colliding retry_after or error there would silently gut a throttle test while still compiling. Do not ask for a merged ...

</details>

<details>
<summary>Candidates for their own tickets (3 items)</summary>

- **cachingPostFunction masks a TypeError as "Cached value does not exist" on a bodyless 200, and can truncate a good cache ...** This is the valuable half of the platform.ts nullable-type carry-over, and it is a real defect rather than a typing nit. Node's fetchData leaves data: null for an empty body, and update_secret, finalize_secret_update, ... Existing ticket: adjacent to KSM-1265 (PR 1133, the caching-security PR), but that PR does not close this
- **nodePlatform.fetchData accumulates the response body with a per-chunk Buffer.concat, which is quadratic** src/node/nodePlatform.ts lines 202 to 216 do `retVal.data = retVal.data ? Buffer.concat([retVal.data, data]) : data` on every data event, with no size limit and no Content-Length check. Measured against a local HTTPS ... Existing ticket: none found
- **PR 1136 rebase: retarget test/cachingFunctions.test.ts after PR 1133 renames the caching export** PR 1136 adds test/cachingFunctions.test.ts importing cachingPostFunction, and PR 1133 replaces that export with createCachingFunction. Because the file is new on one side only, git auto-merges it with no conflict ... Existing ticket: KSM-1209 (PR 1136) and KSM-1265 (PR 1133); belongs on the PR 1136 rebase checklist

</details>

## Test status

Run by me at head 68f066ba in a clean checkout of sdk/javascript/packages/core, not quoted from the PR body.

`npm test` (runs the rollup pretest build, then jest): Test Suites 14 passed, 14 total. Tests 122 passed, 122 total. Snapshots 0 total. Time 3.789 s. Zero failures.

`npx tsc --noEmit`: no output, exit code 0.

Round-3 head 28bb6972 was 121 tests, so the delta is exactly the one added test. Scope check: pr1144-since-round3.diff is 26 lines and touches only sdk/javascript/packages/core/test/keeper.test.ts, confirming no src change since round 3.

Mutation runs, each done in a disposable copy made with mktemp, with node_modules symlinked to the shared install, rebuilt with `npx rollup -c --bundleConfigAsCjs`, and src restored and byte-verified afterwards:
- Guard reverted to base `if (response.data)`: 1 suite failed, 1 test failed of 122, and the failure is the new test. tsc exit 0.
- Guard reduced to `if (response.data.length > 0)`: 14 suites, 122 of 122 passed. tsc exit 0. Mutant survives.
- Decode-then-slice at src/keeper.ts:838 and 839: 122 of 122 passed. tsc exit 0. Mutant survives.
- Cap changed to `MAX_ERROR_BODY_DECODE_BYTES - 1`: 122 of 122 passed. tsc exit 0. Mutant survives.
- Catch-all changed to a character slice: 122 of 122 passed. tsc exit 0. Mutant survives.
- resultCode `??` changed to `||`: 122 of 122 passed. tsc exit 0. Mutant survives.
- KeeperHttpResponse.data widened to `Uint8Array | null`: exactly 4 tsc errors, none in postQuery (src/keeper.ts:1499, src/keeper.ts:1504, src/node/localConfigStorage.ts:69, src/browser/localConfigStorage.ts:215).

Base comparison: with these test files and the merge-base 9a4ebfe3 src/keeper.ts, the suite gives 5 failed and 117 passed of 122.

sdk/javascript/packages/core/test/keeper.test.ts
```diff
@@ -369,6 +370,127 @@ test('key rotation - suggested key id is adopted and persisted', async () => {
...
+        storage,
+        queryFunction: async () => ({ statusCode: 400, data: enc.encode(body), headers: [] })
+    }
+    const error: any = await getSecrets(options).catch(e => e)
+    expect(error.message.length).toBe(1000)
+    expect(error.message).not.toContain(marker)
+    expect(error.message).toContain('"error":"not_recognized"')
+})
+
+test('a zero-length response body on a non-200 response falls through to the generic error, not a blank one', async () => {
```
mgallego-keeper, line 450: **NB2 (low): Node's data: null empty-body shape is still unasserted, so the older half of the guard can be deleted with CI green**

The new test does its job. I reverted the `.length > 0` half of the guard at src/keeper.ts:830, and this test was the only failure of 122, with Received message: "".

One shape stays unasserted. If I reduce the guard to `if (response.data.length > 0)`, all 122 tests pass and `npx tsc --noEmit` exits 0. The compiler cannot help, because KeeperHttpResponse.data is declared Uint8Array at src/platform.ts:58. Node's fetchData leaves `data: null` for an empty body.

Under that reduced guard, a 400 with `data: null` throws TypeError: Cannot read properties of null (reading 'length') instead of "unknown ksm error, code 400". I confirmed both outcomes by probe.

Nothing is broken today. The shipped code handles both shapes correctly.

This is a pre-existing gap, not one you created. Release branch line 816 already read `if (response.data)`, and no base-branch test passed `data: null` either.

Optional one-line close: add a second case to the same test, `queryFunction: async () => ({ statusCode: 400, data: null as any, headers: [] })`, with the same `rejects.toThrow('unknown ksm error, code 400')`. That pins the shape every Node caller hits.

sdk/javascript/packages/core/test/keeper.test.ts
```diff
@@ -369,6 +370,127 @@ test('key rotation - suggested key id is adopted and persisted', async () => {
...
+            data: enc.encode(keyErrorResponse(8, { result_code: 'something-else' })),
+            headers: []
+        })
+    }
+    await expect(getSecrets(options)).rejects.toThrow()
+    await expect(getSecrets(options)).rejects.not.toThrow(/key rotation|unsupported key id/i)
+    expect(await storage.getString('serverPublicKeyId')).toBeUndefined()
+})
+
+test('a response body over the 64KB decode cap is truncated before parsing, not adopted as a valid key rotation', async () => {
```
mgallego-keeper, line 415: **NB3 (low): Nothing pins the decode cap to bytes, and the exact 65536 boundary is untested**

The two 64KB cap tests pad with ASCII, so byte length equals character length. Neither test pins the cap to bytes.

I checked two edits. Replacing src/keeper.ts:838 and src/keeper.ts:839 with `platform.bytesToString(response.data).slice(0, MAX_ERROR_BODY_DECODE_BYTES)` keeps all 122 tests green and tsc clean. Changing the cap to `MAX_ERROR_BODY_DECODE_BYTES - 1` also keeps all 122 green.

The behaviour is correct today. This is a missing regression guard, not a defect.

One test closes both gaps. Use a body whose byte length passes 65536 but whose character length does not, for example `keyErrorResponse(8, { padding: '中'.repeat(30000) })`. That is 30039 characters and 90039 bytes. Assert the rotation is not adopted, because the shipped byte slice cuts the document.

Small related point: MAX_ERROR_BODY_DECODE_BYTES is module private, so the tests hard-code 65536. Exporting it would let the tests derive their sizes from the source of truth.

sdk/javascript/packages/core/src/keeper.ts
```diff
@@ -854,10 +876,9 @@ const postQuery = async (options: SecretManagerOptions, path: string, payload: A
                     keyRotationAttempt++
                     continue
                 }
-            } else {
-                errorMessage = `unknown ksm error, code ${response.statusCode}`
+                throw new Error(platform.bytesToString(response.data.slice(0, 1000)))
```
mgallego-keeper, line 879: **NB5 (low): Two claims in the PR body do not hold against the release branch**

Two claims in the PR body do not hold against the release branch.

First, the Testing section says every new test was confirmed failing on the pre-fix code. I built a tree with these test files and the merge-base src/keeper.ts, then ran the suite. 5 failed and 117 passed. Three new tests pass on the base branch: test/keeper.test.ts:415, test/keeper.test.ts:434, and test/throttle.test.ts:131. They are forward guards on behaviour this PR introduces, not reproductions of a pre-PR defect. They are still worth keeping.

Second, the Fixed section describes a byte truncation replacing a character truncation. Against the base branch that is a no-op. Base line 817 is `platform.bytesToString(response.data.slice(0, 1000))`, already a byte slice on a Uint8Array. The character slice existed only inside this PR's own commits 5d5b03cb and 98262671, and 7983ba08 reverted it.

Your commit message on 3033c066 is accurate, and the CHANGELOG makes no such claim. Only the PR body overstates.

Suggested edit: drop the byte versus character bullet, or restate it as a decision made inside this PR. Then qualify the Testing line, for example: five of these fail on the release-branch code, and three are forward guards on new behaviour.

sdk/javascript/packages/core/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
...
+- KSM-1251 - Fixed throttle (HTTP 403) and server key-rotation detection running on an error body truncated to 1000 bytes. A throttle or key-rotation response padded past that size failed to parse, silently disabling the retry/rotation for that response. Both now run on the full body (decoding is capped at 64KB, well above any realistic body, to bound a pathological or malicious oversized response); the 1000-byte truncation still applies only to the message in the error the SDK throws when the response isn't recognized (the throttle-exhaustion and key-rotation-exhaustion errors are synthetic messages with no body content). Also fixed the non-200 error path checking only truthiness of the response body: an empty body is `null` on Node but a truthy, zero-length array on browser, so a non-200 response with a genuinely empty body threw a blank error message on browser instead of the intended "unknown ksm error" fallback.
```
mgallego-keeper, line 14: **NB6 (low): The CHANGELOG never documents the result_code widening, which changes what the SDK writes to config**

The CHANGELOG never mentions the result_code widening. `grep -c result_code CHANGELOG.md` returns 0.

That widening is user visible. I ran a 400 body of {"result_code":"key","key_id":8}, with no legacy error field, against both builds. On this head the SDK retries and writes serverPublicKeyId as "8" into the caller's configuration. On the release branch it makes one request and throws. So an upgrade changes what the SDK writes to config for that response shape.

The code comment on MAX_KEY_ROTATION_RETRIES was updated, and the PR body describes the change. Only the shipped release note omits it.

Suggested one-clause addition to the KSM-1251 entry: key-rotation detection now matches result_code or the legacy error field, matching the Python, Rust and Ruby SDKs.

I am not asking you to change the KSM-1128 entry. A {"error":"key"} body still triggers the retry, so that sentence stays true.

sdk/javascript/packages/core/src/keeper.ts
mgallego-keeper, line 879: **NB7 (low): The catch-all throw omits the HTTP status code, so an unparseable body gives a near-empty error**

Pre-existing and non-blocking. I raise it because it is the same defect class this PR closes for zero-length bodies.

The catch-all throw at src/keeper.ts:879 carries only the body. It drops the status code. Measured on this head: a 400 body of one newline throws Error("\n"), a 502 body of "\r\n" throws Error("\r\n"), and a 400 body of "0" throws Error("0"). Each message trims to almost nothing and names no status.

The merge base behaves the same way, so this PR did not cause it. The `.length > 0` fix correctly closed a different problem, the Node and browser split for a genuinely empty body.

The Python SDK does not have this gap. sdk/python/core/keeper_secrets_manager_core/core.py:851 always includes the status code in its equivalent throw.

Optional one-line fix: include the status code in the catch-all, for example `throw new Error(`ksm error, code ${response.statusCode}: ${platform.bytesToString(response.data.slice(0, 1000))}`)`. That covers every unparseable body, not only whitespace.

sdk/javascript/packages/core/test/keeper.test.ts
```diff
@@ -369,6 +370,127 @@ test('key rotation - suggested key id is adopted and persisted', async () => {
...
+    await initializeStorage(storage, FAKE_ONE_TIME_TOKEN, 'fake.keepersecurity.com')
+    const enc = new TextEncoder()
+    const marker = 'MARKER_PAST_1000_BYTES'
+    const body = JSON.stringify({ error: 'not_recognized', padding: 'a'.repeat(2000) + marker })
+    const options: SecretManagerOptions = {
+        storage,
+        queryFunction: async () => ({ statusCode: 400, data: enc.encode(body), headers: [] })
+    }
+    const error: any = await getSecrets(options).catch(e => e)
+    expect(error.message.length).toBe(1000)
```
mgallego-keeper, line 445: **NB8 (nit): The 1000-byte truncation test asserts characters, not bytes, so a character slice passes it**

Nit, and nothing ships broken. The test at test/keeper.test.ts:434 is named for bytes, but line 445 asserts error.message.length, a character count. The body pads with 'a'.repeat(2000), so the two counts are equal and the assertion cannot tell them apart.

I changed src/keeper.ts:879 to `platform.bytesToString(response.data).slice(0, 1000)` and all 122 tests still passed, with tsc clean.

On a non-ASCII body a character slice yields up to about 2900 bytes rather than 1000, and it can end on an unpaired surrogate.

Suggested change: assert the exact expected string, `expect(error.message).toBe(platform.bytesToString(bodyBytes.slice(0, 1000)))`. That fails for a character slice and passes for a byte slice at any encoding width. Add one non-ASCII padding case as well, so the assertion matches the test name.

sdk/javascript/packages/core/src/keeper.ts
```diff
@@ -78,6 +87,11 @@ export type {KeeperCryptoFailureReason, KeeperDecryptionErrorInfo} from './error
 // pinned jitter.
 export const throttleJitter = (): number => Math.random() * 0.25
 
+// Server error envelopes use `result_code` in newer responses and the legacy `error` field in
+// older ones for the same discriminator; both throttle and key-rotation detection need whichever
+// one is present, so this is shared rather than duplicated at each call site.
+const resultCode = (obj: any): unknown => obj?.result_code ?? obj?.error
```
mgallego-keeper, line 93: **NB9 (nit): The deliberate nullish coalescing in the resultCode helper is not pinned by any test**

Optional, test only. The `??` in resultCode at src/keeper.ts:93 is a deliberate choice. An empty result_code must count as present, and must not fall through to the legacy error field.

No test pins it. I changed `??` to `||` and all 122 tests passed, with tsc clean.

The precedence half is pinned. Flipping to `obj?.error ?? obj?.result_code` fails test/keeper.test.ts:398.

No backend response is known to send an empty result_code, so this is a consistency guard, not a live bug.

One extra case closes it. Reuse `keyErrorResponse(8, { result_code: '' })`, assert no rotation, and assert serverPublicKeyId stays undefined. Asserting the query-function call count is 1 is the strongest signal, because `||` raises it to 4.

sdk/javascript/packages/core/test/throttle.test.ts
```diff
@@ -17,14 +17,15 @@ import {
 const FAKE_TOKEN = 'YyIhK5wXFHj36wGBAOmBsxI3v5rIruINrC8KXjyM58c'
 const enc = new TextEncoder()
 
-const throttle403 = (retryAfter?: number): KeeperHttpResponse => ({
+const throttle403 = (retryAfter?: number, extra?: Record<string, unknown>): KeeperHttpResponse => ({
```
mgallego-keeper, line 20: **NB10 (nit): Both test helpers apply extra last, so a colliding key silently replaces a named field**

Carry-over from an earlier round, still a nit. Both keyErrorResponse at test/keeper.test.ts:15 and throttle403 at test/throttle.test.ts:20 apply `...extra` last. A key in extra replaces the named field above it, and TypeScript reports nothing.

throttle403 is the sharper case. Its named fields are error, message and retry_after, which are the exact fields the throttle tests assert on. A future `throttle403(7, { retry_after: 1 })` would test 1 second, not 7. A future `throttle403(7, { error: 'x' })` would stop being a throttle body, so the retry would not fire at all. Both compile cleanly.

Two small options. Add one comment above each helper, stating that extra is applied last and can replace a named field on purpose, because test/keeper.test.ts:387 relies on that to drop error. Or tighten the type so a collision is a compile error, and write the one intentional override explicitly.

Separately, inline padding literals now appear five times across the two files: test/keeper.test.ts:422, test/keeper.test.ts:439, test/keeper.test.ts:469, test/throttle.test.ts:119 and test/throttle.test.ts:132. My earlier note said four. Five is the correct count, and it has not grown since the last round.

mgallego-keeper (approved):

PR #1152, opened 2026-09-01
fix(javascript): KSM-1267 classify getFolders() decryption failures

mgallego-keeper:
## Summary
- `getFolders()` now classifies why an undecryptable folder was skipped (`integrity`, `format`, `missing-key`, `malformed-data`, or `unknown`) instead of an opaque, unclassified error, and logs a summary line naming every folder UID it had to omit, so a partial result is never silently presented as complete.
- Adds an optional `onDecryptionError` callback to `SecretManagerOptions`, invoked once per skipped folder. Throwing from it aborts `getFolders()` instead of returning a partial result (fail closed). Additive: existing callers that do not set it see no behavior change.
- Both the Node and browser platforms' `unwrap()` now reject an unwrapped key of the wrong byte length immediately instead of caching a malformed key and failing later at an unrelated call site.
- The underlying finding (the shared-folder key wrap uses unauthenticated AES-256-CBC, a format fixed server-side that the SDK cannot change unilaterally) was reviewed and confirmed low-impact: a manipulated folder key is still caught by the existing AES-GCM authentication on the record keys inside that folder. See KSM-1267 for the full review and scope decision.

## Test plan
- [x] `npm test`: 113 passing across 14 suites (78 pre-existing + 35 new), zero regressions.
- [x] Existing master-regression test (`getFolders skips an undecryptable folder and returns the good one`) unchanged and passing.
- [x] New coverage across 5 files: all 5 classification values, on both platforms individually and side by side (same fixture, both platforms agree on classification despite completely different underlying error text), the `onDecryptionError` contract, fail-closed behavior, and backward compatibility.
- [ ] Optional/QA: re-run `getFolders` against a QA vault containing shared folders and inner folders per KSM-1267's test instructions; this PR was verified against synthetic server responses only.

Ref: KSM-1267


stas-schaller (approved):
👍

PR #1153, opened 2026-09-01
fix(javascript): stop getSharedFolderUid hanging on a cycle (KSM-1297)

mgallego-keeper:
## Summary
- getSharedFolderUid walked the folder parent chain in an unbounded while loop with no visited set; a cycle in server supplied folder data (two folders naming each other as parent, or a folder naming itself) spun forever and, since the loop is synchronous with no await, blocked the entire JS event loop in Node and froze the tab in a browser.
- Bound the walk with a visited set and throw a descriptive error naming the folder UID where the cycle closes. The existing per-folder try/catch in fetchAndDecryptFolders (KSM-1079) already skips and continues on any thrown error, so no caller changes were needed.
- Added regression coverage for both the Node and browser platform builds: self-cycle, two- and three-folder rings with exact log-message assertions, a 500-folder ring proving the fix is genuinely bounded, a real two-level non-cyclic nested folder chain to confirm unchanged behavior, and the pre-existing "parent not found" case to confirm its message stays distinct from the new cycle message.

## Test plan
- [x] `npm test` passes in `sdk/javascript/packages/core` (88/88, including 10 new tests)
- [x] Confirmed the fix is the only change to `keeper.ts`, no caller changes
- [x] Confirmed the new browser-platform test file genuinely exercises `browserPlatform.ts` (real WebCrypto), not a silent fallback to the Node platform


stas-schaller (approved):
Good work!

stas-schaller:
Correction: after approving, I checked CI and found the Node 20/22/24 test jobs failing (7 tests) on the current merge. Not a problem with this fix; it's a landed conflict. #1152 (KSM-1267) added:

    if (!folder.folderKey) {
        throw new KeeperCryptoError(`Folder key missing for UID ${folder.folderUid}`, 'missing-key', folder.folderUid)
    }

right before the `getSharedFolderUid` call, and that landed on `release/sdk/javascript/core/v17.6.0` after this branch was cut. This PR's cycle-fixture folders all set `folderKey: ''` (reasonable at the time, since they were never meant to reach decryption), so the new missing-key check now fires first and the regression tests never reach the cycle-detection code they're asserting on.

`getSharedFolderUid`'s visited-set cycle guard itself is unaffected and still correct; I re-verified it by hand against both the self-cycle and two-folder-ring cases. What's needed before merge: rebase onto the current release branch tip (`9a4ebfe3`) and give the cycle-member fixtures in `keeper.test.ts` and `keeper.browserPlatform.test.ts` a non-empty placeholder `folderKey` (e.g. `'x'`) so they clear the new missing-key check and reach the cycle throw being tested.

Holding off on merging until CI is green.

PR #1160, opened 2026-09-10
Fix interpolate file writes on Windows (fchmod and rename)

### LINKED ISSUE #1159: `interpolate` cannot write files on Windows: `module 'os' has no attribute 'fchmod'`
`ksm interpolate` cannot write output files on Windows. Two separate causes in the write path:

**1. `os.fchmod` is unavailable.** It only gained Windows support in [Python 3.13](https://docs.python.org/3/library/os.html#os.fchmod), and the Windows installer bundles an older interpreter: `keeper-ksm.exe version` reports `Python Version: 3.12.10` on the 1.3.0 installer.

The shipped build always hits an error on Windows:
```
> echo PLAIN=hello > tpl.txt
> keeper-ksm.exe interpolate tpl.txt -o out.txt
Error: Failed to write output file out.txt: module 'os' has no attribute 'fchmod'
```

Exit code 1, no file written. No secrets are needed to reproduce: the template contains no `keeper://` references, so this is purely the file write. Writing to stdout (omitting `-o`) works fine.

**2. `os.rename` cannot overwrite.** It raises `FileExistsError` on Windows when the destination exists, so even with the permission call fixed, writing over an existing file still fails — and that hits `-w/--in-place` unconditionally, since its destination exists by definition. Verified on Windows 11 / Python 3.14.4:

```
os.rename over existing:  FAIL -> FileExistsError [WinError 183]
os.replace over existing: OK
```

**Affected code**, identical on `master` and at tag `ksm-cli-v1.5.0`:

| Line | Call | Paths |
|---|---|---|
| `interpolate.py:407`, `:725` | `os.fchmod` | `-w`, `-o` |
| `interpolate.py:413`, `:731` | `os.rename` | `-w`, `-o` |

Worth noting `-o` is the pattern `--help` recommends as the safe one, with bare stdout gated behind `--allow-unsafe-for-eval` — so on Windows the safe path is the only one that doesn't work.

Suggested fix: guard the `fchmod` call on `hasattr(os, 'fchmod')` and switch to `os.replace()`. PR #1160.

yutamago:
Fixes #1159.

`ksm interpolate` cannot write an output file on Windows. While fixing the reported
`os.fchmod` crash I found a second call in the same path with no Windows-compatible
semantics, so this addresses both.

## The two problems

**1. `os.fchmod()` is unavailable on Windows before Python 3.13.**

The reported crash. Worth correcting one detail from my issue report: `os.fchmod` is *not*
Unix-only — it [gained Windows support in Python 3.13](https://docs.python.org/3/library/os.html#os.fchmod).
The reason every Windows user hits this anyway is that the official installer bundles
Python 3.12, so the affected build ships with the CLI:

```
> keeper-ksm.exe version
Python Version: 3.12.10
```

A `pip install` on Windows with Python 3.13+ would not hit this one — but would still hit
the next.

**2. `os.rename()` cannot overwrite on Windows.**

Not in the original report. `os.rename()` raises `FileExistsError` when the destination
exists on Windows, so even with the permission call fixed, writing to an existing output
file fails. Confirmed on Windows 11 / Python 3.14.4:

```
os.rename over existing:  FAIL -> FileExistsError [WinError 183] Cannot create a file when that file already exists
os.replace over existing: OK
```

This matters because re-running interpolation over an existing file is the normal case, and
it affects `-w/--in-place` unconditionally, since that destination exists by definition.

## The change

- Both permission calls go through a new `_apply_secure_perms()` helper, guarded on
  `hasattr(os, 'fchmod')`.
- Both `os.rename()` calls become `os.replace()`, the cross-platform atomic overwrite.

POSIX behaviour is unchanged — output files are still created with mode `0600`, covered by a
test.

## Testing

Four regression tests in `tests/interpolate_test.py`. The two that cover the bugs fail on
`master` and pass with this change:

```
# master
FAILED tests/interpolate_test.py::TestWindowsFileWrite::test_write_output_without_fchmod
FAILED tests/interpolate_test.py::TestWindowsFileWrite::test_write_output_overwrites_existing_file
2 failed, 2 passed

# this branch
26 passed
```

Since `os.rename()` overwrites silently on POSIX, that test monkeypatches it to raise
`FileExistsError` so the regression stays meaningful on a Linux runner.

Full `integration/keeper_secrets_manager_cli/tests/` suite: `13 failed, 133 passed, 4 skipped`,
against a `master` baseline of `13 failed, 129 passed, 4 skipped` — the same 13 pre-existing
failures, which look like incompatibilities with current dependency versions and are unrelated
to this change. Delta is the 4 added tests.

### DIFF
--- a/integration/keeper_secrets_manager_cli/keeper_secrets_manager_cli/interpolate.py
+++ b/integration/keeper_secrets_manager_cli/keeper_secrets_manager_cli/interpolate.py
@@ -404,13 +404,13 @@ class Interpolate:
                 # Use atomic write for in-place editing
                 temp_path = Path(input_file).with_suffix('.tmp')
                 with open(temp_path, 'w', encoding='utf-8') as f:
-                    os.fchmod(f.fileno(), self.SECURE_FILE_PERMS)
+                    self._apply_secure_perms(f)
                     f.write(result)
                     f.flush()
                     os.fsync(f.fileno())
 
                 # Atomic rename
-                os.rename(temp_path, input_file)
+                os.replace(temp_path, input_file)
 
                 # Create safety marker
                 self._create_safety_marker(input_file)
@@ -696,6 +696,19 @@ class Interpolate:
     # FILE OPERATIONS
     # ========================================================================
 
+    def _apply_secure_perms(self, file_obj):
+        """Restrict an open file to owner-only access, where the platform allows it.
+
+        os.fchmod() only gained Windows support in Python 3.13; on older builds
+        it does not exist and calling it raises AttributeError. Windows has no
+        NTFS equivalent of the POSIX mode bits either - os.chmod() there only
+        toggles the read-only flag - so rather than approximate 0o600 with a
+        call that does not mean the same thing, the step is skipped and the file
+        inherits the permissions of its parent directory.
+        """
+        if hasattr(os, 'fchmod'):
+            os.fchmod(file_obj.fileno(), self.SECURE_FILE_PERMS)
+
     def _write_output(self, content, output_file):
         """Write content to output file with secure permissions"""
         try:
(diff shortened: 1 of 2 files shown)


stas-schaller:
Thank you for bringing this to our attention and opening a PR! We'll review it and if everything looks good we'll include it in our next release.

stas-schaller:
Appreciate the detailed writeup and the regression tests. I've pulled it into `release/tool/cli/v1.6.0` as commit f6a4d849 (your authorship carries over on that commit), tracked under KSM-1446 on our end. Since the code's already moved over, I'll close this PR out rather than merge it separately. Thanks for tracking down the second `os.rename` issue too.

PR #1161, opened 2026-09-11
fix(javascript): widen throttle retry status gate to accept 429 (KSM-1395)

mgallego-keeper:
## Summary
- `postQuery`'s throttle retry only fired on HTTP 403. The backend returned 403 for a throttled Secrets Manager request. A fix changed the shared response code to 429. The SDK's gate was correct when written and has been silently dead ever since: a throttled request now fails immediately with a plain `Error` instead of retrying with backoff. 
- Widened the check to accept 429 in addition to 403, rather than replacing it: leaves open whether a second rate limiter, on some other call pattern, still returns 403. Accepting both keeps that path working if it exists.
- Corrected the two code comments in `keeper.ts` and `KeeperThrottleError`'s doc comment in `errors.ts` that stated the backend throttles with 403.
- Added a CHANGELOG entry under the existing unreleased 17.6.0 section.

## Test plan
- [x] `npm test` passes in `sdk/javascript/packages/core` (127/127, including 18 in `throttle.test.ts`)
- [x] `npx tsc --noEmit` clean
- [x] Rewrote every status-dependent case in `throttle.test.ts` (retry-then-succeed, exhaustion, `retry_after` honored, non-throttle body not retried) via `describe.each([403, 429])`, so the suite proves 403 and 429 are handled identically rather than adding an isolated new test alongside the old one
- [x] Kept the 403-vs-502 "wrong status, throttled body, not retried" case and the `KeeperThrottleError` instanceof case as single, status-independent tests, since neither depends on which of the two gated statuses is used
- [x] Confirmed this is the only behavior change: no caller-facing API changes, `SecretManagerOptions` unchanged

### DIFF
--- a/sdk/javascript/packages/core/CHANGELOG.md
+++ b/sdk/javascript/packages/core/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Change Log
 
 ## 17.6.0
+- KSM-1395 - Fixed throttle retry never firing against the real backend: the retry gate checked only for HTTP 403, but the backend has returned HTTP 429 for a throttled Secrets Manager request since 2026-06-15 (an unrelated login-security fix changed a response code shared with the Secrets Manager throttle path; see KSM-1386 for the full history). The gate now accepts both 403 and 429, so a throttled request is retried as originally intended (KSM-880) instead of failing immediately.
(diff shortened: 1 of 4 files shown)


stas-schaller (approved):
👍👍

PR #1163, opened 2026-09-14
JavaScript SDK: correct KSM-1256 changelog recovery claim

mgallego-keeper:
## Summary
- Victoria Bastos found this while she validated KSM-1256. She posted the finding as a comment on the ticket.
- The changelog entry made two claims about correcting a poisoned server key id: a server rotation hint, or a caller re-supplying serverPublicKeyId. Only the first claim is true.
- persistServerPublicKeyOptions treats a bare serverPublicKeyId as a stale replay once any id already exists in storage. The write becomes a silent no-op, not a correction.
- This change corrects the wording. The adjacent KSM-1255 entry needs no change; its own recovery claim is accurate for its own scenario.

## Test plan
- [x] Documentation only change. No code or test changes.
- [x] Checked the corrected wording against the current generateTransmissionKey and persistServerPublicKeyOptions code on this branch.


stas-schaller (approved):
Thank you!

PR #1166, opened 2026-09-15
JavaScript Azure Key Vault Storage: security dependency updates (KSM-1220)

stas-schaller:
## Summary
JavaScript Azure Key Vault Storage: updates js-yaml, handlebars, minimatch, brace-expansion, picomatch, and other transitive dependencies to resolve every open Dependabot advisory on this package's `package-lock.json`.

### Maintenance
- KSM-1220 - Resolved all flagged advisories via `npm audit fix`. `package.json` unchanged; every affected package resolves within its existing declared range.
- KSM-1220 - Bumped `@azure/identity` 4.13.0 → 4.13.3 (patch); this pulled in `@azure/msal-node` 3.8.3 → 6.0.0 and `@azure/msal-browser` 4.26.2 → 5.21.0 transitively (both jumped major versions even though `@azure/identity` itself only moved a patch)

## Testing
```
cd sdk/javascript/packages/azure && npm audit   # 0 vulnerabilities
cd sdk/javascript/packages/azure && npm run build && npm test   # 50/50 (48 unit + 2 integration, integration suite skips cleanly)
```
- Added a new integration suite (`test/AzureKeyValueStorage.integration.test.ts`) covering both the `ClientSecretCredential` and `DefaultAzureCredential` paths against a real Azure Key Vault key, to catch any authentication break from the msal major-version cascade above. It skips cleanly (not run) everywhere no `AZURE_TEST_*` env vars are set, since no Azure credentials are provisioned anywhere yet. Wiring it into CI is tracked separately: KSM-1418.

## Breaking Changes
None.

## Related Issues
- Jira: KSM-1220
- Follow-up: KSM-1418

---
**Draft, not ready to merge.** This branch needs `@keeper-security/secrets-manager-core` bumped to v17.6.0 once that release publishes. Its range (`^17.3.0`) already allows 17.6.0; it's just not out yet. Opening as a draft now so review on the dependency fixes isn't blocked on that.

### DIFF
--- a/sdk/javascript/packages/azure/package-lock.json
+++ b/sdk/javascript/packages/azure/package-lock.json
@@ -293,13 +293,13 @@
             }
         },
         "node_modules/@babel/code-frame": {
-            "version": "7.27.1",
-            "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.27.1.tgz",
-            "integrity": "sha512-cjQ7ZlQ0Mv3b47hABuTevyTuYN4i+loJKGeV9flcCgIK37cCXRh+L1bd3iBHlynerhQ7BhCkn2BPbQUL+rGqFg==",
+            "version": "7.29.7",
+            "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.29.7.tgz",
+            "integrity": "sha512-Aup7aUOfpbAUg2ROOJN6Iw5f9DMBlzu0mIkm/malLQFN/YQgO48wCj0Kxa3sEHJvPVFg7siR+qRInwXd2qhQKw==",
             "dev": true,
             "license": "MIT",
             "dependencies": {
-                "@babel/helper-validator-identifier": "^7.27.1",
+                "@babel/helper-validator-identifier": "^7.29.7",
                 "js-tokens": "^4.0.0",
                 "picocolors": "^1.1.1"
             },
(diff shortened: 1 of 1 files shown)


mgallego-keeper (changes requested):
Security review.

This review checked the dependency bump for supply-chain integrity and CVE-fix completeness, and traced the side effects of each version bump.

What I verified:
- `npm audit` on the pre-change lockfile found 15 vulnerabilities. The post-change lockfile has 0. I ran both audits myself.
- Every changed lockfile entry resolves to `https://registry.npmjs.org/` with a matching integrity hash. I checked all changed entries, including every package involved in a major-version jump.
- `minimatch` lands at the correct patched version at all 6 nested locations: 3.1.5, 9.0.9, or 10.2.6, depending on the line.
- `package.json` is unchanged. `npm ci` succeeds with no warnings beyond two pre-existing, unrelated ones.

Requested change:

Bumping `@azure/identity` by a patch version pulled in two major-version jumps. `@azure/msal-node` moved from 3.8.3 to 6.0.0. `@azure/msal-browser` moved from 4.26.2 to 5.21.0. Both sit in the OAuth library that authenticates to Azure Key Vault. I confirmed the new versions are legitimate and unmodified.

The test suite mocks `@azure/identity` and `@azure/keyvault-keys` directly, in `test/AzureKeyValueStorage.test.ts` and `test/utils.test.ts`. None of the 48 passing tests exercise this authentication path.

Please add a manual or integration smoke test against a real Azure AD tenant before merge, covering both credential paths in `src/AzureKeyValueStorage.ts`. Please also name this cascade explicitly in the PR description. Right now it is an undocumented side effect of a generic audit-fix commit.

mgallego-keeper (approved):

PR #1167, opened 2026-09-15
JavaScript Oracle KMS Storage: security dependency updates (KSM-1221)

stas-schaller:
## Summary
JavaScript Oracle KMS Storage: updates js-yaml, handlebars, minimatch, brace-expansion, picomatch, and other transitive dependencies to resolve every open Dependabot advisory on this package's `package-lock.json`.

### Maintenance
- KSM-1221 - Resolved all flagged advisories via `npm audit fix`. `package.json` unchanged; every affected package resolves within its existing declared range.

## Testing
```
cd sdk/javascript/packages/oracle && npm audit   # 0 vulnerabilities
cd sdk/javascript/packages/oracle && npm run build && npm test   # 50/50 passing
```

## Breaking Changes
None.

## Related Issues
- Jira: KSM-1221

---
**Draft, not ready to merge.** This branch needs `@keeper-security/secrets-manager-core` bumped to v17.6.0 once that release publishes. Its range (`^17.3.0`) already allows 17.6.0; it's just not out yet. Opening as a draft now so review on the dependency fixes isn't blocked on that.

### DIFF
--- a/sdk/javascript/packages/oracle/package-lock.json
+++ b/sdk/javascript/packages/oracle/package-lock.json
@@ -29,13 +29,13 @@
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.27.1",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.27.1.tgz",
-      "integrity": "sha512-cjQ7ZlQ0Mv3b47hABuTevyTuYN4i+loJKGeV9flcCgIK37cCXRh+L1bd3iBHlynerhQ7BhCkn2BPbQUL+rGqFg==",
+      "version": "7.29.7",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.29.7.tgz",
+      "integrity": "sha512-Aup7aUOfpbAUg2ROOJN6Iw5f9DMBlzu0mIkm/malLQFN/YQgO48wCj0Kxa3sEHJvPVFg7siR+qRInwXd2qhQKw==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
-        "@babel/helper-validator-identifier": "^7.27.1",
+        "@babel/helper-validator-identifier": "^7.29.7",
         "js-tokens": "^4.0.0",
         "picocolors": "^1.1.1"
       },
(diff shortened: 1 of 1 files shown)


mgallego-keeper (changes requested):
Security review.

This review checked the dependency bump for supply-chain integrity and CVE-fix completeness, and compared this package's fix against its sibling packages in the same batch.

What I verified:
- `npm audit` on the pre-change lockfile found 15 vulnerabilities. The post-change lockfile has 0. I ran both audits myself.
- I checked whether this package was missing a fix its sibling packages received. It was not. `browserslist`, `baseline-browser-mapping`, `@humanfs/node`, and `flatted` are all fixed, as a side effect of the `npm audit fix` commit.
- Every changed lockfile entry resolves to `https://registry.npmjs.org/` with a matching integrity hash. I checked this exhaustively, not by sample.
- `minimatch` clears its patched threshold at all 6 nested locations.
- `package.json` is unchanged. `npm ci` succeeds cleanly on the final lockfile.

Requested change:

Clearing an advisory on a pinned `uuid` version inside `oci-common` required bumping `oci-common` and `oci-keymanagement` from 2.127.0 to 2.140.1. That is 19 published releases apart. Neither package name appears in any commit message, in the CHANGELOG entry, or in the PR description.

`oci-common` and `oci-keymanagement` are the actual Oracle Key Management SDK this package wraps. I confirmed the new version is genuinely published by Oracle. I also checked Oracle's own SDK changelog for this range and found no breaking changes to the Key Management or Common modules.

The version jump itself looks safe. Please add one line to the commit message or the CHANGELOG naming `oci-common` and `oci-keymanagement`, with the old and new versions, before merge. A production KMS SDK version change should never ride along as a silent side effect of a generic audit fix.

mgallego-keeper (approved):
Security re-review, round 2.

Confirmed: the new commit (585a72f2a) adds exactly the CHANGELOG line I asked for. It names oci-common and oci-keymanagement. It gives both versions, 2.127.0 to 2.140.1. It gives the reason: a nested uuid advisory.

I diffed this commit against the one I reviewed before. Only the CHANGELOG changed. The lockfile did not change. My previous lockfile audit still holds.

I also re-verified the security claims myself this round, not just by reading:
- Ran `npm audit --package-lock-only` on the actual PR lockfile myself. Result: 0 vulnerabilities.
- Checked all 503 resolved lockfile entries by hand. Every one resolves from `https://registry.npmjs.org/` with a matching integrity hash. None uses another registry.
- oci-common and oci-keymanagement both land on the real 2.140.1 package. This matches the Socket bot scan on this PR.
- The CI workflow for this package runs `npm ci` and `npm test` against the actual latest commit. It passed.

Requested change fully addressed. Approving.

Two notes, both outside this PR's diff. Neither blocks this PR.

1. The PR description still says "Draft, not ready to merge." But this PR is no longer marked as a draft. Please update the description so the two do not contradict each other.

2. While reviewing this package's `OracleKeyValueStore.ts`, I found that it, and the sibling AWS, Azure, and GCP storage backends, write the config file with `fs.writeFile` calls. None of these calls sets a restrictive file mode. On a standard umask, this leaves the file world readable. This includes the fully decrypted plaintext in the `decryptConfig(true)` path. This defect is unrelated to this PR and predates it. Filed as KSM-1444, linked to the existing KSM-1367 epic covering the same four files.

PR #1169, opened 2026-09-15
Azure DevOps Extension: security dependency updates (KSM-1238)

stas-schaller:
## Summary
Azure DevOps Secrets Manager task: updates adm-zip, azure-pipelines-task-lib, js-yaml, minimatch, mocha, and other dependencies to resolve every open Dependabot advisory on this package.

### Maintenance
- KSM-1238 - Resolved adm-zip, js-yaml, minimatch, brace-expansion, follow-redirects, nodejs-file-downloader, uuid, and serialize-javascript via `npm audit fix`.
- KSM-1238 - `azure-pipelines-task-lib` (a runtime dependency that ships in the published task, not a dev tool) resolved from 5.2.8 to 5.280.1 as part of the same fix. Same major version per its existing `^5.2.4` range; the jump pulled in a patched `adm-zip`.
- KSM-1238 - `mocha` (devDependency only, never ships in the published task) bumped from `^11.7.1` to `^12.0.1` via `npm audit fix --force` to clear the last two advisories (`diff`, `serialize-javascript`). This is the one real semver-major change in the diff. Verified the test suite still passes after.

## Testing
```
cd integration/keeper_secrets_manager_azure_pipeline_extension/ksm-azure-devops-secrets-task
npm audit   # 0 vulnerabilities
npm run build && npm test   # 1/1 passing
```

## Breaking Changes
None. `mocha` is a devDependency; the published task's runtime behavior is unaffected.

## Related Issues
- Jira: KSM-1238

---
**Draft, not ready to merge.** This package also depends on `@keeper-security/secrets-manager-core@^17.2.0` and is included in the same batch waiting on JS SDK core v17.6.0's publish before the final version bump and merge. Opening as a draft now so review on the dependency fixes isn't blocked on that. No `CHANGELOG.md` exists for this integration, so none was added.


mgallego-keeper (approved):
Security review.

This review checked the dependency bump for supply-chain integrity and CVE-fix completeness, and separately checked what actually ships in the published task.

What I verified:
- `npm audit` on the pre-change lockfile found 9 vulnerabilities. The post-change lockfile has 0. I ran both audits myself.
- `adm-zip` resolves to 0.6.1. I confirmed against its upstream commit history that this version fixes the symlink-following issue in `GHSA-vwc7-r8mq-g2x9`, even though that advisory lists no patched version.
- The `package.json` change matches the PR description exactly. `mocha` moves from `^11.7.1` to `^12.0.1`. `azure-pipelines-task-lib` resolves from 5.2.8 to 5.280.1, within its existing `^5.2.4` range.
- Every changed lockfile entry resolves to `https://registry.npmjs.org/` with a matching integrity hash.

One item for a separate follow-up, not blocking this PR:

I built the actual VSIX package with `tfx-cli` to check what ships. `mocha`, `ts-node`, `typescript`, and their full dependency trees are present in the built package. The release workflow runs plain `npm install`, with no production-only flag and no file allowlist.

This means the claim that `mocha` never ships in the published task does not match what actually gets published. This is a pre-existing gap in the packaging workflow, not something this PR introduced. Given this task runs inside customer pipelines with access to secrets, filed as a separate ticket to exclude devDependencies from the published package.

Verdict: no changes requested on this PR's own diff.

PR #1170, opened 2026-09-16
feat: allow creation of a user inside a subfolder

TiJof:
Before the ansible role hardcode None for the subfolder_uid, this commit allow to use a custom one.
Now you can create an entry specifying the shared folder (mandatory) and a folder inside this one.

- name: Create a new record in a subfolder
  keeper_create:
    shared_folder_uid: XXX
    _**subfolder_uid: YYY**_
    record_type: login
    title: My Title
    fields:
      - type: login
        value: john.doe@nowhere.com
  register: my_new_record


mgallego-keeper:
@TiJof Thanks for this contribution! We are adopting the subfolder_uid parameter name you proposed for keeper_create, in place of our existing folder_uid. Your name is credited on the commit that makes this change, in PR #1182. Much appreciated!

PR #1181, opened 2026-09-17
JS SDK GCP KMS Storage: write the config file with restrictive permissions

stas-schaller:
## Summary
JavaScript SDK GCP KMS Storage wrote its config file to disk with no restrictive mode, so the encrypted client ID, private key, and app key (and, on the decryptConfig autosave path, the fully decrypted plaintext config) were readable by any other local user or process on a shared host.

### Fixed
- All four write sites (`saveConfig`, `decryptConfig`'s autosave path, and both writes in `createConfigFileIfMissing`) now write through a new `writeFileAtomicSync()` helper: a temp file is created at mode 0600 in the same directory, fsync'd, chmod'd again, then renamed over the target. The previous approach (`fs.writeFile(..., { mode: 0o600 })` followed by `fs.chmod`) was a no-op against a config file already on disk from an earlier SDK version, since `mode:` only takes effect when a file is created. The follow-up `fs.chmod` did fix the final mode, but rewrote the same inode, so a process already holding a read handle on the old file kept reading straight through the rewrite. The atomic rename produces a new inode, closing that gap. (KSM-1450)
- Rebased onto #1180 (KSM-1370) so both fixes land together. `createConfigFileIfMissing()` needed changes from both PRs at the same line; a mechanical merge resolution between the two independent branches was shown to silently drop one fix while the full test suite stayed green.

## Testing
```
npm test
```
The new tests in `GCPKeyValueStorage.atomicWrite.test.ts` use real `fs` against real temp files, not mocked `fs.writeFile`/`fs.chmod` calls, since a mocked-call assertion can pass even when the resulting on-disk file isn't actually secured.

## Breaking Changes
Version bumped to 1.1.0, not 1.0.1. A config file already exposed under the old permissions is now genuinely remediated on its next write (new inode), not just given a new mode bit while an existing reader keeps working. `createConfigFileIfMissing()` failures on a non-ENOENT `fs.access` error (KSM-1370) now propagate instead of silently resetting the config.

## Related Issues
- Depends on #1180 (KSM-1370) - this branch is rebased on top of it and should merge after it
- Jira: KSM-1450


mgallego-keeper (changes requested):
## Summary

This PR hardens the JavaScript GCP KMS storage. It stops the SDK from writing the config file world readable. It adds one helper and routes four write sites through it:

```ts
private async writeSecureConfigFile(path: string, data: Buffer | string): Promise<void> {
  await fs.writeFile(path, data, { mode: 0o600 });
  await fs.chmod(path, 0o600);
}
```

The config file holds the encrypted client ID, private key, and app key. One of the four sites, `decryptConfig(autosave = true)`, writes the full decrypted plaintext.

The goal is correct and the ticket is real. The chosen write method is weaker than a pattern already used in this repository, in the package this one depends on (`@keeper-security/secrets-manager-core`, KSM-1263 and KSM-1265).

**Requesting changes.** Two blocking items below, then a shorter list.

## Blocking

### 1. `mode: 0o600` does nothing on a file that already exists

`fs.promises.writeFile` opens the file with flag `'w'`. That flag is `O_WRONLY|O_CREAT|O_TRUNC`. `open(2)` only applies the `mode` argument when it creates the file. Node's own type definitions say this directly: `mode` sets the file mode, but only if the file was created.

So on a config file that already exists at mode 0644, the write happens first, at the old mode. The `chmod` call only runs after. This is the exact scenario named in the PR description and the changelog entry.

Two consequences follow from this.

First, this creates a real exposure window, not just a narrow race. Any process watching the file with `inotify` gets notified the instant new data lands. One read is enough to capture a long lived credential.

Second, and more important: this does not fix an already exposed config file. Unix checks file permissions only at `open(2)`. `chmod(2)` changes the mode bits on the inode. It does not revoke access already granted to an open file descriptor. `writeFile` truncates and rewrites the same inode. An attacker who opened the old 0644 config once can keep reading every future version of that file. This means the fix does not remediate the exact case it claims to fix.

The correct pattern already exists in this repository, in `sdk/javascript/packages/core/src/node/localConfigStorage.ts`, function `writeFileAtomic`. It opens a randomly named temp file in the same directory, at mode 0600 (which works, because the file is new). It writes, calls `fsync`, chmods the temp file, then renames it over the target. The new inode orphans any earlier open descriptor.

That same file already has a comment stating this exact fact, at line 10:

```ts
// fs.openSync's mode argument is only honored when the file is created; it is a no-op on an
// existing file, so permissions must be re-asserted after every write, not just the first one.
```

And its own test suite states directly why a truncate-then-chmod approach is not enough, at `test/localConfigStorage.test.ts:866`:

```ts
// An in-place truncate-then-chmod would end up at 0600 too - the inode staying the same
// is what actually distinguishes a real rename from that, so mode alone isn't enough.
```

Suggested fix: write to a temp file in the same directory, at mode 0600, then rename over the target. See the code block near the end of this review for a full replacement.

### 2. A failed `chmod` after a successful write creates a new failure case

`writeSecureConfigFile` has no error handling between its two calls. `fs.chmod` runs after the bytes are already on disk. It has no try/catch of its own. A rejection reaches the caller exactly as if the whole write had failed. Before this PR, a resolved `writeFile` call meant success. After this PR, the write can succeed and the function can still throw.

The most likely trigger is ordinary, not exotic. `chmod(2)` requires the calling process to own the file. A container image that creates the config file as root, then runs the app as a different user, gives `writeFile` success and `chmod` `EPERM`. That is the same case this PR is meant to fix.

Effect at each call site:

`decryptConfig` autosave: the full plaintext config is already on disk when the error happens. The catch block throws "Failed to write decrypted config file." That message is false. The write succeeded. An operator told the write failed has no reason to go delete a plaintext credentials file that is sitting there right now.

`changeKey`: the config file is already re-encrypted under the new key when the error happens. The catch block restores the old key in memory and throws "Failed to change the key." The caller is told rotation failed. On disk, rotation actually succeeded, and the restored old key can no longer decrypt the file. A `chmod` failure becomes a permanent lockout. This failure path does not exist before this PR.

`createConfigFileIfMissing`: these errors are not caught anywhere in `loadConfig`. On a filesystem where `chmod` fails, `init()` can never succeed. A plaintext placeholder file is left behind that later gets treated as a valid unencrypted config.

Suggested fix: wrap the `chmod` call in its own try/catch and log a warning instead of failing the write. Adopting the temp file and rename approach from item 1 removes the need for a second `chmod` call entirely, and removes this failure case with it.

## Should fix before merge

### 3. The new tests cannot detect the two defects above

`test/GCPKeyValueStorage.test.ts` mocks the whole `fs` module. `writeFile` and `chmod` become plain mock functions. The two new tests check only that a mock function received certain arguments. They never check a real file's mode on disk.

I traced eight plausible code changes against the new tests. Three of them still pass the test suite:

Reverting only the `decryptConfig` autosave call to a plain `fs.writeFile`: the tests still pass. No test in the file ever calls `decryptConfig`.

Reverting only the second write inside `createConfigFileIfMissing` to a plain `fs.writeFile`: the tests still pass. That line has no coverage anywhere in the suite.

Swapping the order inside the helper, so `chmod` runs before `writeFile`: the tests still pass, even though this breaks the real code. `chmod` on a file that does not exist yet fails with `ENOENT`.

Suggested fix: add a new test file that does not mock `fs`, and creates real files in a temp directory to check the resulting mode. `sdk/javascript/packages/core/test/localConfigStorage.test.ts` already has this pattern to copy from.

### 4. On Windows this fix does nothing

Node maps `fs.chmod` on Windows to a single read only attribute. It does not touch NTFS access control lists. Setting mode 0600 there clears the read only attribute and changes nothing else. Nothing throws, so this failure is silent.

The changelog line "restrictive file permissions (0600)" needs a note that this only applies on POSIX systems. Several other SDKs and the CLI already document this split.

### 5. The self heal only helps a leftover plaintext config, not an encrypted one

Every `chmod` call in this PR is tied to a write. A config file that is already encrypted, and never changes, is never written again after the first load. So a legacy config file at mode 0644 stays at 0644 forever, unless its value changes.

A leftover plaintext config is the one case that does get fixed, since loading it always triggers a write on the next call. That is a real fix, just a narrower one than the PR describes.

Suggested fix: add a best effort `chmod` call in the branch of `createConfigFileIfMissing` where the file already exists. Wrap it in a try/catch, since that branch cannot fail today.

### 6. This should be a minor version bump, not a patch

Locking down permissions on a file the SDK did not create is a change in default behavior. A deployment where a second local process legitimately reads this file today can break after this update. Version 1.0.1 is not yet published, so the number is free to move to 1.1.0 before release.

### 7. This PR conflicts with PR #1180 on the same file, and the easy merge resolution deletes the fix

PR #1180 (KSM-1370) targets the same base branch and edits the same function, `createConfigFileIfMissing`. I ran a real three way merge in both directions. Both directions produce the same conflict, in all three changed files.

The dangerous conflict is in the source file. PR #1180 moves a `configPath` declaration up a few lines. The merge conflict marker offers two choices at the disputed line: the version with this PR's secure write call, or a plain `fs.writeFile` with no redundant declaration. The second choice looks tidier. Taking it silently removes the security fix at that line.

I tested this directly. I merged both branches, resolved the conflict by mechanically taking one side across all three files, the shortcut a reviewer under time pressure might take, then ran the real test suite. The helper function was completely gone from the source file. All 59 tests still passed.

Suggested fix: agree on a merge order between the two PRs, and resolve the conflict by hand rather than by picking one side automatically.

### 8. Three sibling JavaScript packages have the same defect, unfixed, right now

The AWS, Azure, and Oracle KMS storage packages in this same SDK have the identical four site pattern, with no mode set and no chmod call. Their own release PRs for version 1.0.1 are open at the same time as this one: #1174, #1179, and #1178.

I have filed one epic and thirteen child tickets in JIRA for the sibling packages across JavaScript, Python, and Java, plus four secrets cache writers with the same class of gap. The epic is KSM-1463.

## Lower priority items

The directory that holds the config file is created with no restrictive mode, so it stays world readable even after this fix.

Both `writeFile` and `chmod` follow symlinks. A planted symlink at the config path can redirect a write, or cause an unexpected file to be chmod'd.

The first write inside `createConfigFileIfMissing` writes an empty placeholder that gets overwritten a moment later. It serves no purpose on the success path, and it can be left behind as a stray unencrypted file on certain failures.

Neither new test checks how many times `writeFile` was called. A version that writes the file twice, once insecurely and once securely, would still pass both tests.

## Items already tracked as separate tickets

During this review I found several issues in this file that are outside this PR's diff. A parallel review of PR #1180 found several of the same issues independently, and has already filed tickets for them. Rather than duplicate them here:

KSM-1461 covers an incomplete rollback in `changeKey` on failure, which can permanently destroy the app key.

KSM-1455 covers a zero length config file being silently replaced with an empty one.

KSM-1460 covers a KMS failure during auto encryption being misreported as a JSON format problem.

KSM-1456 and its children cover the lack of any atomic write anywhere in this file, which is the same underlying gap as blocking item 1 above.

## What this PR gets right

All four write sites are routed through the new helper. No site was missed.

The `mode` plus `chmod` approach matches an existing pattern already used elsewhere in this repository for the same class of bug.

The new `createConfigFileIfMissing` test already expects an `ENOENT` style error from `fs.access`, which happens to make it compatible with the fix in PR #1180.

The CI workflow for this package does run on this PR, and does exercise the changed code.

## Suggested replacement for the helper

```ts
import { createHash, randomBytes } from "crypto";

// fs.promises.writeFile opens with flag 'w', and open() only applies `mode` when it creates the
// file. On a config file that already exists, this writes the new bytes at the old mode, then
// tightens it after. It also truncates before the first new byte lands, with no fsync, so an
// interrupted write can leave an empty or partial file. Write to a fresh temp file in the same
// directory, sync it, tighten it while still open, then rename it over the target. Rename takes
// on the source file's mode, so the temp file needs to already be mode 0600 before the rename.
private async writeSecureConfigFile(targetPath: string, data: Buffer | string): Promise<void> {
  const finalPath = await this.resolveWriteTargetPath(targetPath);
  const tmpPath = `${finalPath}.${process.pid}.${randomBytes(6).toString("hex")}.tmp`;
  // 'wx' adds O_EXCL, so the open fails instead of reusing a file already at tmpPath. This
  // means mode is always applied at creation, and a planted symlink is rejected, not followed.
  const handle = await fs.open(tmpPath, "wx", 0o600);
  try {
    const payload = typeof data === "string" ? Buffer.from(data, "utf8") : data;
    const { bytesWritten } = await handle.write(payload);
    // write(2) can report fewer bytes than requested. Fail here instead of syncing and
    // committing a truncated file as the new config.
    if (bytesWritten !== payload.byteLength) {
      throw new GCPKeyValueStorageError(
        `Short write: wrote ${bytesWritten} of ${payload.byteLength} bytes to ${tmpPath}`
      );
    }
    await handle.sync();
    // fchmod on the open descriptor, not chmod by path. This pins the exact file instead of
    // resolving the name again, and it sets exactly 0600 no matter what the umask is.
    await handle.chmod(0o600);
  } catch (writeError) {
    await handle.close().catch(() => undefined);
    await fs.unlink(tmpPath).catch(() => undefined);
    throw writeError;
  }
  await handle.close();
  try {
    await fs.rename(tmpPath, finalPath);
  } catch (renameError) {
    await fs.unlink(tmpPath).catch(() => undefined);
    throw renameError;
  }
}

// rename() replaces the directory entry instead of following it. Without this step, a
// symlinked config path would be replaced by a plain file. Some setups manage the config
// path as a symlink on purpose, and Kubernetes mounts a Secret as a symlink chain.
private async resolveWriteTargetPath(targetPath: string): Promise<string> {
  try {
    return await fs.realpath(targetPath);
  } catch (resolveError) {
    const code = (resolveError as { code?: unknown } | null | undefined)?.code;
    if (code !== "ENOENT") {
      throw resolveError;
    }
    return targetPath;
  }
}
```

Two changes worth flagging on purpose: after this change, the write needs execute permission on the parent directory, not just write permission on the file. A hard link to the config file will stop tracking new content, since rename always creates a new inode. Both changes match what the core package already accepts for the same pattern.

## How I checked this

I did not rely only on reading the code. I ran the actual test suite on this PR's head commit: `npm ci` then `npx jest --maxWorkers=5`. All 59 tests pass, matching the green CI check.

I wrote a small script with no mocked `fs` calls. It created a config file at mode 0644, then ran this PR's exact two line helper as two separate steps, and checked the file's mode between the two steps. The mode was still 0644 right after the write, before the chmod ran.

I opened a read handle on a mode 0644 file before running this PR's helper on it. After the helper finished and the file reported mode 0600, I read again through the same handle. It returned the new secret content in full.

I repeated the same test against the temp file and rename approach suggested above. The file's inode number changed after the write. The same old read handle, read again, returned only the old content.

I reverted three different single lines in the source, one at a time, matching the three surviving test gaps from item 3 above, and ran the real test suite each time. All three reverts left the full 59 test suite green.

I ran a real merge of this PR and PR #1180 in both directions in a disposable branch. Both directions produced the same conflict. I then resolved that conflict by taking one side across all three files, committed the result, and ran the real test suite. The security fix was completely gone from the source file. The test suite still reported all 59 tests passing.

mgallego-keeper (changes requested):
## Summary

This follow up covers the latest push, three commits on top of what I reviewed before. The branch now includes PR #1180's two commits as its base, with this PR's own fix rebased on top, and adds a new `src/atomicWrite.ts` module plus a new real filesystem test file.

I re-read every changed line, then ran the actual test suite, wrote independent tests against the real code, and reverted specific lines one at a time to check whether the suite would catch each one. Full detail is in the verification section at the end.

**The two blocking items from the first review are genuinely fixed, and I confirmed this against the real code, not just by reading the diff.** The version bump and the changelog section move are also both done. Credit for that below.

**Requesting changes for one reason: three write paths in this file have no test that would catch a regression in the exact property this PR exists to add.** All three are listed below with a suggested test for each. I ran every suggested test against the current code, where it passes, and against the specific regression it targets, where it fails, before including it here.

## What is fixed, confirmed against the real code

The new `writeFileAtomicSync` in `src/atomicWrite.ts` opens a temp file in the same directory, writes to it, syncs it, sets its mode, then renames it over the target. I wrote a test that opens a read handle on a legacy mode 0644 config file, then calls `storage.saveString()`, the real public method a consumer would call, not a private method. The file's inode changes and lands at mode 0600. The old read handle, read again afterward, still returns only the old content. The fix works, confirmed through the same entry point a real caller uses.

The `chmod` call in the new helper targets the temp file, which the process just created and owns. This removes the specific failure this review flagged before, where a legacy file owned by a different user made `chmod` fail after the write had already succeeded. That specific failure case no longer applies, since the process always owns the file it is chmoding now.

The merge conflict with PR #1180 was resolved by hand, not by picking one side. I read the resulting code directly. It keeps both the check from #1180 and the secure write from this PR at the one line where they collided.

Version 1.1.0, and the changelog entry under a Security heading. Both match what I asked for.

## Requesting changes: three write paths have no test for this property

### 1. `saveConfig()`'s own write has no test, and this is the most used path in the whole class

Every `set`, `saveString`, `saveBytes`, and `delete` call goes through `saveConfig()`. Its write call used to have a test that checked the mode argument passed to a mocked `fs.writeFile`. That test is gone now, and nothing real replaced it. I confirmed this two ways. First, by counting: the old file had two tests for file mode, the new mocked file has one, for an unrelated case, and the new real file has three tests, none of which independently checks this specific call. Second, by reverting this one line back to a plain write and running the real suite. All tests still pass.

This matters because a legacy config file, one that predates this whole fix, is exactly the case this PR is meant to protect. If this line regresses later, the file will keep loading and saving. Nobody will notice its mode stayed at 0644.

Suggested test, added to `GCPKeyValueStorage.atomicWrite.test.ts`. I ran this against the current code, where it passes, and against the reverted line, where it fails with the mode staying at 0644 instead of reaching 0600.

```ts
describe('saveConfig()', () => {
    it('corrects a pre-existing 0644 config file to 0600 on an ordinary save', async () => {
        fs.writeFileSync(configPath, '', { mode: 0o644 });
        const { storage } = makeStorage(configPath);
        expect(fileMode(configPath)).toBe(0o644);

        await (storage as any).saveConfig({ clientId: 'abc' });

        expect(fileMode(configPath)).toBe(0o600);
    });
});
```

### 2. The second write inside `createConfigFileIfMissing()` has no test either

This function writes twice when the file is missing: an empty placeholder first, then the real encrypted blob. I reverted only the second write back to a plain write and ran the suite. All tests still pass, because the file already exists by that point, created moments earlier by the first write, so its mode stays correct either way. The test that exercises this function only checks the final mode, which the first write already set, so it cannot tell whether the second write itself went through the secure path.

The security impact of this one specifically is low, since the file was never in an insecure state for anyone to read before this line runs. It is still worth a direct test, since a version that no longer calls the secure helper here would otherwise ship unnoticed.

Suggested test, same file. I ran this against the current code, where it passes with two calls recorded, and against the reverted line, where it fails, reporting one call instead of two.

```ts
describe('createConfigFileIfMissing() call sites', () => {
    it('routes both the placeholder write and the encrypted write through writeFileAtomicSync', async () => {
        const atomicWrite = require('../src/atomicWrite');
        const spy = jest.spyOn(atomicWrite, 'writeFileAtomicSync');
        const { storage } = makeStorage(configPath);

        await (storage as any).createConfigFileIfMissing();

        expect(spy).toHaveBeenCalledTimes(2);
        expect(spy.mock.calls[0][0]).toBe(configPath);
        expect(spy.mock.calls[0][1]).toEqual(Buffer.from('{}'));
        expect(spy.mock.calls[1][0]).toBe(configPath);
        spy.mockRestore();
    });
});
```

### 3. The `chmod` call inside the new helper has no test, and it is not just a redundant safety net

`writeFileAtomicSync` opens the temp file with mode 0600, then calls `chmodSecure` on it again right before the rename. I assumed at first this second call was pure defense in depth, since the open call already sets the mode. I checked this directly instead of assuming.

Under a normal umask, the assumption holds, the mode is already correct before the `chmod` call runs. But `open()`'s mode argument is still masked by the process umask, the same as any file creation. Under an umask that strips owner bits, for example `0o700`, the file comes out of the open call at mode 0, unreadable even by its own owner, and the explicit `chmod` call is the only thing that puts it back to 0600. I confirmed this with a small standalone script before writing this test, then confirmed the test itself catches it by removing the `chmod` call and running the suite: the mode assertion fails, reporting 0 instead of 0600.

This is a narrow case, an unusual umask is not a normal deployment, but the consequence of losing this line silently is not just weaker security, it is a config file the SDK cannot even read back. Worth a test given how cheap it is to write.

```ts
describe('writeFileAtomicSync() under a pathological umask', () => {
    it('still lands at exactly 0600 even under an owner-stripping umask', async () => {
        const originalUmask = process.umask(0o700);
        try {
            const { storage } = makeStorage(configPath);
            await (storage as any).writeSecureConfigFile(configPath, 'SECRET');
            expect(fileMode(configPath)).toBe(0o600);
        } finally {
            process.umask(originalUmask);
        }
    });
});
```

## Two smaller items, not blocking

`package-lock.json` still says version 1.0.1 in its own two self reference fields, even though `package.json` and the changelog both correctly say 1.1.0. `npm ci` does not fail on this today, but it is worth fixing for consistency.

PR #1180 is still open, with the same two commits this PR's branch now also contains. This repository merges release branch PRs by squash, based on a recent example I checked, so merging this PR first should make #1180 a no-op to merge afterward. Worth closing #1180 explicitly once this one lands, rather than leaving it open and confusing.

## How I verified this

I fetched the actual current branch, not the commit I reviewed before, and confirmed the release branch itself had not moved. I read every changed file in full, not just the diff.

I ran the real test suite on the current branch: 4 suites, 61 tests, all passing.

I wrote a test that drives the fix through `storage.saveString()`, the real public method, against a legacy mode 0644 file with a read handle already open on it, and confirmed the inode changes, the mode reaches 0600, and the old handle still reads only the old content.

I reverted three specific lines, one at a time, restoring each before testing the next, and ran the real suite after each revert: `saveConfig()`'s write call, the second write inside `createConfigFileIfMissing()`, and the `chmod` call inside the new helper. All three reverts left the full suite passing.

I wrote a small standalone script to confirm the umask interaction directly, outside of any test framework, before relying on it in a test.

For each of the three suggested tests above, I added it to a scratch copy of this branch, confirmed it passes against the current code, then applied the exact revert it is meant to catch and confirmed it fails, with the specific numbers shown in each section above.

I also ran a full TypeScript build and eslint against the new source file. Both are clean. Neither runs in this package's CI.

mgallego-keeper (approved):

PR #1184, opened 2026-09-21
fix(sdk/javascript): reject a zero-length GCP KMS config instead of resetting it

mgallego-keeper:
## Summary

`loadConfig()` treated a zero-length config file as an empty config, re-encrypted it, and wrote it back over the damaged file. That destroyed the client ID, app key, and device private key with no recovery path. A zero-byte file can only come from a truncated or interrupted write, so `loadConfig()` now throws and leaves the file untouched.

### Fixed
- `loadConfig()` no longer substitutes `{}` for a zero-length config file and persists it. It logs at error level, throws an error naming the config path, and writes nothing (KSM-1455)

## Testing
```
npm test
```
New suite `test/GCPKeyValueStorage.zeroLengthConfig.test.ts` uses a real zero-byte file on disk with no `fs` mock. It proves `init()` rejects, the file stays at 0 bytes and at the same inode, and `writeFileAtomicSync` is never called. Three control tests confirm a genuinely missing file, an existing encrypted config, and a non-empty plaintext JSON config all still work. All 4 zero-length tests fail against the pre-fix code.

## Breaking Changes

None for any working configuration. `init()` now rejects on a zero-length config file where it previously resolved and silently reset the file. That path only occurs on a corrupted file, and the previous outcome was unrecoverable data loss.

## Related Issues
- Jira: KSM-1455


stas-schaller (approved):
👍 nice one

PR #1185, opened 2026-09-21
JS SDK GCP KMS Storage: treat a blank config file path as absent

mgallego-keeper:
## Summary

An empty `KSM_CONFIG_FILE` environment variable was accepted as a valid config file path, because `??` falls back only on `null` and `undefined`. `path.resolve("")` is the current working directory and `fs.access` succeeds on a directory, so the storage reported the working directory as an existing config file and failed later with a confusing empty path. Both fallback sources now treat a blank value as absent.

### Fixed
- A blank `KSM_CONFIG_FILE` environment variable, or a blank `keyVaultConfigFileLocation` argument, now falls back to the next source instead of resolving to the current working directory (KSM-1457)

## Testing
```
npm test
```
15 new tests cover the empty and whitespace-only forms of both sources, plus the no-regression cases for non-empty values. 12 of them fail against the previous constructor; one asserts through `createConfigFileIfMissing()` that `fs.access` is called on the resolved default config file and never on `process.cwd()`, which is the actual defect rather than just the field value.

## Breaking Changes

None. A blank path was never usable, so no working configuration changes behavior. Version stays at 1.1.0, which is unreleased.

## Related Issues
- Jira: KSM-1457


stas-schaller (approved):
Ship it 🚀

PR #1186, opened 2026-09-21
fix(sdk/javascript): recreate a deleted GCP KMS config file on an unchanged save

mgallego-keeper:
## Summary

`saveConfig()` checked the in-memory config hash before it checked the config
file, so a file deleted while the process kept running was never recreated if
the in-memory value had not changed. `saveString()` and `saveStorage()`
resolved successfully and wrote nothing. A missing file now forces a full save
of the real config.

### Fixed
- `saveConfig()` no longer skips the write when the config file was deleted while the process was running; the no-changes-detected skip now also requires the file to still be present, and a missing file falls through to a full save of the current config rather than being left to `createConfigFileIfMissing()`, which would restore only an empty `{}` placeholder (KSM-1459)

## Testing
```
npm test
```
Five new real-filesystem tests delete the config file out from under a live
storage instance and then save the same value again. They assert that the
recreated file decrypts back to the real config, and that a fresh instance's
`init()` reads the value back. The content assertion is the one that matters:
a fix that only reorders `createConfigFileIfMissing()` ahead of the early
return makes the file exist again while holding `{}`, and that variant fails
these tests. A fifth test confirms the no-changes-detected skip still avoids
the write when the file is present.

## Breaking Changes

None to the public API. One behavior change worth noting: an unchanged save on
a config path whose existence cannot be determined at all (an `fs.access`
failure other than `ENOENT`, for example `ENOTDIR` or `EACCES`) now rejects
instead of resolving with only a warning. The SDK cannot confirm persistence
in that state, and this matches the policy KSM-1370 already set for
`createConfigFileIfMissing()`.

## Related Issues
- Jira: KSM-1459


sdk/javascript/packages/gcp/src/GCPKeyValueStore.ts
```diff
@@ -295,8 +295,15 @@ export class GCPKeyValueStorage implements KeyValueStorage {
...
+      // save of this.config, because createConfigFileIfMissing() would otherwise recreate it
+      // holding only an empty "{}" placeholder.
+      if (
+        !force &&
+        configHash === this.lastSavedConfigHash &&
+        (await this.configFileExists())
+      ) {
         this.logger.warn("Skipped config JSON save. No changes detected.");
         return;
       }
```
stas-schaller, line 313: Once configFileExists() is false, this falls through to createConfigFileIfMissing() (line 312), which writes an encrypted {} placeholder, and then the save below immediately overwrites it with the real config. That's two KMS encrypt calls and two disk writes for one recovery. Let's have this path write the real config directly instead of going through the placeholder first.

sdk/javascript/packages/gcp/src/GCPKeyValueStore.ts
```diff
@@ -295,8 +295,15 @@ export class GCPKeyValueStorage implements KeyValueStorage {
         }
       }
 
-      // Check if saving is necessary
-      if (!force && configHash === this.lastSavedConfigHash) {
+      // A matching hash only proves the in-memory config is unchanged, not that the file on disk
+      // still holds it. A file deleted underneath a running process must fall through to a real
+      // save of this.config, because createConfigFileIfMissing() would otherwise recreate it
+      // holding only an empty "{}" placeholder.
```
stas-schaller: "holding only an empty {} placeholder" reads as if that's the end state. It only persists if the process exits before the save below completes, otherwise it's overwritten within the same call. Worth tightening this to say so.

stas-schaller (changes requested):
Requesting changes, two small fixes:

1. The recovery path encrypts and writes twice. Let's fix that before merge.
2. The placeholder comment overstates when the {} actually lands on disk.

Verified against `457ae70`.

PR #1187, opened 2026-09-21
JS SDK GCP KMS Storage: report KMS failures during config auto-encryption

mgallego-keeper:
## Summary

`loadConfig()` encrypted a plaintext config file inside the try/catch that detects whether the file is plain JSON. A KMS failure therefore looked like a parse failure. The SDK reported an intact config file as damaged.

### Fixed
- `loadConfig()` now encrypts a detected plaintext config file outside the try/catch that tests for plain JSON, so a KMS outage, a revoked permission, or a disabled key version propagates as the real KMS error (KSM-1460)

## Testing
```
npm test
```
The new test writes a real plaintext config file, then makes the KMS encrypt call fail. Three assertions fail against the old code and pass against the new code. Four more assertions pass against both, which proves the corrupt-file messages and the healthy-KMS path did not change.

## Breaking Changes

None. The error for a genuinely unreadable config file stays the same.

## Related Issues
- Jira: KSM-1460


stas-schaller (approved):
Two thumbs up 👍👍

